PR #323, opened 2015-01-04
Add guards to tau/bounders/elementary-bounders.lisp

nadezhin:
Tau-bounders are usually called by tau-system, but my project calls them directly.
So I wrote guards for bounders in tau/bounders/elementary-bounders.lisp.

### DIFF
--- a/books/tau/bounders/elementary-bounders.lisp
+++ b/books/tau/bounders/elementary-bounders.lisp
@@ -534,6 +534,7 @@
 ; BINARY-+
 
 (defun domain-of-binary-arithmetic-function (domx domy)
+  (declare (xargs :guard (and (symbolp domx) (symbolp domy))))
 
 ; In this function, domx and domy are domain names as found in tau-bounders,
 ; e.g., INTEGERP, RATIONALP, ACL2-NUMBERP, or NIL.  The function returns the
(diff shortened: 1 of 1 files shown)


ragerdl:
@MattKaufmann Do you see any problem with merging this?

@nadezhin Thanks for the improvements

MattKaufmann:
Hi, David --

I don't really know anything about that book, which J wrote.  I've
already made a note to discuss the changes with him tomorrow morning.
Please go ahead and merge; if J sees anything that might be good to
change, that can be done tomorrow.  This way, things can be in place
when hopefully I start testing for the Version 7.0 release tomorrow.

(So, thanks in advance for merging!  And @nadezhin, thanks for
contributing!)

Thanks --
-- Matt
   Date: Sun, 04 Jan 2015 13:47:09 -0800
   From: "David L. Rager" <notifications@github.com>
   Cc: MattKaufmann <kaufmann@cs.utexas.edu>

   @MattKaufmann Do you see any problem with merging this?

   @nadezhin Thanks for the improvements

   ---

nadezhin:
Matt, David,

Thank you for the commit.

I have a question.
In the pull request I have to rename some functions in
./books/tau/bounders/elementary-bounders.lisp:

lower-bound->  to lower-bound->-guarded
upper-bound->  to upper-bound->-guarded
squeeze-k to squeeze-k-guarded

I made them guarded and ACL2 reported conflicts with
functions lower-bound->, upper-bound->, squeeze-k from  ./tau.lisp .

What is better ?
A) Leave these functions with the suffix "-guarded"
B) Request you to add these guards to functions in ./tau.lisp
     and then remove "guarded" suffix in
./books/tau/bounders/elementary-bounders.lisp:

If the answer is (B), should I prepare a pull request for ./tau.lisp ?
My guards used "real/rationalp" predicate.
{{{
(defun squeeze-k-guarded (upper-boundp rel k)
  (declare (xargs :guard (and (booleanp upper-boundp)
                              (booleanp rel)
                              (or (null k) (real/rationalp k)))))
}}}
Now I think that it was not correct.
It is better to use "rationalp" predicate in guards
because computable arguments of functions are always rationalp.
{{{
(defun squeeze-k-guarded (upper-boundp rel k)
  (declare (xargs :guard (and (booleanp upper-boundp)
                              (booleanp rel)
                              (or (null k) (rationalp k)))))
}}}

ragerdl:
Hi Matt,

I'm going to let you answer these two questions. My intuition is that we'd
want to replace the non-guarded versions of the functions in tau.lisp with
the new versions that Dmitry has committed. As for whether we want to do
that before 7.0, well, that's also up to you.

Thanks,
David

PR #324, opened 2015-01-05
Fix ACL2(r) certification of elementary-bounders.lisp broken by pull req...

nadezhin:
...uest #323 .
Certification ACL2(r) was broken by my previous pull request #323. Sorry. This fix repairs it.
The file "elementary-bounders.lisp" contains many other guards with "real/rationalp",
but they don't break ACL2(r) certification.
I need an advice. Is it good to use real/rationalp in guards ? Should I replace other occurances of real/reationalp ?


MattKaufmann:
Thank you very much for the fix!  I've merged your pull request (my
first merge of a pull request!).

>> I need an advice. Is it good to use real/rationalp in guards ? Should
>> I replace other occurances of real/reationalp ?

I don't think there's a general answer.  Here's all I can think of to
say about that, which might already be obvious to you -- but it's a
good question and perhaps others on the list will find it to be
helpful.

The guiding principle is that it is best to require rationalp when
that truly is what is required; otherwise, real/rationalp is best (for
compatibility with ACL2(r)).  In your particular case (function
combine-intervals1), rationalp truly is required; here's why.  The
question is (for example) whether the guard should require

  (or (null hi1) (real/rationalp hi1))

or:

  (or (null hi1) (rationalp hi1))

The body of your function has a call of make-tau-interval for which
hi1 might be passed as the last argument.  But note the guard for
make-tau-interval:

  ACL2 !>:args make-tau-interval

  Function         MAKE-TAU-INTERVAL
  Formals:         (DOM LO-REL LO HI-REL HI)
  Signature:       (MAKE-TAU-INTERVAL * * * * *)
		   => *
  Guard:           (AND (OR (NULL LO) (RATIONALP LO))
			(OR (NULL HI) (RATIONALP HI)))
  Guards Verified: T
  Defun-Mode:      :logic
  Type:            (CONSP (MAKE-TAU-INTERVAL DOM LO-REL LO HI-REL HI))

   MAKE-TAU-INTERVAL
  ACL2 !>

So here, we really do need to require that hi1 is nil or a rational,
so that the last actual for make-tau-interval will be nil or a
rational.

-- Matt
   Date: Mon, 05 Jan 2015 04:16:40 -0800
   From: nadezhin <notifications@github.com>

   ...uest #323 .
   Certification ACL2(r) was broken by my previous pull request #323. Sorry. This fix repairs it.
   The file "elementary-bounders.lisp" contains many other guards with "real/rationalp",
   but they don't break ACL2(r) certification.
   I need an advice. Is it good to use real/rationalp in guards ? Should I replace other occurances of real/reationalp ?
   You can merge this Pull Request by running:

     git pull https://github.com/nadezhin/acl2 master

   Or you can view, comment on it, or merge it online at:

     https://github.com/acl2/acl2/pull/324

   -- Commit Summary --

     * Fix ACL2(r) certification of elementary-bounders.lisp broken by pull request #323

   -- File Changes --

       M books/tau/bounders/elementary-bounders.lisp (8)

   -- Patch Links --

   https://github.com/acl2/acl2/pull/324.patch
   https://github.com/acl2/acl2/pull/324.diff

   ---
   Reply to this email directly or view it on GitHub:
   https://github.com/acl2/acl2/pull/324

PR #326, opened 2015-01-06
Synchronize a few guards in tau.lisp and elementary-bounders.lisp

nadezhin:
These functions were defined both in tau.lisp and in books/tau/bounders/elementary-bounders.lisp :
lower-bound->
upper-bound-<
squeeze-k  .
When I added guards to elementary-bounders.lisp, ACL2 reported that the guarded functions conflict with original functions in tau.c. Suffix "-guarded" was appended to resolve it quickly.

It is necessary to clean this solution ? This pull request tries to clean it by copying guards to tau.lisp .

### DIFF
--- a/books/tau/bounders/elementary-bounders.lisp
+++ b/books/tau/bounders/elementary-bounders.lisp
@@ -1624,11 +1624,11 @@
 ; Since (mod x y) = (- x (* (floor x y) y)), the domain of MOD is given
 ; by domain-of-binary-arithmetic-function:
 
-(defun lower-bound->-guarded (a-rel a b-rel b)
+(defun lower-bound-> (a-rel a b-rel b)
   (declare (xargs :guard (and (booleanp a-rel)
-                              (or (null a) (real/rationalp a))
+                              (or (null a) (rationalp a))
                               (booleanp b-rel)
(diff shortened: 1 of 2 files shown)


MattKaufmann:
These changes look reasonable (thanks!).  I'll check with J Moore, since
he is the one who maintains tau.lisp.  If he agrees with the changes, I might
make them manually (and credit you in the release notes), since normally
only J and I write to ACL2 source files (as enforced by the script bin/purity.sh);
I'm  not sure what's best.

ragerdl:
That seems fine to me -- you often implement my changes to the ACL2 System
manually too, but the change list starts with a patch that I submit
(sometimes via a pull request).  So, I think the process is working great.

MattKaufmann:
Quoting my commit (see https://github.com/acl2/acl2/commit/ffcda602ce1c266037054a5b3089e4ec02416ebc):

    The former (tau-related mods) closes pull request #326.  I am doing
    that manually, rather than by automated merging, because the changed
    files included an ACL2 source file (tau.lisp), and we continue the
    convention that only J and I write to the ACL2 source files, as
    enforced by bin/purity.sh (which is called by bin/pull.sh).

PR #329, opened 2015-01-06
Changed path of ACL2s books.

harshrc:
I had to do this in preparation for the next commit which puts ACL2s exports into a new package and follows the package discipline followed by STD (and advocated by Jared in one of the emails).


MattKaufmann:
Thanks, David.  Your sentiment about not breaking the build is right
on target (according to my preferences, at least), and your decision
to merge was perfect, for the reason you gave.  I plan to rebuild the
manual tonight and I can get with Harsh about any problems.

Thanks --
-- Matt
   Date: Tue, 06 Jan 2015 15:42:32 -0800
   From: "David L. Rager" <notifications@github.com>


   @harshrc Okay, since you're on a laptop, we might have you start creating pull requests for the "testing" branch, so that we can check that things are stable before merging into the master branch.  That being said, the `manual` target is the gatekeeper for that process, so it sounds like things are equivalently in order this time.

   I'm particularly trying to avoid breaking the build right before Matt starts release, but I expect that you're trying to get these files into the release, so merging is probably the right answer regardless.

PR #330, opened 2015-01-07
ACL2 Sedan changes synced.

harshrc:
-  (Big change : New package - ACL2S)
  Moved to the package/portcullis discipline followed by STD in
  acl2-books. Most API functions for Defdata, Cgen, and the defdata
  type namespace is in now in the ACL2S package. To do this we also
  moved cgen,defdata and ccg to a new directory in books called
  acl2s where the ACL2S package and portcullis files reside.

- ACL2s defaults (in cgen/acl2s-parameter.lisp) are now package agnostic.

- All names created in course of defdata are in the current-package.

- Defdata/Sig changes:
  + Fixed bug in sig to take into account signatures with more than one
    type variable.
  + Singleton types handled now by using (lambda (x) (equal x
    constant)) as the predicate name -- this works like a charm for
    functional instantation in ACL2.

  In addition we restrict sig in following ways.
  + If polymorphic type expression to be matched is a variable i.e :a, we skip.
  + When finding matches in the type table, we skip over types that do
    not have non-empty prettyified-def entry

- New files added
  + Added defunc which is defun with contracts.
  + Added base-theory.lisp and custom.lisp to let acl2s-mode.lsp reside
    here.  This way we can simulate the ACL2s environment without
    going through the ACL2s website and downloading ACL2s Eclipse.

  + Base-theory.lisp builds on std library and sets up polymorphic
    theory support (automated instantiation of generic type
    theorems).

- Numerous bug fixes in Cgen, Defdata and improved error messages.

- Fixed bug in CCG to get it to work in encapsulate/include-book contexts.
  Synced ccg defuns functions with the corresponding changes in defuns.lisp.

We are now in sync with Sedan svn info:
Last Changed Rev: 1497
Last Changed Date: 2015-01-06 22:54:15


harshrc:
I tried to 'make manual' but I got a quicklisp error. I hope Matt can check this for me.

harshrc:
Discarding this pull request because I overwrote Matt's changes to ccg.lisp.

kini:
By the way, you can edit pull requests after the fact -- simply push a different local branch to the branch on github associated with the pull request (with `-f` as necessary). And it looks like in this case, that's exactly what you did with your `master` branch before making pull request #331, so you could have just left this open, pushed the commits, and then this PR would be the same thing as #331.

harshrc:
thanks @kini 
i think now im more versed with git, but i believe i am still ignorant of github-specific what-to-do.

kini:
If you ever need help, feel free to ping me on IRC in the #acl2 channel!

PR #332, opened 2015-01-07
minor doc fixes to get rid of broken links as reported by Matt.

harshrc:


ragerdl:
With the exception of the reference to CGen, these all look like improvements to me, so I'll go ahead and merge it.  I'll trust harsh on the CGen reference.

PR #333, opened 2015-01-07
Fixed parent links in various acl2s doc topics and also added ccg doc (acl2s/doc.lisp) to manual

harshrc:


MattKaufmann:
(I'll start a manual rebuild right now as an extra check.)

harshrc:
I checked that the manual built successfully and that there were no non-existent parents in acl2s doc topics

PR #354, opened 2015-03-20
Add a lemma to std/lists/duplicity

kini:
We already have `duplicity-when-non-member-equal'; this commit adds
a complementary lemma, `duplicity-when-member-equal'.  I found it useful
in a proof I was doing, and since `duplicity' needs to be enabled in
order to prove it, it seemed like putting it in std/lists/duplicity made
sense.


kini:
Fixed in 4149b83ddcb47f595e161acfb3edb8e799dee473

PR #360, opened 2015-03-30
Remove ``'' from code fragments

kini:
Maybe these were added by mistake at some point...?


ragerdl:
Thanks for these cleanups -- the given code doesn't work without them.

ragerdl:
These changes introduce a dependency loop -- I'm working on the fix.

PR #366, opened 2015-04-12
Added theorem about sets as lists.

acoglio:


ragerdl:
Thanks for the suggestion.  May I ask, do all of the books build with this change incorporated?

acoglio:
Yes, the regression succeeded after the change.

jaredcdavis:
Perhaps it would be better to instead prove:

```
(defthm set-equiv-of-nil
  (equal (set-equiv nil x)
         (atom x))
  :hints(("Goal" :in-theory (enable set-equiv))))
```

This would combine nicely with `std/lists/append:consp-of-append` to prove your theorem, but would also be more generally applicable.  I haven't tried a regression with it yet, though.

acoglio:
That seems like a good idea. I’ve replaced ours with yours and successfully run the regression. I believe that this new commit has become part of the outstanding pull request, but I’m not a github expert so please let know if that’s not the case.

jaredcdavis:
Looks good to me -- thanks for running the regression.  Merging.

Cheers,
Jared

PR #369, opened 2015-04-24
Prove theorems in rtl/rel11/support/basic.lisp not using previous releas...

nadezhin:
...es of RTL.
Books in `rtl/rel11/support/` directly include  `rtl/rel11/rel9-rtl-pkg/lib/top` and transitively include other books from previous releases of RTL.
It is difficult to learn and to prepare enhancement when it is necessary to look through the history of RTL releases. I started from `basic.lisp` and tried to make proofs in `rtl/rel11/support/basic.lisp` self-sufficient.

### DIFF
--- a/books/rtl/rel11/support/basic.lisp
+++ b/books/rtl/rel11/support/basic.lisp
@@ -1,16 +1,14 @@
 (in-package "RTL")
 
-(local (include-book "../rel9-rtl-pkg/lib/top"))
-
-(include-book "../rel9-rtl-pkg/lib/util")
+(include-book "std/util/defrule" :dir :system)
 
 (local (include-book "arithmetic-5/top" :dir :system))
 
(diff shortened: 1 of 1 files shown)


ragerdl:
I'll take ownership of this issue.  I'll need to call Russinoff and see whether he wants rel11 cleanedup, a new rel12, or what a good path towards cleaning these up is.  I'd also be delighted if we were able to strip out the mess, so thanks for the start on this!

ragerdl:
Will merge this after 7.1 release.

See #374 for additional details.

ragerdl:
I ran this in a separate checkout, and doc/top built.  Given only a support book was modified and that doc/top includes the most recent rtl library (plus, the books built), this implies that eb139f6 is a reasonable commit to merge.

PR #393, opened 2015-05-27
Added some examples that illustrate program execution on the x86isa

ragerdl:
Added some examples that illustrate program execution on the x86isa model. Specified the SYSCALL and SYSRET instructions in the
system-level mode. Some guard proof clean-up for these instructions is
pending...


ragerdl:
I ran this on my infrastructure since leeroy is still out of pocket.

PR #402, opened 2015-06-04
Improving the call of zip.sh to use tshell instead of sys-call+

ragerdl:


ragerdl:
Merging this into the testing branch.

PR #406, opened 2015-06-08
Configurable ACL2 path and image extension in SV cosims

arbrock:
Hi @solswords @jaredcdavis  ,
Here's a pull request for making the cosim Makefile more generic and basing it off the default CCL settings.  Could you please merge this when you have a moment to update your scripts that call cosim and have them pass in IMG_EXT=ccl ?

-- Andrew and David

### DIFF
--- a/books/centaur/sv/cosims/Makefile
+++ b/books/centaur/sv/cosims/Makefile
@@ -36,8 +36,10 @@ all:
 #  Building the cosim image
 
 STARTJOB ?= $(SHELL)
+ACL2 ?= acl2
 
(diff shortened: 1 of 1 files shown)


solswords:
I'd rather not have to remember IMG_EXT=ccl each time I run it, either.  Could you try just removing the references to $(CORE).ccl and cosim-core.ccl?  I don't think it should be necessary to delete those files before running acl2, and we don't really need to see the ls -l output either.

arbrock:
That makes sense: I've made those changes.

Issue #399, opened 2015-06-03
Swapped indices in SV arrays


arbrock:
When attempting to model an array of registers, we noticed that one variable was created per bit in the registers. We suspect that this is not the intended model, in that we expected one variable per array entry.

One can see this strangeness in the referenced example, where "wires" r._0_ through r._3_ were created for an array of _two_ 4-bit registers.

In this example, overriding each register to 1'b1 assigns the low two bits, leaving the high bits high-impedance. (This can be seen using svtv-debug.)  Access, however, occurs in row-major order, instead of the column-major order that assignment uses. This results in r[0] being read as having value 3 and r[1] being read as having value 0, despite both being overridden to 1.

https://gist.github.com/arbrock/49974330bc9bde95633b

solswords:
reg [1:0] foo [3:0] is a 4-slot, unpacked array of 2-bit registers, not 2
4-bit registers -- does that account for your problems?

centaur/sv/cosims has a bunch of tests where VL/SV is tested against
commercial simulators (NCV and VCS specifically).  Unfortunately at the
moment it can only test combinational logic, but most bugs can still be
revealed in such a test.  Each subdirectory contains a spec.sv which
defines a module named spec with the given interface --

module spec (input logic [127:0] in,
         output wire [127:0] out);

and an inputs.data file.  If you can set up a failing test (but without
initial blocks or sequential logic), that would be a good bug report.  If
you don't have both NCV and VCS, you can change the cosims/Makefile and
run.lsp to only check one implementation or the other (or maybe another
simulator instead).

solswords:
Looking a bit closer, it seems this is a bug in the SVTV wire name parsing.  Will consider what to do with it.

arbrock:
That solves some of our problems: the array now has the number of bits expected, and there are no high-impedance nets. However, the swapped indices appear to still be an issue.

I will try to set up a failing test.

solswords:
I think I fixed this in commit 36a17fbaee5a.  There is still room for bugs -- parsing of these wire names is trying to emulate SystemVerilog wire parsing, but completely different implementation than VL.

PR #407, opened 2015-06-08
Adding two regression tests to SV cosims:

arbrock:
This is a pull request for a pair of cosim regressions I've written:
  - _regfile_ ensures arrays are indexed in the correct order: designed to detect bug #399 
  - _nested_ detects a bug in handling of else statements.
At present _nested_ does not pass. I hope these are useful: please merge them whenever it is appropriate.

Thank you,
Andrew


ragerdl:
@solswords @jaredcdavis 

Is it okay to have cosim regression test that's currently failing as part of the repository?  If so, I'll merge this.

solswords:
AFAIK nobody's depending on that makefile building successfully, but I'd rather there not be failing tests in there with no plans for a fix.  (We should probably have some tests that "pass" by failing the right way, but we don't have this set up yet.)  I'll merge this for now & check on what the problem is.

solswords:
This is fixed; it turned out that in some cases IF tests (and other similar expressions) were getting translated to svex extended with Xes rather than correctly sign or zero-extended.

solswords:
BTW, in SystemVerilog if/then/else statements have truly horrible X behavior:
 if (1'bX) is the same as if (1).  VL/SV is not going to support this particular travesty, which means some cosim tests need to have binary-only inputs.  I changed that "nested" test's inputs.data file so that this is the case.

PR #410, opened 2015-06-09
Cosim regressions

arbrock:
I have narrowed the failure case for the "nested" cosim test added yesterday, so I have added a comment explaining it.


ragerdl:
I didn't want to change the example out from under @solswords 's feet, but reportedly it fails with an even simpler version that uses a constant in place of an array reference.

PR #418, opened 2015-06-24
Moved theorem (added some time ago) under a suitable defsection so it…

acoglio:
… appears in the manual.


ragerdl:
This looks okay to me, but I'm going to let Jared merge this since it's in his code.

jaredcdavis:
Sure, seems just fine.  Thanks!

PR #421, opened 2015-06-25
Fixing typo in VL syntax version datestamp

ragerdl:
@jaredcdavis or @solswords Please confirm that decreasing the timestamp isn't going to cause weird bugs for you and merge.


jaredcdavis:
Feel free to merge if you like.  However, the current value is as intended (it is the expected date for VL to exit alpha status.)

PR #424, opened 2015-06-29
From Russinoff: improvements to RTL Rel11

ragerdl:


ragerdl:
I hear that this will likely break the ACL2(r) build (where R is short for Real, as opposed to Rager or Russinoff ;).  Performing the three-way merge has been painful enough that I'm just going to let that build break, and then someone (probably Dima or Russinoff) can fix it.

PR #449, opened 2015-07-10
Added SSE/SSE2 instructions to x86isa project.

myall86:
Added SSE/SSE2 instructions to x86isa project.


ragerdl:
Nice!

I'm assuming Cuong has permission from @shigoel and merging into testing.

PR #450, opened 2015-07-11
New tarai-measure in coi/termination/complex

nadezhin:
The existing proof of `tarai-always-terminate` in `coi/termination/assuming/complex.lisp` didn't work in ACL2(r). It was problematic to extend the proof to ACL2(r) because the proof extracted numerators/denumerators of its arguments.

Instead, another `tarai-measure` was formulated. It is similar to `tarai-measure` from `coi/termination/assuming/zero.lisp`, but `m2` and `m3` convert their results to integer by `fl-ext` and `m4` was added to the measure list.

Now both standard ACL2 and ACL2(r) certifies `complex.lisp` in 5 seconds. So it is not necessary to keep this file in `SLOW_BOOKS` list and to keep special `complex.acl2`.


ragerdl:
Wow, this is a really nice improvement -- from however many seconds to 5 seconds.  Assuming this is truly the case, nice job @nadezhin !

I'm going to merge into testing and provided it passes, it'll be merged into master.  If it turns out that it still is a slow book in a different environment, we can deal with that then.

PR #456, opened 2015-07-15
Quotient remainder lemmas

nadezhin:
There are incompatible definitions of `realp-mod` in ACL2(r).
This hinders simultaneous include of `ihs` and `arithmetic-5` in ACL2(r).
This pull requests modifies the definition in `ihs`.


ragerdl:
This looks like an improvement to me based on my limited knowledge of ACL2(r).  If anyone else has a stronger opinion, we can revert it.

@myall86 @russinoff These change lists by Dmitry might be interesting as ways to learn about the presence of ACL2(r) principles in the libraries.

PR #458, opened 2015-07-16
RTL DIV demo: Remove usage of old-style rounding modes

nadezhin:
Previously name of rounding functions were `trunc`, `away`, `near` ... .
Now they are `rtz`, `rta`, `rte` ... .
RTL division demo was written in old names .
These names were not changed by new ones.
Instead glue lemmas were written which proved new rounding-mode formulation of division theorems using old rounding-mode formulation.

This patch renames old rounding modes to new rounding mode in the main demo code and then removes glue code.


ragerdl:
This seems good to me.  ~1400 lines of code removed and 500 added.  Given these changes are only in the support directory, this is probably very good progress for the health of the support libraries.

Since some of these changes are a bit strange, @russinoff , could you please state your approval, and then I'll merge it?

ragerdl:
@russinoff has approved these changes.  Thanks @nadezhin !

PR #465, opened 2015-07-26
Combine rel9-rtl-pkg/*.mult files and move the result to rel11/support

nadezhin:


ragerdl:
@nadezhin This is an impressive list of changes.  It looks like you slimmed down ~7.7k LoC to 1.2k LoC -- nice!  This is within the spec of changes agreed upon with @russinoff , so I am merging it.

PR #467, opened 2015-07-28
Create rtl/support/definitions.lisp and use it

nadezhin:
Copy public definitions from rtl books `basic`, `bits`, `float`, `round`, `reps` into new support file `rtl/rel11/support/definitions.lisp`. Measure and guard proofs are also there.

Now client modules in support directory can include non-locally `definitions` book and include locally books with necessary lemmas.

Combine `gl.*` and `masc.*` and move them from `rtl/rel11/rel9-rtl-pkg` to `rtl/rel11/support`.

I still have a question. Do we need a file `rtl/rel11/support/gl.acl2` ? ACL2 certifies `gl.lisp` without it.


ragerdl:
Re: `Now client modules in support directory can include non-locally definitions book and include locally books with necessary lemmas.`  Very nice!  This is probably particularly good now that RTL is in its own package (since including extra function definitions won't pollute the ACL2 namespace).

ragerdl:
@russinoff Is this design change of putting all definitions in one book okay with you?

russinoff:
I need to think some more about this, and maybe understand better what is
being suggested.

ragerdl:
Hi David,

Sure, give me a call if that would help.

David
(512) 401-1016

nadezhin:
David,

My suggestion is to combine numerous files from `rtl/rel11/rel9-rtl-pkg` into a few `rtl/rel11/support` files. I want to make `support` self-sufficient and smaller. After this I would like to make `support` certifiable by ACL2(r).

This pull request converts three topics: `gl.lisp`, `masc.lisp`, `excps.lisp`.

The converted files included non-locally lower modules like  `basic`, `bits`, because formulation of theorems need definitions of functions like `fl`, `bits`. All theorems from lower modules also were included. I tried to change non-local include-book to local include-book by moving definitions into a separate file.

However, non-local include-book of `basic.lisp`, `bits.lisp` into support books is not a trouble. If you dislike style with `definitions.lisp`, I can rework this change. It would convert `gl.lisp`, `masc.lisp`, `excps.lisp` without creation of `definitions.lisp`.

russinoff:
I think this is a good idea.  Please proceed.

nadezhin:
It is in the branch other than the branch of the pull request:
https://github.com/nadezhin/acl2/tree/rtl-without-definitions

PR #469, opened 2015-07-28
Updates to centaur/bitops/fast-rotate and build/make_cert_help.pl

shigoel:
Removed dependency on arithmetic-5 for centaur/bitops/fast-rotate.lisp. Also pushing in Jared's edits to build/make_cert_help.pl to support extracting PBS directives from .acl2 files.


ragerdl:
Thanks for removing the arith-5 dep.

PR #475, opened 2015-08-06
Fixing a typo

ragerdl:
@solswords Can you please confirm that this was indeed a typo and merge?  It looks to me like `:boolvars` is the right argument to `svtv-run`, not `:boolmasks`, and I'm guessing this documentation just got outdated as you cleaned up the interface.


ragerdl:
It looks like boolvars didn't fix the problem about ```ERROR: some bits assumed to be Boolean were not```, so maybe this suggestion is wrong.

solswords:
Setting :boolvars nil on the svtv-run should prevent that error; if not, maybe you can isolate the problem & make another bugreport?

PR #482, opened 2015-08-12
Added some lemmas, mainly to "rtl/rel11/lib/round".

ragerdl:
@nadezhin Could you have a look at this pull request, which is originally from Russinoff and let me know what you think?  My general inclination is to auto-accept RTL changes from Russinoff, but it looks to me like there's some backwards progress included in this patch (which suggests something went awry with the merge).  For example, lots of the trailing whitespaces are back.

@russinoff I made this new pull request so that the pull request will be submitted to the testing branch, ensuring that we won't break the master branch without at least testing it on the leeroy.defthm.com server first.


russinoff:
I think of the lib directory as being pretty much off limits to everyone
but me, so I just copied my files into that directory.  I accounted for
Dmitry's recent changes in masc.lisp, but not the whitespace thing.  If you
like, I can try to fix that and issue a new request.

ragerdl:
Dmitry and I spoke, and the only two reversions that we see are the return of whitespaces and the commenting of the `defsection-rtl`'s.  Based on that, I'll go ahead and merge it.  It'd be good to remove the whitespaces.  Dmitry also might have a suggestion or two.

russinoff:
In the last pull request, I uncommented the "defsection"s in reps.lisp and
removed the trailing whitespace.  At least there is no occurrence of the
string ")  " in lib/*lisp.

nadezhin:
David @russinoff ,
I suggest to disable executive counterparts of `(sp)`, `(dp)`, `(ep)` in `lib/defs.lisp`.

russinoff:
Yes, I should have done that.  Should I make yet another pull request?

russinoff:
OK, I did that in my repository.  By the way, you see that I borrowed your
idea of making a separate book of definitions.

nadezhin:
> OK, I did that in my repository.
Yes. And now the pull request contains two commits.

>idea of making a separate book of definitions.
Yes. I like it.

nadezhin:
By the way, support/reps.lisp and support\round.lisp also contain trailing
whitespaces.

On Wed, Aug 12, 2015 at 10:27 PM, Dmitry Nadezhin <dmitry.nadezhin@gmail.com

russinoff:
Oh, I thought it was just in lib that you cared about whitespace.  I'll fix
those as well.

russinoff:
I found one trailing space in support/reps.lisp, but nothing in
support/round.lisp.

nadezhin:
It is at line 10 of support/round.lisp just after `reduce-additive-constant-<`, but I don't think that this space deserves a new pull request.

Can you explain why do you split "round" into "round" and "old-round" ?

russinoff:
I wanted to add some lemmas to lib/round.lisp that I proved to support a
recent verification project.  These lemmas were proved in the context of
the existing lib.  If I just stuck them in support/round.lisp, the proofs
would not go through because the environment woyld be different; it woyuld
take significant work to prove them in that environment.  This is
essentially the same method that we've used in the past to address this
problem, hence lib<m>.delta<n>, etc.

PR #486, opened 2015-08-13
Adding conversion from 4vec, resp. 4veclist, to string, resp. list of…

slobodova:
… strings of 0s, 1s, Xs, and Zs.


ragerdl:
Nice!  I'll let @solswords or @jaredcdavis merge it since it's their baby.

PR #491, opened 2015-08-17
 RTL: Combine reps and rom-helpers, bad formats handling, chop-r lemmas

nadezhin:
This is a batch of rtl changes:
Combine `reps` and `spd-mult` without using `rel9-rtl-pkg`;
Lemmas about `chop-r` similar to new lemmas about `chop`;
`(encodingp x f)`, `(nrepp x f)`, `(drepp x f)` are false when `f` is not good fromat,
remove hyps in many lemmas;
two theorems about classification of floating-point data;
Combine `rom-helpers` for `rel9-rtl-pkg` to `rel11/support` .


ragerdl:
@russinoff There are some non-trivial changes to `books/rtl/rel11/lib/defs.lisp` and `books/rtl/rel11/lib/reps.lisp`.  Are these okay with you?

ragerdl:
@russinoff I'm still trying to get to your masc stuff, but does this patch look good?  See above comment about the two files in lib.

russinoff:
This is OK with me.

PR #500, opened 2015-08-29
ACL2 can certify now one of the demos without including "arithmetic"

nadezhin:
Remove `(include-book "arithmetic/top-with-meta" :dir :system)` from `demos/tutorial-problems/equivalence-of-two-functions/kaufmann.lisp`.

### DIFF
--- a/books/demos/tutorial-problems/equivalence-of-two-functions/kaufmann.lisp
+++ b/books/demos/tutorial-problems/equivalence-of-two-functions/kaufmann.lisp
@@ -36,10 +36,7 @@
 ; Well, reading the definition, I can see that (sum2 i j) adds the numbers from
 ; (j - i) up to j, at least if i <= j.  That's the same as adding all the
 ; numbers from 0 to j, and then subtracting the numbers from 0 to (j - i - 1).
-; That observation is realized in the next lemma, which failed to prove until I
-; loaded an arithmetic library, as follows.
-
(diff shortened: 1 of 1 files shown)


MattKaufmann:
Cool!  I wonder why I thought I needed that lemma -- maybe ACL2 has
improved since the time I developed that book....

-- Matt

MattKaufmann:
P.S. Wow -- turns out that ACL2 6.5 back through 5.0 (I didn't test
before that) need the include book, but ACL2 7.0 and the current git
snapshot (and presumably everything inbetween) do not.  So something
changed in 7.0.  It wasn't tau -- 7.0 accepts the events without the
include-book even after (in-theory (disable (tau-system))).

-- Matt

nadezhin:
Can it be caused by pushing `commutativity-2-of-+`, `fold-consts-in-+`, `distributivity-of-minus-over-+` into `axioms.lisp` ?

MattKaufmann:
Yep, that's it!  Indeed, if I first execute

(IN-THEORY (DISABLE FOLD-CONSTS-IN-+))

then that defeats the proof in ACL2 7.0 and the current github version
(and probably everything inbetween).

Thanks --
-- Matt

ragerdl:
Based on the conversation, it sounds like Matt is good with this patch.  So, I'm merging it.  Please let me know if I misunderstand.

PR #509, opened 2015-09-10
Moved misc/dead-events.lisp to tools/dead-events.lisp

shigoel:


ragerdl:
Thanks for using the reloc_stub methodology.

PR #514, opened 2015-09-17
Update to fix certification problem.

pennyan:
I did the regression-fresh run and fixed a bug in config-template.lisp file that prohibits successful certification. I also did other minor changes.


ragerdl:
Thank you for fixing the build and for submitting your patch against `testing`.  This should go through to the `master` branch soon.

PR #520, opened 2015-09-29
adding supporting material for acl2 2015 workshop paper on skiping re…

miteshjain:
…finement

This has three set of experiments reported in the paper. 
1. JVM-inspired stack machine
2. Optimized memory controller
3. Vectorizing compiler


ragerdl:
Thanks for making this request against acl2:testing and commenting out the expensive theorem (shown in file ____3.lisp).  Merging into testing -- it will be in master after the regression builds successfully with these files in them.

PR #521, opened 2015-09-30
defung mv/stobj support

TheBeaNerd:


ragerdl:
Thanks for pulling against acl2:testing!  Since the author is basically white-listed for coi changes, I'm merging into testing without looking at it in much depth.

ragerdl:
@TheBeaNerd It looks like these changes broke the testing build.  Can you please fix?

PR #523, opened 2015-10-06
Added 'passing' make target to cosims

ragerdl:
Hi @solswords or @jaredcdavis ,

Could you please look at these changes and approve/comment?  I won't accept the pull request until after the "for" cosim tests are added, but I wanted to get your feedback before I was ready to be finished with this.

Thanks,
David

### DIFF
--- a/books/centaur/sv/cosims/Makefile
+++ b/books/centaur/sv/cosims/Makefile
@@ -66,12 +66,19 @@ all: cosim-core symlinks
 
 # ------------------------------------------
 #  Test setup.
-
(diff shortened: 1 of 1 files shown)


solswords:
Seems fine, I guess -- I'm not sure what deleting simv.daidirs does or why I haven't needed to do it.

ragerdl:
One may need to clean simv.daidirs if you change ncverilog versions, because otherwise the files show up as corrupt.  There are a lot of ncv/vcs simulation files that could use cleaning, but I'm not going to worry about the others for now.

PR #524, opened 2015-10-07
Added word-count and array-copy correctness proofs

shigoel:
Added the proofs of correctness of the word-count program (ref.: FMCAD'14 paper) and a small array copy sub-routine. Also sped some books up. However, the newly added book of word-count correctness takes a long time (~6m with vanilla ACL2 on a 2.7 GHz Intel Core i7, 16GB RAM) to certify, so probably no net speed up on the x86isa books as a whole will be observed.


ragerdl:
Congrats on the progress!  Merging.

PR #527, opened 2015-10-14
Jvm overload

nadezhin:
Method lookup in JVM searches method by class name, method name and method descriptor.
JVM M5 model searched method by class name and method name only. This pull request replaces method names by method name-and-types in all class-decls of JVM M5 which I found in acl2-books.


nadezhin:
I close this pull request because I accidently created it against master branch instead of testing branch.

PR #528, opened 2015-10-14
Overloaded methods in JVM M5

nadezhin:
Method lookup in JVM searches method by class name, method name and method descriptor.
JVM M5 model searched method by class name and method name only. This pull request replaces method names by method name-and-types in all class-decls of JVM M5 which I found in acl2-books.


ragerdl:
I've sent an email to J requesting that he look at this.  

I'm trying to decide whether J should always look at these changes before they're made.  I think Dmitry and I are glad he looked at the changes from a couple weeks ago, because he found a couple typos, and it's harder to catch bugs once they're actually in the codebase.  But, perhaps he'd rather not be a bottleneck.  Maybe E Smith or H Bing could help out with this?  I don't know....

nadezhin:
I cancel this pull request because Oracle lawyers didn't approve it.

PR #529, opened 2015-10-16
Adding 4vec utility functions

ragerdl:
@solswords Does this seem okay to you?

### DIFF
--- a/books/centaur/sv/svex/4vec-base.lisp
+++ b/books/centaur/sv/svex/4vec-base.lisp
@@ -488,3 +488,20 @@ fast), if so executes @('2vec-body'), and otherwise executes
                   (<= 0 (4vec->lower x))))
     :rule-classes :forward-chaining)
   (deffixequiv 4vec-index-p))
+
+(defmacro 4vec-size-p (x &key size)
+  `(and (4vec-p ,x)
+        (or (unsigned-byte-p ,size ,x)
+            (and (consp ,x)
+                 (<= (- ,size) (car ,x))
+                 (<= (car ,x) ,size)
+                 (<= (- ,size) (cdr ,x))
+                 (<= (cdr ,x) ,size)))))
+
+(defmacro 4vec-non-z-p (x)
+  `(or (atom ,x)
+       (equal (cdr ,x) 0)))
+
+(defmacro 4vec-non-x-p (x)
+  `(or (atom ,x)
+       (equal (car ,x) 0)))


solswords:
4vec-size-p uses size in a manner that doesn't make sense to me -- its use in unsigned-byte-p suggests a width but then it's used as a positive/negative bound otherwise?

Also, the definitions break the 4vec-p abstraction unnecessarily, by using car/cdr/atom instead of 4vec->upper/lower (and 2vec-p, although that might not be necessary).  Not only that but since they're macros, that means wherever they're used they cause breakage of the 4vec abstraction (functions could at least be disabled and become part of the abstraction).

ragerdl:
Doh, it's supposed to be `signed-byte-p`, but it wouldn't surprise me if there are still issues even after changing that and using lower/upper.  I'll do some examples by hand -- thanks for the feedback.

ragerdl:
The macro vs function still needs to be fixed, but that aside, do these definitions seem more inline?

```
(defmacro 4vec-size-p (x &key size)
  `(and (4vec-p ,x)
        (or (unsigned-byte-p ,size ,x)
            (and (not (2vec-p ,x))
                 (unsigned-byte-p ,size (4vec->lower ,x))
                 (unsigned-byte-p ,size (4vec->upper ,x))))))

(defmacro 4vec-non-x-p (x)
  `(or (2vec-p ,x)
       (equal (4vec->upper ,x) 0)))

(defmacro 4vec-non-z-p (x)
  `(or (2vec-p ,x)
       (equal (4vec->lower ,x) 0)))
```

ragerdl:
I'm currently having trouble understanding why a upper or lower can be negative and still fit in a given size, since negative values yield either infinite X's, Z's, or 1's.  So, my current definition is based off the idea that it can't be.

ragerdl:
Closing for now.  Not merging.

pennyan:
Hi, I revisited this problem and thought maybe we can write it as below:

(defun 4vec-size-p-fn (x size)
  (and (sv::4vec-p x)
       (or (unsigned-byte-p size x)
           (and (not (sv::2vec-p x))
                (signed-byte-p size (sv::4vec->lower x))
                (signed-byte-p size (sv::4vec->upper x))))))

;; not sure if we want this macro or not, :SIZE key seems nice :)
(defmacro 4vec-size-p (x &key size)
  `(4vec-size-p-fn ,x ,size))

For 4vec-non-z-p, it seems that 3vec will do the correct job of defining a 4vec that excludes Z's.

But then when I tried to prove a shape lemma, the theorem failed with this subgoal:

(IMPLIES (AND (INTEGERP IN)
              (NOT (EQUAL (SV::4VEC->UPPER IN)
                          (SV::4VEC->LOWER IN)))
              (<= -1 (SV::4VEC->LOWER IN))
              (< (SV::4VEC->LOWER IN) 1)
              (<= -1 (SV::4VEC->UPPER IN))
              (< (SV::4VEC->UPPER IN) 1))
         (<= -2 IN))

This subgoal seems obviously right. If (integerp in), then (sv::4vec-upper in) and (sv::4vec-lower in) should both return just "in". Then the antecedent is just false and we should conclude the theorem. But ACL2 is not doing it. Why is that?

solswords:
The basic answer to your question is that you should prove a rewrite rule
outside of your GL proof and then adjust the keywords to your def-gl-thm to
allow that rule to work.  You can use :do-not-expand to tell GL not to open
up a particular function when doing a coverage proof (keeping it disabled
is insufficient), and you can use :cov-theory-add to add rules that can be
used during coverage.  I'd suggest
(implies (signed-byte-p size x) (4vec-size-p x size))
and
(implies (and (consp x) (signed-byte-p size (car x)) (signed-byte-p size
(cdr x)))
             (4vec-size-p x size))

then use :do-not-expand '(4vec-size-p) and :cov-theory-add
'(4vec-size-p-rule1 4vec-size-p-rule2).

But further, I really don't like (or undesrstand) that definition of
4vec-size-p. First, it doesn't make any sense to ask unsigned-byte-p in one
case and signed-byte-p in the other.  Also, it's breaking the 4vec
abstraction; this is a fine thing to do for execution performance but the
logical story of such a function should really just be [un]signed-byte-p of
4vec->upper and 4vec->lower.  For super-optimized execution performance you
shouldn't even need to break the abstraction that much -- I think the
following is an appropriate definition:
(mbe :logic (and (signed-byte-p (4vec->upper x))
                 (signed-byte-p (4vec->lower x)))
     :exec (if (2vec-p x)
               (signed-byte-p (2vec->val x))
            (and (signed-byte-p (4vec->upper x))
                 (signed-byte-p (4vec->lower x))))

Note in order to prove the two (implies ... (4vec-size-p ...)) theorems
above, you'll have to let ACL2 break the 4vec abstraction -- this probably
means enabling 4vec->upper, 4vec->lower, 4vec-p, and maybe more.  But once
you prove those theorems (and disable them, because normally you don't want
to break the abstraction the way those theorems do), GL should only need
them (or something like them) to do its coverage proofs.

- Sol

PR #533, opened 2015-10-24
nary modular arith support

TheBeaNerd:


ragerdl:
I didn't look at in detail, but since it's coi and thebeanerd, I assume it's good.

PR #537, opened 2015-11-03
Tuned the hypotheses for occurrences-llvm.lisp

david-s-hardin:
Slightly tuned hypotheses for occurrences-llvm.lisp


ragerdl:
Nice improvements!

PR #544, opened 2015-11-13
Lists of symbols are lists of pseudo-terms.

acoglio:
Based on Jared's suggestions on the mailing list, I'm adding this theorem, enabled by default.

In regard to the question whether it should be enabled or disabled by default, I noticed that nat-listp-when-unsigned-byte-listp is enabled by default, so I'm leaving this one enabled by default as well. But if people think that it should be disabled, let me know and I'll do that (and the pull request will get updated automatically as I push a new commit).


ragerdl:
This LGTM, but I'll let @jaredcdavis handle this.

PR #545, opened 2015-11-22
I86isa

nadezhin:
Specification of some SSE move instructions:
MOVSS/MOVSD
MOVUPS/MOVUPD
MOVAPS/MOVAPD


ragerdl:
Interesting Dmitry!  @shigoel , this is yours to manage, so please let me know your thoughts on this pull request.  Feel free to place them here or send me a message offline.  Thanks!

shigoel:
@nadezhin Thanks for adding the FP MOV instructions and for correcting the entry of the immediate OR opcode in the "implemented-opcodes-table". I'm excited that you're looking into the x86isa books and maybe using them!

It turns out that @myall86 contributed similar changes to the x86isa books --- in fact, he has contributed many FP instructions (including bitscans, shuffles, converts, etc.), which are still in my local branch. Is it alright with you if I push my local branch in a few days?  I need to get some other things working before I push. I aim to ensure that our implementations "match" in some reasonable sense.

Meanwhile, if you need a list of instructions that I have in my local branch (in order to avoid duplication of effort), let me know. I can send them on to you.

Thanks again!

nadezhin:
@shigoel . Thank you for the answer. I describe my topic here. Developers
in OpenJDK are rewriting math library fdlibm from C to Java. They want to
be sure that they preserve bit-exact equivalence between old and new
implementations. I want to prove the equivalence between C and Java
implementations mechanically.
As there are no ACL2 models of C and Java languages, I have to compare
object files at a particular machine and JVM classfiles. The particular
machine is x86-64, of course.
 I expect that it is not very difficult because:
1) Optimisation in C compiler were suppressed because some optimizations
are not equivalent in floating-point world: (1 + x) - 1 is not the same as
x.
2) Developers of Java created intermediate class FdlibmTranslit
 which is line-by-line transliteration of C to Java.
So I hope that equivalence between Fdlibm object files and FdlibmTranslit
classfiles can be verified almost mechanically.

I explore mechanical approach on cubic root function cbrt(x) now. I attach
its code in cbrt-code.lisp . I still need addsd, mulsd, divsd instructions.
I noticed @myall86 's implementation of these instruction in the file
projects/x86isa/machine/instructions/floating-point.lisp in the main ACL2
branch.
So the work is almost done. However, this book uses rtl/rel9 version of
Russinoff's library. I woould request @myall86 to migrate to latest
rtl/rel11 version.
This version contains a new excps book . It implements specials and
exceptions, so the code in floating-point.lisp could be simplified. If
Cuong has no time for this, I would do it. Meanwhile I'm working on adding
lacking guards to functions in rtl/rel11 book for faster execution.

I also attach javamath.tar.gz with assembler listing of all fdlibm
functions, though I haven't scanned it for lacking x86 instructions yet.

myall86:
@nadezhin Attached contains the list of SSE/SSE2 instructions I already implemented. They are not in the main branch, but I believe @shigoel will release them at some point. Regarding the change in RTL library from rel9 to rel11, I'm not working on the x86isa project; so if you can handle that and @shigoel is fine with that, please go ahead.

Thanks,
Cuong

[SSE-SSE2.txt](https://github.com/acl2/acl2/files/44519/SSE-SSE2.txt)

shigoel:
@myall86 Thanks, Cuong, for the list!

@nadezhin Thanks for the description of your project! It sounds exciting and certainly do-able with the x86isa books once the FP instructions are in.

This is just a progress update from my side --- I'm currently working on porting over Cuong's work to the "new" style of decoding (so that it uses all the stuff that's in x86isa/utils/decoding-utils.lisp). When Cuong had made his contribution, we needed to do all of that manually. So, in maybe a couple of days or so, I'll push my changes to the main branch of ACL2. 

It would be good to switch over to rtl/rel11, but I wouldn't get the chance to do it myself in the next few weeks. You're welcome to open a pull request if you switch the x86isa books from using rel9 to rel11, definitely (and I'd be grateful for your work!). Also, if you're working on adding guards to the rel11 books, maybe you'd also be willing to add type declarations to further increase the execution speed.

Thanks! I'll write back when I have more to report.

nadezhin:
I close this pull request in favour of Shilpi's pull request #549 .

nadezhin:
@shigoel <https://github.com/shigoel>
Thank you for your commit. I will try the updated x86isa .
Here is a reminder about MOVAP command. You might wish to place it in your
TODO list.

MOVAP xmm2/m128, xmm1 command uses
function x86-operand-to-xmm/mem to write to memory.
This function checks alignment.

However MOVAP xmm1, xmm2/m128 command uses
function x86-operand-from-modr/m-and-sib-bytes to read from memory.
This function doesn't check alignment yet, but it should.

I think that type declaration will not enhance performance of rtl exps.lisp
functions much.
It converts floating-point operands to rationals and compute operation on
rationals.
For example 0x1.0p-1023 will be a rational with numerator 1 and denominator
2^1023 . The slowness is here.
Perhaps it is worth to implement floating-point operations in terms of
integer exponent and significand instead of rationals
for better performans. But it is TODO, not now.
<https://github.com/shigoel>
​

PR #546, opened 2015-11-23
Add theorems about ACL2-COUNT of NTH.

acoglio:
I saw the comment in std/lists/acl2-count.lisp about whether theorems like these should go into that file or into the files for the various list functions. Perhaps they could go into the files for the various list functions, as is already the case in nthcdr.lisp. So this is why I added these to nth.lisp.

### DIFF
--- a/books/std/lists/nth.lisp
+++ b/books/std/lists/nth.lisp
@@ -323,4 +323,22 @@
 (defmacro equal-by-nths-hint ()
   '(equal-by-nths-hint-fn clause))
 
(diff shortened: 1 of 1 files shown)


ragerdl:
This one is yours: @jaredcdavis

jaredcdavis:
Hi,

Sorry, I'm in Omaha and can't look at anything until next Tuesday, please
feel free to merge.

Cheers,
Jared

ragerdl:
@jaredcdavis Could you please have a look and merge?

PR #549, opened 2015-12-01
New FP instructions and libraries for reasoning about equivalence of x86 states

shigoel:
This bunch of commits adds some new FP instructions and libraries to
reason about paging data structures. A description of the changes is
as follows.

----------------------------------------------------------------------
Re-organized the code proof libraries

Separated out proof libraries in proofs/utilities into two directories
--- programmer-level-mode and system-level-mode.

----------------------------------------------------------------------
New proof libraries to reason about paging data structures

An equivalence relation between two x86 states, called
xlate-equiv-x86s, has been defined. This has been used to prove, using
congruence-based reasoning, that a walk of the paging data structures
does not affect any future walks. See books in 
proofs/utilities/system-level-mode/paging-lib for details.

Alternate definitions of paging data structure traversal functions
(originally defined in machine/x86-ia32e-paging.lisp) have been
defined in x86-ia32e-paging-alt.lisp. These new functions do a paging
structure walk only if the structure is set up in a reasonable manner.

Books gather-paging-structures.lisp and
gather-paging-structures-thms.lisp define the equivalence relation
xlate-equiv-x86s and functions to gather all the paging data
structures.

Books paging-page-table-lemmas.lisp,
paging-page-directory-lemmas.lisp,
paging-page-dir-ptr-table-lemmas.lisp, and
paging-pml4-table-lemmas.lisp include the theorems about independence
of a paging structure walk from a past walk.

----------------------------------------------------------------------
Modified the top-level definitions of memory accessor and updater
functions

Affected functions are rm16, wm16, rm32, wm32, rm64, wm64, rm128,
wm128, write-canonical-address-to-memory.

The programmer-level mode behavior of these functions is the same as
it was before, but these functions have changed in the system-level
mode. We describe this change using the function rm16 as an example.

Previously, for (rm16 lin-addr r-w-x x86), we would figure out if both
lin-addr and (1+ lin-addr) fall in the same physical page. If so, we
would walk the page tables to translate the lin-addr to its
corresponding physical address phy-addr, and the physical address
corresponding to (1+ lin-addr) would simply be (1+ phy-addr).  This
helped avoid a page table walk for the translation of (1+ lin-addr)
and thus, increased the execution efficiency in the common case.

Now, we explicitly perform the translation for both lin-addr and (1+
lin-addr). This has definitely slowed down our simulation speed (but
only in the system-level mode). However, our optimizations were not
strictly correct. Because the second page walk was optimized away, our
functions essentially said that the second page walk never modifies
the x86 state, when this is not the case because a page walk may
modify the accessed and/or dirty bits in the translation-governing
entries of the paging data structures.

With this change, we lose on the execution efficiency in the
system-level mode, but gain accuracy.

Simulation speed in the System-Level Mode with identity page tables
(1GB pages):

  Old Model: Around 900K instructions/second
  New Model: Around 795K instructions/second

We can imagine "correctly" optimizing away the page walks by
performing the accessed/dirty bit updates without actually walking the
paging data structures, but that optimization will complicate
reasoning and thus, we haven't implemented it for now.

----------------------------------------------------------------------
Decreased the initial memory footprint of the model

I changed the constant *initial-mem-array-pages* in
portcullis/sharp-dot-constants.lisp from 10 to 2.  Therefore, the
initial memory requirement for mem-array went down from 1.2 GB to 256
MB. If more initial memory is needed, simply change the value of
*initial-mem-array-pages* --- all the proofs should proceed smoothly.

Also sped up machine/x86-concrete-memory.lisp by reducing the number
of local include-books.

----------------------------------------------------------------------
Ported over FP instructions added by @myall86  --- thanks a lot,
Cuong, for specifying so many instructions!

@nadezhin and @myall86  independently contributed the
specifications of MOVSS/SD and CMPSS/SD instructions. This set of
commits incorporates both their contributions.

The only significant change that I made here was for the BSF
instruction (opcode 0F BC) --- I specified the undefined behavior for
the destination and flags, as appropriate.

I extended utils/decoding-utilities.lisp and machine/x86.lisp to
support decoding of instructions requiring mandatory prefixes in the
two-byte opcode map.

NOTE: @myall86 and I will probably make at least a couple of bug fixes
(mostly in the FP flag computations but maybe in other places too) in
the next few days.

----------------------------------------------------------------------
@nadezhin pointed out typos in the :reg field of immediate OR
(opcode 80) in the implemented-opcodes-table --- thanks, Dmitry! This
set of commits fixes those typos.

----------------------------------------------------------------------


ragerdl:
There's some really great stuff in here.  Congrats on getting this organized and done!

PR #552, opened 2015-12-02
Fixed two bugs in rtl/rel11/lib/excps

russinoff:


ragerdl:
@russinoff Did you mean to commit the changes to make_cert_help.pl?  If the regression builds in your environment without that change, it's probably better to leave the version unspecified.

ragerdl:
In case it was accidental, to undo the change, change the code back to `use File::Path qw(make_path);`, commit it locally, push it to your repo, and then let me know that you've updated your repo.  You won't need to create a new pull request for that reason.

russinoff:
DR, I left that change because of Jared's observations (reproduced below.),
but I don't care if you leave it out.

This might possibly be due to an old version of the
File::Path Perl module. To investigate whether this
is the case, you might try changing:

books/build/make_cert_help.pl

And replace line 67,

use File::Path qw(make_path);

With:

use File::Path 2.07 qw(make_path);

And then try rebuilding again. This might produce a
better error message, hopefully along the lines of:

File::Path version 2.07 required--this is only version 2.04

If that indeed works, then it might be good for us to make
this change to make_cert_help.pl and perhaps other scripts
that are using File::Path.

ragerdl:
Hi David,

Unless you express that you need that change in order for you to build the
books (it sounds like this particular change didn't matter, since your perl
version was 1 point something), I'll change it back after merging your pull
request.

David

russinoff:
OK.

PR #553, opened 2015-12-02
Fixed two bugs in rtl/rel11/lib/excps

ragerdl:
Submitting against the testing branch.


ragerdl:
Merging.  See discussion in https://github.com/acl2/acl2/pull/552

PR #555, opened 2015-12-11
Added define-macros flag to defsvtv

ragerdl:
@jaredcdavis or @solswords Could you please have a look and verify?


solswords:
Looks good, thanks.  At some point it'd be nice to add back in the ones that just take an alist, but since at the moment they call the non-alist versions, this seems fine for now.

ragerdl:
Thanks for the advice and merging.

PR #556, opened 2015-12-17
Misc. paging, memory alignment, and FP specification related changes

shigoel:
1. Paging: Sped up proofs about equivalence of states after paging
   stucture traversals and added preliminary work towards proving
   paging WoW lemmas.

2. Alignment: Added alignment check to the functions that read and
   write operands.  Thanks to Dmitry Nadezhin for requesting
   this. Checking unaligned memory accesses was unimplemented before,
   but as a result of this addition, I've attempted to do alignment
   checks when appropriate. There might still be cases where I've not
   accounted for them; if you encounter them, please let me know.

   Also note that though we do detect and report some erroneous
   conditions, we don't use IDTs, segments, etc. to handle these
   conditions yet; exceptions and interrupts in their entirety are
   still unimplemented.

3. FP specifications: The specification of some legacy SSE/SSE2
   instructions is now taken from rtl/rel11.

   In addition to FP rounding and other basic functions, we rely on
   rtl/rel11 for the specifications of addition, multiplication,
   subtraction, division, and square root operations.  Home-grown FP
   specifications include those of FP minimum, maximum, comparisons,
   and converts (they aren't yet present in rtl/rel11). Thanks a lot
   to Cuong Kim Chau for writing and testing the x86isa FP
   specifications!

   The x86isa books leverage years of work done on the rtl books,
   which are very well vetted. This decision also enables the
   developers of the x86isa books to focus on their current research
   goal --- developing capabilities to reason about system software
   --- without spending too much time on specifying and testing yet
   more x86 FP instructions. Supporting more instructions in the
   x86isa interpreter will almost always be a task that the developers
   will perform as needed, and we are hopeful that switching to
   rtl/rel11 for our FP specifications will allow others to execute
   and reason about their FP application programs on the x86isa
   interpreter.

   It would be simply wonderful if the specification functions in
   rtl/rel11 had guards and type declarations (and also no consing ---
   e.g., for formatp), both for ease of comprehension and for
   execution efficiency. But that's a task and gripe for another day.


shigoel:
@nadezhin You might find these changes to the FP specifications interesting. @myall86 and I now rely mostly on rtl/rel11 for most of them.

ragerdl:
Really nice write-up.  Congrats @shigoel !

Merging.

PR #557, opened 2015-12-18
Start directory for Kestrel Institute's contributions.

acoglio:


ragerdl:
If our directories become overwhelming, we can figure out a better directory scheme than centaur/oracle/kestrel.  But, this is fine for now.  Ideally we'd sort through this stuff, but I think Centaur thinks of the Centaur directory as "this is our stuff, you're welcome to use it, but we're also welcome to change it."  This is in contrast to std, which is reasonably stable.

Glad to see some more contributions on the way!

PR #558, opened 2015-12-22
Rtl guards

nadezhin:
Guards in `rtl/rel11/lib/excps.lisp` and in dependent books


nadezhin:
There are projects which respect guards and execute methods from RTL library:
x86isa, internal oracle project.
They have to wrap executable calls of RTL methods with ec-call .
https://github.com/acl2/acl2/blob/master/books/projects/x86isa/machine/instructions/fp/fp-base.lisp#L265
https://github.com/acl2/acl2/blob/master/books/projects/x86isa/machine/instructions/fp/fp-add-mul-base.lisp#L56

This is inelegant.
So I request to add guards in `excps.lisp` and in dependent books.
I tried to use `:guard t` where possible, but many functions requires more restrictive guards.

ragerdl:
Thanks, Dima.  I know @shigoel is looking forward to having more RTL guards.  @russinoff , could you please have a look and advise?

ragerdl:
We're putting off accepting this patch until after the new year.

@shigoel , do you have any comments as to whether having verified guards in RTL would benefit or harm the X86 project?

shigoel:
Now that the x86isa books depend on rtl/rel11 books for most of the FP specifications, verifying guards of rtl functions gives us the usual benefits. I'm sure the execution efficiency would improve. Plus, as Dmitry observed, having ec-calls around rtl functions in the x86isa books (where almost everything is guard verified) looks inelegant. Understanding the interfaces of rtl functions will be easier too. Thanks a lot, @nadezhin!

In fact, I'd go so far as to say that adding very strict type declarations to the rtl books would be awesome (if David Russinoff is okay with it, of course), but an issue as far as this is concerned is that rtl functions are very general. For example, we can't declare an intermediate result of rtl::sse-binary-spec to be <tt>(unsigned-byte 32)</tt> because this width depends on the floating-point format. A solution can be to define different heavily type-declared versions of rtl::sse-binary-spec (e.g., one each for single, expanded, and double precision), all logically equivalent to the general version. However, this sort of work is tedious (and maybe the results are not worth the effort for most people?) and I wouldn't want to wish this on anybody, especially since I can't attend to it myself in the near future.

russinoff:
Shilpi, let's talk. I'm on campus. Are you around?

shigoel:
I can be there at around 3pm. Does that suit you?

russinoff:
No. We could talk on the phone, but face-to-face is probably better. Don't
know when I can get back here. Can you visit Intel?

shigoel:
<took the conversation with David Russinoff offline>

nadezhin:
@russinoff , what is your decision about guards ?
This pull request will be outdated after accepting of #562 .
If can discuss with you how to rework it according to your style if you want.

books/rtl/rel11/lib/defs.lisp
```diff
@@ -141,10 +148,10 @@
     0))
 
 (defund exactp (x n)
+  (declare (xargs :guard (and (real/rationalp x) (integerp n))))
```
ragerdl: @nadezhin This is a style thing, but do you mind changing the conjuncts of guards to be on different lines?  We'd just like it to be consistent.  Thanks again for this patch!

nadezhin:
This pull request have been merged with #562 changes and cleaned.

books/rtl/rel11/lib/gl.lisp
```diff
@@ -9,44 +9,7 @@
...
-(defund binary-cat (x m y n)
-  (declare (xargs :guard (and (integerp x)
-                              (integerp y)
-                              (natp m)
-                              (natp n))))
-  (if (and (natp m) (natp n))
-      (+ (* (expt 2 n) (bits x (1- m) 0))
-         (bits y (1- n) 0))
-    0))
+(include-book "defs")
```
ragerdl: This is really nice cleanup, thanks!

books/rtl/rel11/support/bits.lisp
```diff
@@ -11,16 +11,13 @@
 ;;;				DVECP
 ;;;**********************************************************************
 
-(defund dvecp (x k b)
-  (declare (xargs :guard (and (natp k) (radixp b))))
-  (and (integerp x)
-       (<= 0 x)
-       (< x (expt b k))))
+; (defund dvecp (x k b) ... )
```
ragerdl: How come you left these definitions stubs commented with the "..."?

books/rtl/rel11/support/definitions.lisp
```diff
@@ -34,14 +36,18 @@
...
 
 ;; From bits.lisp:
 
-(defund dvecp (x k b)
-  (declare (xargs :guard (and (natp k) (radixp b))))
-  (and (integerp x)
-       (<= 0 x)
+(defnd dvecp (x k b)
+;  (declare (xargs :guard (and (natp k)
+;                              (radixp b))))
```
ragerdl: If appropriate, can you please remove these commented guards?

books/rtl/rel11/support/definitions.lisp
```diff
@@ -343,20 +392,21 @@
   (declare (xargs :guard (and (integerp x)
                               (integerp y)
                               (natp m)
-                              (natp n))))
+                              (natp n))
+                  :verify-guards nil))
```
ragerdl: Did you mean to leave this :verify-guards nil call in here?

books/rtl/rel11/support/definitions.lisp
```diff
@@ -408,15 +458,15 @@
 (in-theory (disable (:type-prescription binary-cat)))
 
 (defund mulcat-r (l n x b)
-  (declare (xargs :guard (and (integerp l) (< 0 l) (acl2-numberp n) (natp x)
-                              (radixp b))
+  (declare (xargs :guard (and (natp n)
+                              (dvecp x l b))
```
ragerdl: Nice!

ragerdl:
I've gone ahead and merged in this code.  I think it'll be easier to review another pull request if it turns out my questions result in changes.  Thanks for all of your hard work @nadezhin !  It's impressive that you're able to add this level of details to such a well-developed library that someone else authored.

PR #562, opened 2016-01-07
rtl/rel11/support: Radix-aware versions of functions in bits.lisp and…

nadezhin:
… float.lisp

Radix-aware version of almost each function in
rtl/rel11/support/bits.lisp and rtl/rel11/support/float.lisp .
These versions have an additional argument b. It is radix b >= 2.

One of new theorems is valid for prime radix now. A book euclid.lisp
is imported. Import in this book was patched to avoid circular dependency.

New functions and methods are not mirrored in rtl/rel11/lib .
Also they are not documented yet.

ACL2(r) certifies bits.lisp and float.lisp now. It also attemts to
certify books in curve25519. A trivial fix was necessary in one of these
books to avoid ACL2(r) certifcation failure.

### DIFF
--- a/books/projects/curve25519/ecp.lisp
+++ b/books/projects/curve25519/ecp.lisp
@@ -106,7 +106,7 @@
   :hints (("Goal" :use ((:instance mod*rewrite-2 (a (* a b)) (b c))))))
 
 (defthm mod*rewrite-4
-  (implies (and (integerp a) (integerp b) (integerp c))
+  (implies (and (integerp a) (integerp b) (integerp c) (integerp d))
            (equal (mod (* a b c (mod d (p))) (p))
                   (mod (* a b c d) (p))))
   :hints (("Goal" :use ((:instance mod*rewrite-2 (a (* a b c)) (b d))))))
@@ -124,7 +124,7 @@
   :hints (("Goal" :use ((:instance mod*rewrite-5 (b (* b c d)) (c e))))))
 
 (defthm mod*rewrite-7
-  (implies (and (integerp a) (integerp b) (integerp c) (integerp d))
+  (implies (and (integerp a) (integerp b) (integerp c) (integerp d) (integerp e))
            (equal (mod (* a b c d (mod e (p))) (p))
                   (mod (* a b c d e) (p))))
   :hints (("Goal" :use ((:instance mod*rewrite-2 (a (* a b c d)) (b e))))))
--- a/books/projects/quadratic-reciprocity/support/euclid.lisp
+++ b/books/projects/quadratic-reciprocity/support/euclid.lisp
@@ -6,8 +6,8 @@
 
 ;;   (2) If p is a prime divisor of a*b, then p divides either a or b.
 
-(include-book "rtl/rel11/lib/basic" :dir :system) ;; Properties of fl and mod
-(include-book "rtl/rel11/lib/util" :dir :system)  ;; Utility macros
+(include-book "rtl/rel11/support/basic" :dir :system) ;; Properties of fl and mod
+(include-book "rtl/rel11/support/util" :dir :system)  ;; Utility macros
 (local (include-book "arithmetic-5/top" :dir :system)) ;; It's hard to do any arithmetic without something like this
 
 ;; We first list some basic properties of divisibility.
--- a/books/rtl/rel11/support/basic.lisp
+++ b/books/rtl/rel11/support/basic.lisp
@@ -6,13 +6,6 @@
 (local (acl2::allow-arith5-help))
 (local (in-theory (acl2::enable-arith5)))
 
-;; The following lemmas from arithmetic-5 have given me trouble:
-#|
-(local (in-theory #!acl2(disable |(mod (+ x y) z) where (<= 0 z)| |(mod (+ x (- (mod a b))) y)| |(mod (mod x y) z)| |(mod (+ x (mod a b)) y)|
-                    simplify-products-gather-exponents-equal mod-cancel-*-const cancel-mod-+ reduce-additive-constant-<
-                    |(floor x 2)| |(equal x (if a b c))| |(equal (if a b c) x)|)))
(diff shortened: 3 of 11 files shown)


books/rtl/rel11/support/harrison.lisp
```diff
@@ -1,12 +1,14 @@
 (in-package "RTL")
 
+(include-book "tools/with-arith5-help" :dir :system)
```
ragerdl: Can this be local?

nadezhin:
This pull request conflicts with #558 .  When any of #558 or #562 will be accepted, I will merge the another one.

books/projects/quadratic-reciprocity/support/euclid.lisp
```diff
@@ -6,8 +6,8 @@
 
 ;;   (2) If p is a prime divisor of a*b, then p divides either a or b.
 
-(include-book "rtl/rel11/lib/basic" :dir :system) ;; Properties of fl and mod
-(include-book "rtl/rel11/lib/util" :dir :system)  ;; Utility macros
+(include-book "rtl/rel11/support/basic" :dir :system) ;; Properties of fl and mod
+(include-book "rtl/rel11/support/util" :dir :system)  ;; Utility macros
```
ragerdl: I can imagine that @russinoff prefers to import `lib`, rather than the `support` books.  How come the change?

books/projects/curve25519/ecp.lisp
```diff
@@ -106,7 +106,7 @@
   :hints (("Goal" :use ((:instance mod*rewrite-2 (a (* a b)) (b c))))))
 
 (defthm mod*rewrite-4
-  (implies (and (integerp a) (integerp b) (integerp c))
+  (implies (and (integerp a) (integerp b) (integerp c) (integerp d))
```
ragerdl: How come you added the extra hypothesis?  Typically extra hypotheses make a rule harder to use?  Or maybe we actually need it to be harder to use in this case, so that one branch of rewriting gets cut short?

books/projects/curve25519/ecp.lisp
```diff
@@ -124,7 +124,7 @@
   :hints (("Goal" :use ((:instance mod*rewrite-5 (b (* b c d)) (c e))))))
 
 (defthm mod*rewrite-7
-  (implies (and (integerp a) (integerp b) (integerp c) (integerp d))
+  (implies (and (integerp a) (integerp b) (integerp c) (integerp d) (integerp e))
```
ragerdl: How come you added the extra hypothesis here too?

books/rtl/rel11/support/basic.lisp
```diff
@@ -510,8 +503,13 @@
 ;;;                         CHOP-R
 ;;;**********************************************************************
 
-(defmacro radixp (r)
-  `(and (integerp ,r) (>= ,r 2)))
+(defnd radixp (b)
```
ragerdl: Changing this to a function seems like a nice improvement to me.

books/rtl/rel11/support/reps.lisp
```diff
@@ -819,7 +819,7 @@
   :prep-lemmas (
     (defrule lemma
       (implies (and (posp p)
-                    (real/rationalp x)
+                    (rationalp x)
```
ragerdl: Does this still work with ACL2(R)?

ragerdl:
Thanks @nadezhin ,   @russinoff What do you think?  Feel free to call me

books/projects/quadratic-reciprocity/support/euclid.lisp
nadezhin: A circular dependency will happen without this patch.
`rtl/rel11/suport/float.lisp` imports 
`project/quadratic-reciprocity/euclid.lisp`
It imports  `project/quadratic-reciprocity/support/euclid.lisp`.
It import `rtl/rel11/lib/basic.lisp`.
It imports `rtl/rel11/support/top.lisp`.
It imports `rtl/rel11/support/float.lisp`.
Loop.

books/projects/curve25519/ecp.lisp
nadezhin: ACL2(r) starts to certify this book after the change. It fails without this hypothesis.

books/projects/curve25519/ecp.lisp
nadezhin: The same.

books/rtl/rel11/support/reps.lisp
nadezhin: Lemma with real/rationalp passes in ACL2 byut fails in ACL2(r).
Lemma rationalp passes in both of them.

books/rtl/rel11/support/harrison.lisp
nadezhin: No. Certification fails at the end of file processing.
This include only builds table. It doesn't change theory.
Theory is changed later by local events.

ragerdl:
@nadezhin has addressed all of my questions.  What do you think @russinoff ?

russinoff:
I don't understand much of this, but I trust Dmitry not to do anything
crazy.

ragerdl:
@nadezhin What does it mean to be radix-aware?

nadezhin:
Radix-aware means that for each old function with fixed radix 2 there is a new function with additional argument radix b.  For example (defun bitn (x n) ...) ->  (defin digitn (x n b) ...) .

russinoff:
But this only happens in the support directory, right?

nadezhin:
Yes. Only in support directory.

ragerdl:
DMR has approved the admission of both #562 and #558.  Based on a conversation with Dmitry, it is easier for him if we accept #562 first, so we'll do that.  The cleanup for #558 will be relatively trivial, and I'll make sure that it doesn't surprise us.

Thus, I am merging #562.

PR #564, opened 2016-01-08
Utilities to query defun-sk's.

acoglio:
Utilities to recognize functions introduced via defun-sk and to retrieve their
defun-sk-specific constituents.


ragerdl:
Nice documentation.  We should probably start including some of these books in the manual.

MattKaufmann:
Absolutely!  Has anyone written a utility to find books that define
doc topics but aren't included (even indirectly) in
books/doc/top.lisp?  If not, would this be a good idea, and are there
any volunteers?

-- Matt

acoglio:
I believe that the documentation in these files that I've just committed get included into the ACL2+Books manual, because I had previously added an entry to include books/kestrel/top in books/doc/top. I hope that doing that was okay; I assumed so, since the pull request was accepted. The Kestrel Books documentation is directly under the software-verification topic, based on previous discussions with Matt.

ragerdl:
Hi Alessandro,

Ah yes, I had forgotten that you had added that entry.  Nicely done!

David

PR #570, opened 2016-01-13
Strengthened xlate-equiv-x86s and other misc. changes

shigoel:
This set of commits includes the following updates:

1. Redefined xlate-equiv-x86s: Now, two x86 states are xlate-equiv-x86s if their paging structures are equal, modulo the accessed and dirty bits (see xlate-equiv-structures); other memory locations not containing the paging structures must be exactly equal (see all-mem-except-paging-structures-equal); and, all other components of the two states, excluding MS and FAULT fields, must be exactly equal (see all-rgfs-equal, all-ctrs-equal, etc.). Note that the old definition of xlate-equiv-x86s was what the current definition of xlate-equiv-structures is.

2. Most functions that deal with address translation now check whether we are in the system-level mode or not; the guard of these functions also includes this check.

3. Converted some rewrite rules to congruence rules and heeded double-rewrite warnings.

4. Minor doc fixes.


MattKaufmann:
Hi, Shilpi --

I think it's really cool that you've found the double-rewrite warnings
to be useful!  I'm curious (but please only reply if it's easy) how
you decided which ones to act on.

Thanks --
-- Matt

shigoel:
@MattKaufmann I didn't really have a discipline for determining when to obey double-rewrite warnings. I did two things mainly. One, I created new congruence rules to avoid the warnings (as suggested in the the double-rewrite doc topic). Two, because initially I'd ignored all the double-rewrite warnings and painstakingly forced my rules to be used via use hints (I know, ugh), I figured that adding double-rewrites in the right places of these rules will make the use hints go away, and it did.

However, now whenever I encounter a double-rewrite warning involving equivalence relations that I defined myself and consider "complicated" in some sense, I usually heed the warning. So far, I haven't experienced any significant slowdown.

MattKaufmann:
Thanks, Shilpi.

Perhaps we could provide a way for users to label a ("complicated")
congruence so that double-rewrite is (a la Hardin) automatically done
in such cases, at least when stable under simplification.  Then maybe
we'd provide single-rewrite to override that behavior.

Maybe that's needlessly complicated though; we could discuss further
if you like.

-- Matt

PR #576, opened 2016-01-17
Move SOFT under Kestrel Books.

acoglio:
The previously existing files books/tools/soft.lisp and books/tools/soft-paper-examples.lisp now just point to (via include-book) to the appropriate files under books/kestrel/soft/. Those two deferencing files under books/tools can be removed eventually.
    
The license text has been aligned to the other files under the Kestrel Books.
    
There is now an initial XDOC page for SOFT, which links to the ACL2 Workshop paper that in fact documents all the SOFT macros. Nonetheless, the XDOC for SOFT will be expanded to document the SOFT macros without reference to the paper.


books/tools/soft-paper-examples.lisp
```diff
@@ -1,414 +1,3 @@
...
-; This file contains the SOFT ('Second-Order Functions and Theorems') examples
-; in the ACL2-2015 Workshop paper "Second-Order Functions and Theorems in ACL2".
-
-; Comments indicate the sections and subsections of the paper.
-
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-
 (in-package "ACL2")
 
-(include-book "soft")
```
ragerdl: It's unclear whether soft should be in kestrel or a more general place.  Anyway, I'm happy to have it moved to kestrel for now.  Do you mind creating another pull request that tags these as "relocation stubs"?  We didn't have this in a well-documented place, so I added it to the wiki.
https://github.com/acl2/acl2/wiki/How-to-move-a-book

Also, your CL highlights our thus-far-unstructured desire to come up with a structure for testing outside a normal build -- I think you're pretty close to what we want already. :)

ragerdl:
Thanks!

books/tools/soft-paper-examples.lisp
acoglio: Sure, done.

BTW, the formatting of the (in-package "ACL2") ... part of the new wiki page may need to be fixed with some line breaks.

MattKaufmann: Hi, David --

I don't see anything in the acl2+books manual about relocation stubs.
Could you add something about that, which could point to your wiki
page?  Probably I'm not the only ACL2 user who tends not to look for
wiki pages, instead expecting to find pretty much everything in the
manual other than what's easily accessible from the ACL2 home page.

I'm willing to help, but I know very little about such infrastructure,
which is why I'm asking you....

Thanks --
-- Matt

ragerdl: Hi,

Okay, I've added an issue and assigned it to me.  I'll probably get to it
someday, but if someone else wants to pickup the issue, I wouldn't object.

Thanks.
David

PR #577, opened 2016-01-17
Euclid guards

nadezhin:
Clear executable file mode from files in `projects/quadratic-reciprocity`
and add guards to `projects/quadratic-reciprocity/euclid.lisp`.


ragerdl:
@russinoff This seems to change your theory for euclid a little bit.  Though, I think it's in an acceptable way (e.g., look at `divides-mod-0`, which uses `acl2-numberp` and equal instead of `integerp` and `zp`).  Anyway, could I please get your approval or objections?  Thanks for the suggested patch Dima!

russinoff:
I think I'm OK with all of this.  I don't think it will break anything --
we'll see.

PR #584, opened 2016-02-06
Type of alists from symbols to symbols, nil-terminated.

acoglio:
I used 'programming' as additional xdoc parent for now, but perhaps 'std/alists' might be more appropriate?

### DIFF
--- a/books/kestrel/general/alists.lisp
+++ b/books/kestrel/general/alists.lisp
@@ -1,0 +1,32 @@
+; Alist Utilities
+;
+; Copyright (C) 2016 Kestrel Institute (http://www.kestrel.edu)
+;
+; License: A 3-clause BSD license. See the LICENSE file distributed with ACL2.
+;
+; Author: Alessandro Coglio (coglio@kestrel.edu)
+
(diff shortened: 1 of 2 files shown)


ragerdl:
I'm pretty sure we want things like this in std, but I'd want to think about whether a `symbol-symbol-alistp` would slow down ACL2 at all first.  Also, we'd want to move it into the `std` directory.  Or, even better, if there's a file that has types like this already in `std`, we could just add this type.  I know there's a file for `fty`'s in the centaur directory that has a lot of fixes for types built-in to ACL2 -- this is kind of along these lines.

acoglio:
I agree. I've put it under kestrel-books for now just so that it's out there, but I'm happy for it to go elsewhere. In fact, as things under the kestrel, centaur, and similar directories get more stable and used, perhaps they could be gradually moved to more "central", organization-independent positions. The directory hierarchy doesn't have to aligned with the xdoc hierarchy (which is easier to change), but it seems generally better for them to be as aligned as possible (modulo the fact that xdoc is a graph that may not be a tree).

PR #585, opened 2016-02-07
A new book about coprime integers.

nadezhin:
New book projects/quadratic-reciprocity/coprime.lisp
contains defintion of coprime integers and its corollaries:
     (1) Existence of modulo inverse
     (2) Chinese remainder theorem for two remainders.

### DIFF
--- a/books/projects/quadratic-reciprocity/README
+++ b/books/projects/quadratic-reciprocity/README
@@ -10,6 +10,11 @@ number theory.  It includes the following ACL2 books:
      (1) The infinitude of the set of primes
      (2) if p is a prime and p divides a * b, then p divides either a or b.
 
+  coprime
+     Defintion of coprime integers and its corollaries:
+     (1) Existence of modulo inverse
+     (2) Chinese remainder theorem for two remainders.
+
   fermat
      Fermat's Theorem: if p is a prime and p does not divide m, then
      mod(m^(p-1),p) = 1.
@@ -37,7 +42,7 @@ number theory.  It includes the following ACL2 books:
 
   eisenstein
       A formalization of Eisenstein's proof of the law of quadratic reciprocity:
-        If p and q are distinct odd primes, then 
+        If p and q are distinct odd primes, then
           (p is a quadratic residue mod q <=> q is a quadratic residue mod p)
             <=>
           ((p-1)/2) * ((q-1)/2) is even.
@@ -50,4 +55,3 @@ number theory.  It includes the following ACL2 books:
 
   pratt
       Vaughn Pratt's method of prime certification applied to the prime 2^255 - 19.
- 
--- a/books/projects/quadratic-reciprocity/coprime.lisp
+++ b/books/projects/quadratic-reciprocity/coprime.lisp
@@ -1,0 +1,203 @@
+(in-package "RTL")
+
+(include-book "support/coprime")
+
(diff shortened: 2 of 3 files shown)


nadezhin:
I just have found a directory `workshops/2000/russinoff-short` where @russinoff did this work 16 years ago. I didn't notice it because it is not in ACL2 documentation. I think that it would be nice to add to documentation a section arithmetic/number-theory and gather there links to various number-theory books. Now I'm looking at @russinoff 's book and request to postpone the review of this pull request.

jaredcdavis:
> I think that it would be nice to add to documentation a section arithmetic/number-theory and gather there links to various number-theory books.

FWIW, it's definitely unfortunate that so many books aren't included in the documentation.  If anyone with time has interest in documenting the details or mere existence of more books, especially important books like the various arithmetic libraries, that would (of course) be great.

nadezhin:
I discussed addition `arithmetic/number-theory` topic to XDOC and got an answer that `number-theory` is not part of `arithmetic`. This is true.
However, `arithmetic` topic in XDOC has already subtopics that are not parts of arithmetic. `arithmetic/cowles` is about algebra -groups and rings. 
I think about renaming XDOC `arithmetic` topic to `math` topic.
Then `math/number-theory` and `math/cowles` topics seem more natural.

ragerdl:
That change makes sense to me, though it may be a non-trivial amount of work to fix up all of the references to what is currently `acl2::arithmetic`.  Also, we'd leave most of `acl2::arithmetic` unchanged -- just put it under `math`.  Thoughts from others like @MattKaufmann , @jaredcdavis ?

russinoff:
Maybe math/cowles is not the best name for that directory.  Maybe
math/algebra?

MattKaufmann:
I like the idea of adding a new MATH topic directly under TOP, moving
the existing ARITHMETIC topic to be a child of MATH, and moving or
renaming some subtopics of ARITHMETIC as appropriate.

Perhaps someone would like to propose to this list an explicit
structure of topics under MATH, so that once nobody has objected for a
couple of days, it will be trivial to implement that structure (by
changing a few parents and write a very short MATH topic).

-- Matt

nadezhin:
I shall try to prepare such a list.
Can this list contain links to topics from workshops directories as well ?

MattKaufmann:
Thanks, Dmitry!  If you care to include links to topics under
books/workshops/, I think that would be great.

I was thinking of something simple.  The simplest version would just
be the following, which means that nothing changes except that a new
top-level topic, MATH, would have the existing topic, ARITHMETIC, as
its only child.

MATH
  ARITHMETIC

But a more sophisticated version could be something like the following
(but I haven't really thought about it -- I'm too lazy to do that, so
this is just to give a rough idea).

MATH
  ALGEBRA [formerly COWLES]
  ARITHMETIC
  BIT-VECTORS
    ARITH-EQUIVS
    BITOPS
    IHS
    RTL

-- Matt

ragerdl:
Note that Matt's reply doesn't show up formatted properly in the email.  Glad it's right here!

nadezhin:
The XDOC changes are submitted in a new pull request #608 . 
This pull request will be cancelled and probably reimplemented.

russinoff:
Thanks, I'll take a look.

nadezhin:
I cancel this pull request.

PR #587, opened 2016-02-15
Memoize: Add a heuristic to decide whether to use CLRHASH or simply d…

solswords:
…elete hash tables when clearing memoize/pons tables.

Matt -- This obviously touches ACL2 source code, so please have a look when you have time.

Previously, whenever we wanted to clear a function's memoization table, we'd set it to NIL, and next time we added an entry we'd create a new hash table.  This works fine, but is fairly memory inefficient in some circumstances where we're clearing the table a lot without adding very many entries.  In fact, we now think it's usually better to use CLRHASH -- the exception being when the table has grown very large and is not expected to grow as large again once we clear it.  This patch sets up a heuristic that chooses between using CLRHASH and setting to NIL/reallocating at each clear.  Calling clear-memoize-tables (as opposed to clear-memoize-table on a particular function) still always sets tables to NIL in order to free as much memory as possible.

This gives us a major speedup in a particular computational bottleneck where we happened to be repeatedly clearing a memoization table.  Admittedly, we have less empirical evidence of the impact in general, but we think this can only improve things.

### DIFF
--- a/memoize-raw.lisp
+++ b/memoize-raw.lisp
@@ -118,16 +118,6 @@
 ; - Consider eliminating some uses of our-syntax, instead relying on normal
 ;   ACL2 printing.
 
-; - Look at the following comment in memoize-fn-outer-body and consider the
-;   possibility of using clrhash in some cases (e.g., there and in
-;   clear-one-memo-and-pons-hash) to save spec.
-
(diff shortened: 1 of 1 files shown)


MattKaufmann:
Incorporated into commit c098a636113db5ccbe3bc4d0faea9c66a6792e65.

PR #589, opened 2016-02-23
ACL2s/Cgen/Defdata changes

harshrc:
* Summary
- Cgen/Defdata now supports type refinement/specialization (additional
  constraints on the variable under test) via constraint-refine-rules
  that can be specified using the defdata-attach macro. This feature
  is used to provide automatic elim support for Cgen for variables
  with record and map types.
- Re-sequenced location of hooks (pre, post, post-pred etc) in a way
  that makes more sense.
- Defdata interface macros are now logic-mode wrapped, to allow
  uniform usage in program-mode.
- Always check if counterexamples/witnesses are sound.
- Fixed bugs and made other performance improvements in
  incremental-search mode of Cgen.
- Refactored incremental-search code into multiple files.
- Type refinement (above) is implemented in refine-enum-shape function
  and the assign-value function of incremental-search.
- Added a new file that collects all ACL2s mode dependencies.


ragerdl:
Congrats on the ACL2(s) progress!

PR #590, opened 2016-02-24
Heuristic tweak to choice of whether to clrhash or reallocate memo ta…

solswords:
Matt -- another tweak to memoize heuristics.  When we are predicting how big a memo/pons table is likely to grow, we only care about how big it'll get if it gets written to at all.  This change makes it so we won't average zeros into that prediction when we clear a memoize table that doesn't have any entries.


MattKaufmann:
Completed with commit 12852f59b269fe3f268ec8c6409a660581a9293b.

PR #593, opened 2016-03-02
Minor improvements to the system utilities in the Kestrel Books.

acoglio:


ragerdl:
Hi @acoglio ,

I think I see what happened here.  It looks like Matt made some changes in the master branch, you submitted your pull request, and then jenkins copied the changes from master to testing an hour later.  As a result, "changes" are showing up in your pull request that aren't actually intended to be associated with the request.

If I merge the pull request, everything will actually work out just fine, and I don't think Matt's checking script will be triggered.  However, his two commits are showing up in this pull request (right now, they may not show up later), so I'm going to close this one and re-open one that just contains your two commits.

You did everything we asked.  If you want to avoid this in the future, then it'd be best if master and testing were in sync when you merge, but that's so trivial, that it's probably not worth worrying about.  Thanks for the contributions!

David

ragerdl:
Merged in #594

PR #595, opened 2016-03-08
Fix a few typos in the comments.

acoglio:


ragerdl:
Always good to see cleanup, thanks!

PR #604, opened 2016-04-15
Adding documentation for UOR and UAND.

ragerdl:
This may not be the right place for it, or even the right topics.

@jaredcdavis or @solswords , please review.


solswords:
Svex-eval interprets UOR using 4vec-reduction-or (see *svex-op-table*).  So maybe this should just be a pointer to that topic, or maybe we should fix up the naming convention so that the symbols used for functions in svex correspond to the function names that svex-apply uses to interpret them.   I think the current phrasing -- "Not currently defined as a function but used by the SVex interpreter as @('UOR')" -- is more confusing than helpful.

PR #608, opened 2016-04-29
Math XDOC topic.

nadezhin:
There were a discussion in #585 a few months ago about adding a new MATH topic directly under TOP, moving the existing ARITHMETIC topic to be a child of MATH. 

Sorry for a long delay. This pull request tries to implement that discussion

Organize some of math XDOC stuff in such a hierarchy:
```
Math
  Algebra
    Abelian-groups
    Abelian-semigroups
    Commutative-rings
  Arithmetic
    Arith-equivs
    Arithmetic-1
    Lispfloat
    Proof-by-arith
  Bit-vectors
    Bitops
    Ihs
    Rtl
```

### DIFF
--- a/books/centaur/bitops/top.lisp
+++ b/books/centaur/bitops/top.lisp
@@ -51,7 +51,7 @@
 (include-book "signed-byte-p")
 
 (defxdoc bitops
-  :parents (acl2::arithmetic)
+  :parents (acl2::bit-vectors)
   :short "Bitops is <a href='http://www.centtech.com/'>Centaur Technology</a>'s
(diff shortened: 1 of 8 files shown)


ragerdl:
To anyone following this discussion, #608 is the one you want to look at.  If there's no feedback within the next 24 hours, I'll probably just merge the pull request since it looks reasonable to me.

MattKaufmann:
Thanks, David; I think that simply merging after 24 hours is a great
idea.  And many thanks to Dmitry for carrying out this
reorganization!

-- Matt

PR #609, opened 2016-05-02
Xdoc number theory

nadezhin:
Avoid name conflict between `s` functions in `euclid.lisp` and `record.lisp`.

Create XDOC topic `number-theory`. It's parent is `arithmetic`.
Write markup for some books of @russinoff and attach them to the new topic.


ragerdl:
Cool!

I'll take your word about the redundancy of the type-prescription rules -- that's the only thing risky in this that I see, so I'm merging.  Thanks!

PR #610, opened 2016-05-04
Bitp updates

solswords:
These commits update some bit-level reasoning libraries (primarily ihs and centaur/bitops) to take advantage of ACL2's new ability to regard `bitp` as a typeset.  This is fairly disruptive to some libraries.  I have fixed the breakage this causes within the community books but I decided to make a pull request rather than just pushing because it might also cause breakage in non-public books.
Especially, @shigoel -- this affects proofs in the x86 model in a few places.  That said, I think this is an improvement in reasoning power and should eventually be incorporated.

### DIFF
--- a/books/centaur/bitops/congruences.lisp
+++ b/books/centaur/bitops/congruences.lisp
@@ -354,6 +354,13 @@
                            (b-and (logcar a) (logcar c))
                            (b-and (logcar b) (logcar c))))))))
 
+   (local (defthm logxor-1
+            (implies (bitp x)
+                     (equal (logxor 1 x)
+                            (b-not x)))))
+
+   (local (set-default-hints nil))
+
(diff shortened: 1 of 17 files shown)


MattKaufmann:
I'd appreciate waiting till after 5pm Central time tomorrow (May 5) at the earliest, before anyone closes this pull request.  (Thanks, Sol, for holding off on the push.)

shigoel:
@solswords Thanks a lot for fixing up the x86 books. I'm fine with your changes (and thanks for the annotations too!) as long as the x86 model still certifies.

solswords:
Sounds good.  If there are no further objections, then @MattKaufmann, you
can accept the pull req whenever you're ready for it (or let me know
otherwise).  Thanks!

PR #619, opened 2016-07-18
Adding farray to projects

nwetzler:
Adding farray to projects, more books to follow


ragerdl:
Thanks for the contributions!

PR #623, opened 2016-07-23
Change verify-guards-program + small doc fixes

ericsmithkestrel:
Matt, Let me know if the change to verify-guards-program isn't okay (it addresses a TODO from that file by returning an error if things fail, which lets me test it with must-succeed).  Feel free to fix the style if it's not up to snuff, of course.


MattKaufmann:
This looks good.  I plan to accept it this morning, after I get an unrelated change pushed (I'm hoping nobody accepts this before I do that push).

PR #633, opened 2016-08-14
A proposal to add a :PRED option to DEFAGGREGATE.

acoglio:
Here is a proposal to add a :PRED option to DEFAGGREGATE. While I suspect agreement on the usefulness of this addition, I'm not sure if the proposed code changes are the preferred way to implement the addition. Please feel free to reject the pull request if that's the case.

### DIFF
--- a/books/centaur/getopt/top.lisp
+++ b/books/centaur/getopt/top.lisp
@@ -688,11 +688,12 @@ and so forth.  This only applies to visible options."
           (make-parse-long-cases basename (cdr formals) world))))
 
 (define make-parse-long ((basename symbolp)
+                         (pred     symbolp)
                          (formals  formallist-p)
                          (world    plist-worldp))
   (b* ((parse-foo  (parser-name basename))
        (parse-long (parser-name-long basename))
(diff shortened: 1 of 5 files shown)


MattKaufmann:
I'm inclined to accept this request, even though I know essentially nothing about defaggregate, based on my confidence in Alessandro.  It's on my calendar to do so on Friday unless someone objects first.

ragerdl:
I don't really have a comment on the code.  I looked at it, and it looks reasonable to me.  But, I'd prefer to defer to @jaredcdavis 's judgement.

WRT the concept in general, I guess everyone has their own style.  Personally, it bothers me that we have cltl-compatible suffixes at all.  If were making my own package, I'd probably make macros named `rager::string-p` and `rager::integer-p`.  But, I recognize that that's a style choice, and this patch strives to accommodate the diversity of style preferences that exist.

jaredcdavis:
No objections here.  I should have used ? years ago to avoid the whole p versus -p thing.  With this patch we can fix or exacerbate that mistake, one aggregate at a time. :)

PR #637, opened 2016-09-06
Make fty::defprod with (unsigned-byte-p NNN x) fields faster.

nadezhin:
fty::defprod allows type of field to be a macro.
However, defprod is slow in this case because
it can't find user-supplied fix-pred and pred-of-fix theorems.
Often this macro is expanded to (unsinged-byte-p NNN x)
where NNN is a natural constant.
THis commit modifies flextypes-collect-fix/pred-pairs-aux so that it can handle
this particular case.
General solution for any macro is difficult for me now.

### DIFF
--- a/books/centaur/fty/deftypes.lisp
+++ b/books/centaur/fty/deftypes.lisp
@@ -542,6 +542,47 @@
           (mv t rune)
         (find-fix-when-pred-thm-aux fix pred (cdr fix-rules))))))
 
+(defun find-fix-when-unsigned-byte-p-pred-thm-aux (fix pred size fix-rules)
+  (if (atom fix-rules)
+      (let ((body `(implies (acl2::unsigned-byte-p (acl2::quote ,size) x)
+                            (equal (,fix x) x))))
+        (mv nil `(local (make-event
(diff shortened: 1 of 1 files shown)


ragerdl:
Hi @nadezhin ,

Does this patch address the problem described in:

https://groups.google.com/forum/#!searchin/acl2-books/defprod%7Csort:relevance/acl2-books/PT_dLee7pK0/A3Q32pvlBgAJ

If so, this is a well-motivated improvement.  Otherwise, I still think this change is probably worth doing, as all of us use unsigned-byte-p to describe our defprod fields.  Though, I don't know how many in the community define macros for unsigned-byte-p (we should probably release our book...).  I think @shigoel might too?

Anyway, even though it's not general, I'm in favor of this patch.  @solswords , do you have any recommendations/objections?

Thanks,
David

solswords:
I'm going to accept this patch for now but I think there's a more complete solution that wouldn't be too hard to implement -- we have the world available so we should just translate e.g. `(pred (fix x))` and look for rewrite rules that unify with the translation.

PR #641, opened 2016-09-17
Icarus verilog

nadezhin:
Add Verilog cosims test on wildcards in case statements

### DIFF
--- a/books/centaur/sv/cosims/case_wilds/spec.sv
+++ b/books/centaur/sv/cosims/case_wilds/spec.sv
@@ -1,0 +1,89 @@
+module spec (input wire  [127:0] in,
+	     output wire [127:0] out);
+
+
+   wire a = in[0];
+   reg[127:0] tmp;
+
+   always @* begin
+      case ({a})
+	1'b0: tmp[0] = 1;
+	default: tmp[0] = 0;
+      endcase // case ({a})
+
+      case ({a})
+	1'b1: tmp[1] = 1;
+	default: tmp[1] = 0;
+      endcase // case ({a})
+
+      case ({a})
+	1'bx: tmp[2] = 1;
+	default: tmp[2] = 0;
+      endcase // case ({a})
+
+      case ({a})
+	1'bz: tmp[3] = 1;
+	default: tmp[3] = 0;
+      endcase // case ({a})
+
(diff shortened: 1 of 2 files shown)


nadezhin:
ncverilog and iverilog give the same result.
They both differ from result of ACL2 verilog.
I didn't test vcs because something wrong in my setup.

ragerdl:
Hi @nadezhin ,

It looks to me like the cosims for `case_wilds` don't pass.  I'll submit a patch to disable these tests.  Do you know why they pass for you but not us?  Maybe we can move it from the failing to the passing tests if we figure out what happened.

Thanks~

MattKaufmann:
Hi, David --

It seems a bit odd perhaps that you and Jared get a failure that I
didn't get with my "everything" regression.  Presumably it's because
the "cosims" tests that you're talking about aren't part of that
regression.  If you guys figure out a way to include them, at least
for CCL and perhaps SBCL, that might lead to earlier reports of
problems, since I do a lot of "everything" regressions with those
Lisps.

-- Matt

jaredcdavis:
Hi Matt,

They aren't part of the everything regression because they require
commercial Verilog simulators that many users won't have.

Cheers,
Jared

MattKaufmann:
Good to know; thanks, Jared.  (I had the impression that Dmitry added
support for using a non-commercial simulator, but even if so, maybe it
would still be painful to include such testing in regressions.)

-- Matt
> Date: Wed, 21 Sep 2016 16:47:07 -0700
> From: Jared Davis <notifications@github.com>
> Reply-To: acl2/acl2 <reply@reply.github.com>
> Cc: MattKaufmann <kaufmann@cs.utexas.edu>,
>         Comment <comment@noreply.github.com>
> 
> 
> [1:text/plain Hide]
> 
> Hi Matt,
> 
> They aren't part of the everything regression because they require
> commercial Verilog simulators that many users won't have.
> 
> Cheers,
> Jared
>

nadezhin:
They don't pass for me.
I said: "ncverilog and iverilog give the same result. They both differ from
result of ACL2 verilog."

ragerdl:
Ah, we should have caught that.  Looks like everything got resolved as appropriate; thanks!

PR #671, opened 2016-11-09
Fix cert.pl relocation warnings

kini:
I think this should be a pretty safe commit, but running a regression anyway...


ragerdl:
Thanks for the cleanup and submitting it to testing!

PR #676, opened 2016-12-02
Eschew gratuitous boiling

kini:
Inspired by the recent acl2-help thread :)


ragerdl:
Does this change the message that's printed?

kini:
Nope, it shouldn't change anything, just a small code cleanup (converting `(if a (if a b c) c)` to `(if a b c)`). I have another commit which changes all the Seussian stuff to something more boring, but II wasn't sure if there was a consensus on the list. I've added it to the PR in any case, but I can remove it if that's better.

PR #679, opened 2017-01-04
Display cert.pl build times to the millisecond

kini:
A lot of books certify in "0s" or "1s", so I thought it would be nice to
offer a little higher precision for ranking purposes.  Millisecond
precision is also the default output for the `time` command in recent
Bash versions.

The Time::HiRes module is shipped with the base Perl system since Perl
5.8, which was released in 2002, so I think it should be safe to use.

### DIFF
--- a/books/build/make_cert_help.pl
+++ b/books/build/make_cert_help.pl
@@ -71,6 +71,7 @@ use FindBin qw($RealBin);
 use POSIX qw(strftime);
 use Cwd;
 use utf8;
+use Time::HiRes qw(time);
 
 binmode(STDOUT,':utf8');
 
@@ -818,7 +819,7 @@ if ($success) {
 	$hostname = " " . extract_hostname_from_outfile($outfile);
     }
 
-    printf("%sBuilt %s (%ds%s)%s\n", $color, $printgoal, $ELAPSED, $hostname, $black);
+    printf("%sBuilt %s (%.3fs%s)%s\n", $color, $printgoal, $ELAPSED, $hostname, $black);
 
 } else {
     my $taskname = ($STEP eq "acl2x" || $STEP eq "acl2xskip") ? "ACL2X GENERATION" :


MattKaufmann:
This change won't affect books/build/compare.pl, will it?  While it could be amusing to see a bit more precision, nevertheless I've always considered short certification times subject to noise, especially when using -j; so I'd hate to see any negative impact from this change.

kini:
I don't know enough about the build infrastructure as a whole to really understand what implications this change might have, so this pull request is just a suggestion. I'm hoping for comments from those who are more knowledgeable. Thanks for writing the first one!

Looking at `books/build/compare.pl`, it seems like the information it uses is ultimately derived from `.time` files produced by cert.pl when the environment variable `TIME_CERT` is set to `1`. The change proposed in this pull request doesn't seem like it should change anything related to that, as far as I can tell. Furthermore, those `.time` files actually already use millisecond timing, it seems.

MattKaufmann:
Thanks for the reply.  I suppose the worst thing that could happen would be that someone needs to deal with the fallout if something goes wrong, perhaps simply by reverting this change.  The chance of something going wrong seems small (though so is the benefit perhaps); I'll leave it to someone else to decide whether to accept this pull request (fine with me).

ragerdl:
Thanks for the conversation thus far.  In summary, it seems fine.

@solswords you have intimate knowledge of the perl scripts.  Could you please merge it?

kini:
Here's what I used this PR for, if anyone's interested.

A "cactus plot" of build times for community books on different lisps, for ACL2 v7.3 (before this PR was merged):

performance-7 3

A similar plot for ACL2 v7.4 (after this PR was merged):

performance-7 4

jaredcdavis:
This is really, really cool.  How did you generate these?

kini:
Glad you found it interesting :) I made a writeup for the plots from 7.3 [here](http://www.cs.utexas.edu/~kini/books-benchmark/performance.html), and you can see the source code that produced the images by looking at [the org-mode source file](http://www.cs.utexas.edu/~kini/books-benchmark/performance.org). I made this a couple months ago and was going to clean it up a bit and post it to the list, but now I'll have to reanalyze the plots for 7.4 since 7.4 was released. It looks like this time around, SBCL is not as dominant as it was last time...

kini:
Oh, and the log files referenced in the code are from `/projects/acl2/releases2/v7-3/` at UTCS.

MattKaufmann:
Regarding:

> It looks like this time around, SBCL is not as dominant as it was last time...

Right, I sent a bug report to the sbcl-bugs list on March 14, about
seeing reduced performance when I transitioned from SBCL 1.3.11 to
1.3.15.  I'm glad you brought this up because it led me to check an
SBCL bugs page -- indeed there was a brief reply to my message (which
I didn't get via email) that needed some follow-up, which I just sent.

-- Matt

PR #680, opened 2017-01-06
Paco changes

nadezhin:
1) Use `*ts-one*` and `*ts-integer>1*` in `type-set-XXX` functions
2) Fix bug in `fn-count`
3) Fix wrong calls of `fn-nume`
4) Use `o<` instead of `e0-ord-<` . 
    Speedup measure proofs by disabling unnecessary functions.
    Remove speedup lemmas which becomes unnecessary


ragerdl:
@MattKaufmann these look like fairly straightforward cleanups to me.  I'll leave it to you to consult with J as you think appropriate.

PR #682, opened 2017-01-09
Paco: Port recent fix of disabling compound-recognizer rules

nadezhin:


ragerdl:
@MattKaufmann This should look pretty familiar to you.  Mind being the gate-keeper for it?

PR #683, opened 2017-01-16
Extend svtv-run theorems to other keyword args

kini:
At some point, more keyword arguments were added to `svtv-run`, but
the theorems associated with `svtv-run` don't account for these
arguments.  This commit attempts to fix that situation by generalizing
the `svtv-run` theorems to account for cases where the user used
`:include`, `:simplify`, and `:readable`.

----

Note: some of these changes are just simple generalizations of the LHS, but others materially change the RHS of the rule, so please check that I haven't done anything stupid.

### DIFF
--- a/books/centaur/sv/svtv/process.lisp
+++ b/books/centaur/sv/svtv/process.lisp
@@ -1570,28 +1570,26 @@ stvs-and-testing) of the @(see sv-tutorial) for more examples.</p>"
     :hints(("Goal" :in-theory (enable svtv->outs))))
 
   (defthm lookup-in-svtv-run-consp
-    (equal (consp
-            (assoc key (svtv-run
-                        svtv inalist
-                        :include include :skip skip :boolvars boolvars
-                        :simplify simplify :quiet quiet :readable readable)))
-           (and (member key (svtv->outs svtv))
-                (if include
-                    (member key include)
-                  (not (member key skip)))
-                t))
+    (iff (consp
+          (assoc key (svtv-run
+                      svtv inalist
+                      :include include :skip skip :boolvars boolvars
+                      :simplify simplify :quiet quiet :readable readable)))
+         (and (member key (svtv->outs svtv))
+              (if include
+                  (member key include)
+                (not (member key skip)))))
     :hints(("Goal" :in-theory (enable svtv->outs))))
 
   (defthm 4vec-p-lookup-in-svtv-run
-    (equal (4vec-p (cdr (assoc key (svtv-run
-                                    svtv inalist
-                                    :include include :skip skip :boolvars boolvars
-                                    :simplify simplify :quiet quiet :readable readable))))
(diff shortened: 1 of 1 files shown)


books/centaur/sv/svtv/process.lisp
```diff
@@ -1559,30 +1559,60 @@ stvs-and-testing) of the @(see sv-tutorial) for more examples.</p>"
...
-              (not (member key skip))))
+    (equal (consp
+            (assoc key (svtv-run
+                        svtv inalist
+                        :include include :skip skip :boolvars boolvars
+                        :simplify simplify :quiet quiet :readable readable)))
+           (and (member key (svtv->outs svtv))
+                (if include
+                    (member key include)
+                  (not (member key skip)))
```
kini: Maybe rather than an `if`, this should be two theorems with `syntaxp` to help ACL2 choose which one to use. I imagine that one is never going to be reasoning about a call of `svtv-run` which doesn't have at least one of `:include nil` or `:skip nil` in its arguments, and I mean literally `nil` rather than just something provably equal to `nil`.

books/centaur/sv/svtv/process.lisp
```diff
@@ -1559,30 +1559,60 @@ stvs-and-testing) of the @(see sv-tutorial) for more examples.</p>"
...
+                t))
```
kini: Surely there's a less ugly way of making the RHS a Boolean (to match the LHS) than to `and` it with `t`...

solswords: The only problem with (and (foo) t) is that it will casesplit on (foo).  But in this case I think we're anticipating that we'll mostly see calls of svtv-run with constants for skip and include so it should be OK.  Otherwise we could use (consp (member ...)) or just change the equiv relation to IFF.

solswords:
I'm fine with these as long as all the community books still certify -- I imagine you've already done that?

books/centaur/sv/svtv/process.lisp
kini: My original thought was to change the relation to `iff`, but I feared that might break a proof where this rewrite rule was used in an `equal` context. If you think that works fine, I'll do that instead of this `(and ... t)` thing.

kini:
Yes, the books do certify fine with this change. I used `make all` in the top `books/` directory.

kini:
The books still certify after changing a couple `equal`s to `iff`s as suggested. Incidentally, it seems very few books in the community books actually depend on this code -- indeed, nothing outside of the SV library itself. Perhaps that is to be expected, since SV is probably mostly used internally at Centaur, Oracle, etc. rather than in the community books.

ragerdl:
@kini One thing you could check when making these types of changes is whether all of the centaur tutorials still work.  Some of those files may end in `.lsp` and not be included as part of the typical books build.  Thanks for the contributions and @solswords for weighing in on them.

PR #684, opened 2017-01-18
Remove DA warning message when fields aren't used

### LINKED ISSUE #668: Note printed on screen when including books/centaur/fty/top.lisp
When including "centaur/fty/top" from :dir :system, the following gets printed on screen:

Note: not introducing any FLEXTRANSSUM-MEMBER field bindings for MEMBER,
since none of its fields appear to be used.

Matt noted that this may come from books/std/util/da-base.lisp.

Is it necessary for this message to be printed? It seems a bit odd to me, given that I'm just including a library, but perhaps there's a good reason for it or it would be inconvenient to remove.

kini:
Fixes #668.

----

As sort of suggested by @solswords in #668. If you don't like this solution, another possibility is to actually fix various places where defaggregate fields are unused in a b* binding.

### DIFF
--- a/books/std/util/da-base.lisp
+++ b/books/std/util/da-base.lisp
@@ -899,10 +899,6 @@ fty::deftranssum).</p>"
        (full-vars-alist (da-patbind-make-field-acc-alist var fields-accs))
        (field-vars      (strip-cars full-vars-alist))
        (used-vars       (da-patbind-find-used-vars rest-expr field-vars nil))
-       (- (or used-vars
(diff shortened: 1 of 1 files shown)


ragerdl:
Should I be merging this?  @solswords , I think this is your call

kini:
Actually I think I'll close it for now -- actually fixing the errors as they crop up, like I did in #688, might be a better solution.  I wish it was possible to generate those errors statically, though...

PR #687, opened 2017-01-29
Make `cert.pl -v` not crash

kini:
Struct `Certinfo` has no member `certparams`, but it does have a
member `params`.  I presume that's what was intended here.

### DIFF
--- a/books/build/certlib.pl
+++ b/books/build/certlib.pl
@@ -1734,28 +1734,28 @@ sub add_deps {
     $depdb->certdeps->{$target} = $certinfo ;
 
     if ($print_deps) {
-	print "Dependencies for $target:\n";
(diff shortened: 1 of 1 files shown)


kini:
Actually would anyone object to me rewriting this to dump a JSON dict? Since `cert.pl -v` always crashed before this commit, I presume nobody is using it as part of any toolchain...

solswords:
Would you be OK with adding the JSON capability as a separate feature?  This was just supposed to provide human-readable debugging, but I'd be very happy to have cert.pl be capable of writing json as well if you'd like to implement it.

kini:
Sure, that makes sense.  I'll make a separate PR for that when I do it, then.

PR #697, opened 2017-02-17
Make `find-lemmas` much faster

kini:
`find-lemmas-fn` was repeatedly calling `deref-macro-name` on the
function names supplied by the user, every time it inspected an event
recorded in the world.  When the world is very long, this can be very
time consuming because `deref-macro-name` uses tables, which as
pointed out in `:doc tables` are "not especially efficient".

This commit makes it so that `deref-macro-name-list` is only called
once per invocation of `find-lemmas`.  In one example, running
`find-lemmas` in a world containing more than 1.5 million triples used
to take more than 5 minutes, but now it completes in 0.15 seconds.

### DIFF
--- a/books/misc/find-lemmas.lisp
+++ b/books/misc/find-lemmas.lisp
@@ -14,32 +14,31 @@
 
 (defun find-lemmas-fn (fns omit-boot-strap acc wrld-tail wrld)
   (declare (xargs :mode :program))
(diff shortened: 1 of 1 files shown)


MattKaufmann:
Looks great to me!  I'm embarrassed that I wrote this so inefficiently the first time.

kini:
Thanks, Matt!

Issue #693, opened 2017-02-06
ACL2(r) book nonstd/nsa/continuity.lisp won't certify


MattKaufmann:
The recent fix for a long-standing functional instantiation soundness
bug has caused the book nonstd/nsa/continuity.lisp, because one of its
theorems, intermediate-value-theorem, exploited the bug in its proof.
My attempts so far to fix the proof were not successful, but I'm
hoping it will be easy for Ruben Gamboa to fix, as I think it's his
book.  I'd assign this Issue to Ruben, but he doesn't seem to be a
contributor to the project; I've contacted him by email.  Anyone is
welcome to work on it as far as I'm concerned (but you might want to
contact Ruben and the acl2-books list first, to make sure you're not
duplicating effort).  I'm about to push a change that excludes this
book from ACL2(r) builds (it was already excluded from non-ACL2(r)
builds because it truly depends on ACL2(r)).

rubengamboa:
This has been resolved.

PR #698, opened 2017-03-02
Remove obsolete comments (followup to #693)

kini:
The comments Matt added in 04a6f6d9084a647dccee1e508e108895bb8c0e6a are no longer meaningful, so this commit removes them.  An alternative would be to edit the comments so it's clear they are historical.  In any case, the `cert_param` declaration seems unnecessary.

(Please check -- I'm not too familiar with how ACL(r) books are certified.)

### DIFF
--- a/books/nonstd/nsa/continuity.lisp
+++ b/books/nonstd/nsa/continuity.lisp
@@ -1,12 +1,5 @@
 (in-package "ACL2")
 
-; There is a problem with intermediate-value-theorem, which no longer proves
-; after the soundness bug fix for ACL2 7.3.  So we exclude this book from all
-; builds: it already requires ACL2(r), since it depends on nsa.lisp, which
-; specifies the cert_param of uses-acl2r, but now we also specify:
-
-; cert_param: (uses-acl2r)
-
 ;; This book establishes some facts about real continuous functions.
 ;; First, it shows that a function that is continuous on a closed
 ;; interval is uniformly continuous.  Second, it proves the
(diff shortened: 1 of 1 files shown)


MattKaufmann:
Thanks for removing that comment.  As for the cert_param -- did you
run an ACL2(r) regression?  I already checked with Ruben about that,
and he thought it might be needed (though I realize it wasn't needed
previously).  If you say that you ran such a regression, I'll accept
the pull request; if you're not sure, we might save time by being lazy
and leaving it in (it seems harmless).

Come to think of it: Can anyone tell me if the leeroy mechanism tests
ACL2(r)?  If not, might it be easy to make that happen whenever a
change is made under books/nonstd/?

-- Matt

kini:
I ran `make ACL2=$PWD/saved_acl2r regression` in the base ACL2 directory, and it completed without issue, if that helps.

ragerdl:
I used to test the books with all builds but decided to trade such builds
for better responsiveness (w.r.t. merging in commits from testing to
master).

Currently there's a build named "acl2-multi-arith2".  Its intent is mostly
to see whether ACL2 builds on the various combinations of features, and
then test each of those combinations by building the arith2 book.  I could
change that build to build the nonstd books too (since I assume cert.pl
will exclude the books that don't apply to a particular feature set).  Are
the nonstd books quick and have only a few dependencies?




> Come to think of it: Can anyone tell me if the leeroy mechanism tests
> ACL2(r)? If not, might it be easy to make that happen whenever a
> change is made under books/nonstd/?
>

MattKaufmann:
Hi, David --

Thanks for dealing with such matters.

> .... Are
> the nonstd books quick and have only a few dependencies?

Not all that quick -- here is the time (an hour and 38 minutes) from
such a run that I did based on CCL for Version 7.4.

43479.842u 627.953s 1:38:21.13 747.4%	0+0k 984736+3136272io 634pf+0w

-- Matt

ragerdl:
Ah, that's probably too long to warrant the change.

The build that's supposed to capture this problem is
"acl2-multi-everything", but it's too long of a build to enable on our
current infrastructure.  I'll probably have the time to look into pricing
for more cores during the workshop, so come ask me about it then :)

MattKaufmann:
Hi, David --

No problem as far as I'm concerned.  I'll try to fold the occasional
ACL2(r) run into my process, and at any rate, I always do such a run
as part of the release process.

Thanks --
-- Matt

PR #701, opened 2017-03-23
New completeness theorem in shnf.

russinoff:

PR #702, opened 2017-03-23
New completeness theorem in shnf."

ragerdl:
@russinoff proved the completeness theorem for sparse Horner normal forms in books/projects/shnf:
If two polynomials have the same value for every set of variable assignments, then they produce the same normal form.

This is a pull request that should be from russinoff's master into acl2's testing.


kini:
For posterity, this is a reroll of #701.

PR #704, opened 2017-03-27
Org-ify the README for the 2017 ACL2 Workshop

kini:
This commit makes the README show up in a nicely formatted way on
github, while remaining easy to read as text in an editor as well --
especially in Emacs, which has special handling for .org files.


kini:
(preview of how the README will look on github: https://github.com/kini/acl2/tree/workshop-readme/books/workshops/2017 )

MattKaufmann:
Since I'm the one maintaining this README.org and I don't want to take the time now to learn org mode markup, could you summarize what is necessary to know in order to maintain this README.org for org mode?  Perhaps all you need to say is that ~...~ is for typewriter font and [[...]] is for links, though those are probably incorrect guesses.  Alternatively, I could send you things to add to the file and you could do it.

kini:
Those guesses are right on the money. Also, for linking to locations inside the repository, if the path starts with `/` then it's relative to the root of the repository, and if not, then it's relative to the current directory.  (There are examples of both in the file I've added in this pull request.)

I'd be happy to update this file for you whenever you like -- just send me an email. That said, you can also feel free to reject this pull request if you think it causes more problems for you than it solves.  It's just a suggestion :)

MattKaufmann:
Thanks for the explanation.  I can live with this, but when I tried saving a copy in Emacs 24.5.1 on my Mac I noticed some issues.  For one, there was lots of underlining that I found useless without knowing how to follow the links -- I guessed that C-c C-j would work, but it put me in a state where I had to do meta-x abort-recursive-edit to get out of it.  For another, I noticed that some titles were italicized sometimes when I entered org-mode but not others.

I'm not sure these oddities make it worth setting a precedent of using org-mode files.  But I'm not sure they don't.  I'm just nervous about time syncs (I've spent several minutes already on this, though of course that's my fault).

So I'll stop now and give others a chance to discuss.  I'll leave it to someone else to accept the pull request if they like, since I'm neutral.  Thanks for looking for ways to move us forward -- I don't want to get in the way of anything useful, but I wanted to point out the glitches, at least.  (And of course I don't want to debug them.)

ragerdl:
Hi,

I don't know about README.org, but org mode in emacs is awesome.  That
being said, I don't have an opinion on whether we should be using org mode
for our readmes.

As far as underlines go, they're typically used to provide emphasis (as
opposed to just links).

Best,
David

kini:
Ah -- about the links, to make a long story short, I used a github-specific path format that's more concise and easy to read but that doesn't work in Emacs itself.  My main motivation for this change was to make the README look nicer on github, so I didn't consider that you (or others) would want to follow the links directly in Emacs -- but I guess it makes sense that people would want to do that.

I can change to the "real" link format, but then it's no longer as simple as just wrapping things in [[]], so you'd probably want me to edit the file for you if you don't want to learn org-mode syntax. By the way, in general, the command for following a link in org-mode is `C-c C-o` (for "open").

As for the italics, I've sometimes seen issues with fontification "not kicking in" in org-mode.  I've never bothered to really debug it, but usually hitting `M-q` (i.e. `M-x fill-paragraph`) on the text in question will make Emacs recompute the font info.  It seems to occur more often when there's a line break in the middle of something.

kini:
Thanks for the input, David. I agree that org-mode is very nice :)  I think Matt was referring to the underlines that Emacs puts on links when viewing a buffer in org-mode, as opposed to actual underline markup, though.

By the way, you have both alluded to this PR being a precedent for using org-mode elsewhere in the repository, or "moving us forward" in some way -- so I just want to say that it wasn't really my intention to start that kind of a discussion, nor am I suggesting we start using org-mode all over the place, for whatever that's worth. I just wanted to prettify this one README for the moment :)

bigos:
Sublime text has partial support for org-mode
https://github.com/danielmagnussons/orgmode
And could be used as a stepping stone for someone who is not familiar with org-mode.

ragerdl:
Ah, if the web browser (i.e., github) knows how to display org mode, then I'm in favor of this pull request.

bigos:
It's not the browser. It's a library Github uses. But you can also easily export to HTML and PDF. IMHO it is easier to plan and write documentation in org-mode.

kini:
@bigos: Matt has been using Emacs for decades, so I don't think he'll find the Sublime Text suggestion that useful, I'm afraid :)  Actually, I'd normally have just put this in Markdown, but I figured that since most ACL2 users (including me and Matt) seem to use Emacs, I might as well go with org-mode.  Markdown is not as nice as org-mode, but on the other hand github's understanding of it is better than github's understanding of org-mode.

Anyway, I'm having second thoughts about this PR.  If I were directly maintaining this file I think it would be great to have it in org-mode, but at the end of the day, I'm not the one in charge of maintaining the `books/workshops/` directory, and I don't want Matt to have to spend time learning org-mode for such a trivial reason.  I thought I could keep it pretty simple at first, but the latest commit introduces markup that is probably not immediately obvious.

I'm going to close this PR for now.  Thanks for your feedback, everyone!

PR #705, opened 2017-03-27
Remove notice about ACL2 v7.4

kini:
This text needed to be included in the version of the README which was
included in the v7.4 release of ACL2, but now that that release has
passed, the text is no longer needed.

Adapted from b6b7b03d84da45a87a5cc058d3d329cf6e63f52a, one of
the commits in #704.

### DIFF
--- a/books/workshops/2017/README
+++ b/books/workshops/2017/README
@@ -1,12 +1,3 @@
-Note: This directory will likely be updated with more contributions
-after the ACL2 Version 7.4 release.  Updates are on github:
-
-https://github.com/acl2/acl2/tree/master/books/workshops/2017
-
-============================================================
-
-To authors:
-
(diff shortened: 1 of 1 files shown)


MattKaufmann:
Thanks for looking out for this sort of thing, but actually the comment is still valid, in that some additions are still missing.  If you prefer a different wording (e.g. to remove mention of 7.4) that's fine; or we cans chat privately about this if you like.  I'll hit "Close and comment" to avoid accepting -- if that's wrong, we can fix it later.

kini:
Ah, OK.  I guess I misunderstood the purpose of the comment, then.

MattKaufmann:
Aha, right, I can see how it's confusing.  I've just removed any
reference to github or Version 7.4.  Thanks for catching this.

-- Matt

PR #707, opened 2017-03-31
Make cert.pl try harder to find the system books

kini:
Currently, cert.pl looks for the system books in the following
locations, in order, until one succeeds (defaulting to the second
one):

1) the `books` subdirectory of the directory containing the ACL2
   executable cert.pl is using

2) the parent directory of the directory where cert.pl itself lies

This commit adds another option in between the two:

1.5) the value of `(@ system-books-dir)` reported by the ACL2
     executable when it is run

This is useful if, for example, your ACL2 executable is actually a
script that sets environment variables or does other kinds of setup
before actually calling ACL2 itself, and that script resides in a
totally different location from the actual ACL2 image that it
eventually launches.

This commit also cleans up a couple of comments, and adds an output
warning message in the case that the system books directory could not
be determined, where previously cert.pl would silently continue.

----

Note: I've only tested this on Linux.  I believe it should work on OS X as well, since I don't think I used any non-POSIX features of sh or awk.  I'm not sure about Cygwin.

### DIFF
--- a/books/build/cert.pl
+++ b/books/build/cert.pl
@@ -48,6 +48,7 @@
 
 use strict;
 use warnings;
+
 use FindBin qw($RealBin);
 use Getopt::Long qw(:config bundling_override);
 
@@ -94,6 +95,8 @@ my $make = $ENV{"MAKE"} || "make";
 my @make_args = ();
 my $acl2 = $ENV{"ACL2"};
 my $acl2_books = $ENV{"ACL2_SYSTEM_BOOKS"};
+my $startjob = $ENV{"STARTJOB"};
+if (! $startjob ) { $startjob = "bash"; }
 my $keep_going = 0;
 my $var_prefix = "CERT_PL";
 my %certlib_opts = ( "debugging" => 0,
@@ -487,6 +490,15 @@ USEFUL ENVIRONMENT VARIABLES
          certification instructions) and .cert.out files.  Note that this
          does not affect .cert.time files.
 
+    STARTJOB (default: "bash")
+         Can be set to the name of a command to use instead of bash
+         when launching a subprocess that will run ACL2.  The command
+         will be called as `$STARTJOB -c "bash code"`.  This is mainly
+         useful if you wish ACL2 to always be run in some environment
+         other than the current shell -- for example, if you want to
+         run ACL2 on a managed compute cluster, you might set
+         $STARTJOB to the name of a script that queues a job on the cluster.
+
 ';
 
 GetOptions ("help|h"               => sub { print $summary_str;
@@ -635,12 +647,15 @@ if (! $acl2_books && $acl2 ) {
 }
 
 if (! $acl2_books && $acl2 ) {
-    my $dumper =
-        "echo '(cw \"~%CERT_PL_VAL:~S0~%\" (@ system-books-dir))' | " .
+    my $dumper1 = # command to send to ACL2
+        '(cw \"~%CERT_PL_VAL:~S0~%\" (@ acl2::system-books-dir))';
+    my $dumper2 = # command to send to STARTJOB
+        'echo "' . $dumper1 . '" | ' .
         "$acl2 2>$devnull | " .
-        "awk -F: '/CERT_PL_VAL/ { print \$2 }'" .
-        "\n";
-    my $tmp_acl2_books = `$dumper`;
+        'awk -F: "/CERT_PL_VAL/ { print \$2 }"';
+    my $dumper3 = # command to send to the shell
+        "$startjob -c '$dumper2'";
+    my $tmp_acl2_books = `$dumper3`;
     chomp($tmp_acl2_books);
     if (-d $tmp_acl2_books) {
         $acl2_books = $tmp_acl2_books;
--- a/books/build/cert.pl
+++ b/books/build/cert.pl
@@ -589,9 +589,8 @@ certlib_set_opts(\%certlib_opts);
 
 my $cache = retrieve_cache($cache_file);
 
-# If $acl2_books is still not set, then:
-# - set it based on the location of acl2 in the path, if available
-# - otherwise set it to the directory containing this script.
+# If $acl2 is still not set, then set it based on the location of acl2
+# in the path, if available
 
 unless ($acl2) {
     $acl2 = "acl2";
@@ -620,18 +619,46 @@ flag or ACL2 environment variable.\n";
     }
 }
 
-if (! $acl2_books) {
-    if ($acl2) {
-	# was:
-	# my $tmp_acl2_books = rel_path(dirname($acl2), "books");
-	my $tmp_acl2_books = File::Spec->catfile(dirname($acl2), "books");
-	if (-d $tmp_acl2_books) {
-	    $acl2_books = $tmp_acl2_books;
-	} else {
-	    $acl2_books = "$RealBin/..";
-	}
-    } else {
-	$acl2_books = "$RealBin/..";
+# If $acl2_books is still not set, then try the following in order
+# until one succeeds:
+#   - set it based on the location of acl2 in the path
+#   - set it by running acl2 and dumping (@ system-books-dir)
+#   - set it to the parent of the directory containing this script.
+
+if (! $acl2_books && $acl2 ) {
+    # was:
+    # my $tmp_acl2_books = rel_path(dirname($acl2), "books");
+    my $tmp_acl2_books = File::Spec->catfile(dirname($acl2), "books");
+    if (-d $tmp_acl2_books) {
+        $acl2_books = $tmp_acl2_books;
+    }
+}
+
+if (! $acl2_books && $acl2 ) {
+    my $dumper =
+        "echo '(cw \"~%CERT_PL_VAL:~S0~%\" (@ system-books-dir))' | " .
+        "$acl2 2>$devnull | " .
+        "awk -F: '/CERT_PL_VAL/ { print \$2 }'" .
+        "\n";
+    my $tmp_acl2_books = `$dumper`;
+    chomp($tmp_acl2_books);
+    if (-d $tmp_acl2_books) {
+        $acl2_books = $tmp_acl2_books;
+    }
+}
+
+if (! $acl2_books ) {
+    my $tmp_acl2_books = "$RealBin/..";
+    if (-d $tmp_acl2_books) {
+        $acl2_books = $tmp_acl2_books;
+    }
+}
+
+if (! $acl2_books ) {
+    unless ($quiet || $no_build) {
+        print
+"ACL2 system books not found.  Please specify with --acl2-books
+command line flag or ACL2_SYSTEM_BOOKS environment variable.";
     }
 }
 


kini:
Sorry, there was a dumb typo in the patch.  I updated the patch to fix the typo.

solswords:
This sounds basically reasonable.  The books/GNUMakefile does something like this to determine the features supported (hons, parallel, etc).  The only problem is that some corporate users run the build system from machines where they can't actually run ACL2 (jobs are launched on a cluster instead).  The fix is to call ACL2 through startjob, like the makefile does.  (Startjob is just some program that reads a shell command from its '-c' argument and runs it, defaulting to bash.) You can find the relevant code in GNUMakefile by searching for STARTJOB.  Maybe you can replicate this in perl?

kini:
Thanks for the suggestion.  Looking at `books/GNUMakefile`, I see that `$(STARTJOB)` is typically not quoted. I'm not sure whether that's because make autoquotes `$()` substitutions in shell commands, or because you intend `$STARTJOB` to possibly contain whitespace and some preliminary flags to be passed before the `-c`, e.g. `STARTJOB='submit_job --node=some_compute_server --priority=-5`.  Can I assume that the value of `STARTJOB` is merely the path to an executable, or could it be a fragment of shell code?

kini:
Never mind -- I just pushed a solution which should work even if `STARTJOB` is a fragment of shell code.  I just hope the multi-layered string escaping is not too unreadable...

kini:
BTW, Matt emailed me and @harshrc asking if he (Harsh) could test this on Windows, to which Harsh agreed.  Rather than replying to that thread, I'll continue that conversation here so that other people can see it.

Harsh, you asked what you need to do in order to test this.  Thanks for volunteering!  Mainly I'd like you to

1) make sure this didn't break any pre-existing functionality, i.e. try running a community books regression on Windows, and try using cert.pl on any other proof developments you may have lying around outside of the community books which you know were working before this patch.

2) make sure the added functionality works as intended, i.e. create a script that execs ACL2 by reference to its absolute path, then put that script in some unrelated location far away from the actual ACL2 executable, set `$ACL2` to contain the filepath of the script, and then try to use cert.pl to certify something.

Here's an example script, though I don't know if you have to do anything slightly different to get it to work under cygwin:

```sh
#!/bin/sh
exec /absolute/path/to/acl2 "$@"
```

(To others: any ideas for what else needs to be tested?)

Thanks!

MattKaufmann:
> (To others: any ideas for what else needs to be tested?)

Mac, definitely.  If you don't have a Mac and can send me files to replace or add (I don't want to mess with github), I'll test it for you on my Mac.

kini:
Thanks, Matt.  You can wget the cert.pl file from https://raw.githubusercontent.com/kini/acl2/cert-pl-find-books-harder/books/build/cert.pl -- it replaces `books/build/cert.pl` in the repository.

MattKaufmann:
Thanks, Keshav.  OK, I've done a quick test on my Mac, and it seems to be working fine.  I'll let you know if I encounter any problems when running a more complete test (probably overnight), but I don't expect any problems given the success of the quick test.

harshrc:
1a. Whole regression. Not tried.
1b. Certify some books lying outside of community books. Tried. Works!

Btw when you say 'execs acl2', or you say 'ACL2 executable', it can be
confusing [1]. I guess by latter you mean the saved_acl2 script that
executes a lisp runtime executable (lx86cl64) and loads it with the ACL2
heap image (saved_acl2.lx86cl64).

2. Copied sources/acl2/saved_acl2 into ~ and used ACL2=~/saved_acl2 to
certify sources/acl2/books/misc/dijskstra-shortest-path.lisp. Worked!
   Also tried having your example script placed in random location invoking
the forementioned saved_acl2 script and then calling cert.pl on
dijkstra-shortest-path from some tmp directory using absolute names. Worked!

By the way placing saved_acl2 (i.e the script or program that runs/execs
the ACL2 theorem prover) outside its usual location where it is co-located
with the books breaks the assumption that scripts I use have. Its good you
can get away with it. Its also nice that the cert.pl modification does not
parse the source of the saved_acl2 script and uses (@ system-books). I
looked at the changes, the first two options to find ACL2_BOOKS are good,
the third one:
+#   - set it to the parent of the directory containing this script.
has a use case, because of books/build/cert.pl, but probably we should
avoid it?
What if we have multiple ACL2  regressions, and we are testing/certifying
them and collecting some statistics with a third ACL2 build, will that
cause some inconsistent stats/results?
Will we ever fail the second option, in the first place, and fallback to
the third option?
What if cert.pl itself is found useful and someone moves it to a separate
location for a separate utility task. Why add another assumption to the
code of cert.pl?
Some questions for thought, none of which affect me, so I leave it to
others.

[1] By confusing I mean: acl2 can be a environment variable, alias, script
invoking lx86cl and the saved_acl2.image, a script invoking another script
invoking *, file name, binary program, link, etc etc etc

kini:
On 04/03/2017 03:32 PM, harshrc wrote:
> Btw when you say 'execs acl2', or you say 'ACL2 executable', it can be
> confusing [1]. I guess by latter you mean the saved_acl2 script that
> executes a lisp runtime executable (lx86cl64) and loads it with the ACL2
> heap image (saved_acl2.lx86cl64).

Right, that is what I meant by "ACL2 executable".

> By the way placing saved_acl2 (i.e the script or program that runs/execs
> the ACL2 theorem prover) outside its usual location where it is co-located
> with the books breaks the assumption that scripts I use have. Its good you
> can get away with it. 

Are those some personal scripts of your own?

At Oracle we have a pretty indirect way of calling acl2, which motivated
my patch.

> I
> looked at the changes, the first two options to find ACL2_BOOKS are good,
> the third one:
> +#   - set it to the parent of the directory containing this script.
> has a use case, because of books/build/cert.pl, but probably we should
> avoid it?
> What if we have multiple ACL2  regressions, and we are testing/certifying
> them and collecting some statistics with a third ACL2 build, will that
> cause some inconsistent stats/results?

Yes, I agree, it should probably be avoided.  But it was already there
in cert.pl so I just left it in.  If people agree, I can remove it.

> Will we ever fail the second option, in the first place, and fallback to
> the third option?

Hmm.  I guess if you move the entire books/ directory somewhere else,
then the value reported by `(@ system-books-dir)` will be a nonexistent
path, so cert.pl will fallback to the third option, which will be
wherever you moved books/ to, assuming you're running the cert.pl which
lives in that copy of the books.  Then cert.pl will set
ACL2_SYSTEM_BOOKS to the correct path of the books, and then when ACL2
is actually run to do the certification, it will have the correct value
of `(@ system-books-dir)` again.

> What if cert.pl itself is found useful and someone moves it to a separate
> location for a separate utility task. Why add another assumption to the
> code of cert.pl?

Dunno.  I think the only case where the third option will actually be
used is when acl2 can't find any existing books directory without extra
help from cert.pl.  So I think it can't hurt.  Anyway, it was already
there before so I just left it in, as I said earlier.

-Keshav

solswords:
I'm good with this, pending complaints by anyone for whom it breaks things.  Any further objections?

harshrc:
>
>
> Are those some personal scripts of your own?
>
> Yes, some ACL2s-specific and also some other.

>
> Hmm. I guess if you move the entire books/ directory somewhere else,
> then the value reported by `(@ system-books-dir)` will be a nonexistent
> path, so cert.pl will fallback to the third option, which will be
> wherever you moved books/ to, assuming you're running the cert.pl which
> lives in that copy of the books. Then cert.pl will set
> ACL2_SYSTEM_BOOKS to the correct path of the books, and then when ACL2
> is actually run to do the certification, it will have the correct value
> of `(@ system-books-dir)` again.
>
> Ahh I see. Thanks for pointing that.
I wonder how system-books-dir is calculated; I am guessing everything
depends
on where the heap image resides or the directory is which it was first
created.

>
> Dunno. I think the only case where the third option will actually be
> used is when acl2 can't find any existing books directory without extra
> help from cert.pl. So I think it can't hurt. Anyway, it was already
> there before so I just left it in, as I said earlier.


If it was there, best leave it. I thought it was added, because of the ++
in the git diff.
Also it has come back to me that ACL2 image and the books are truly
independent, the third option provides the link via cert.pl. I spoke too
soon..

Ohh btw from my side, your changes are good on Cygwin. I didnt test the
full regression, but I dont think testing that separately is necessary. (I
ought to also test absolute path locations from different drives, i.e. D:
vs C:, cygwin can act funny with its cygdrive discipline, but I dont have
access to such a machine right now; and ohh theres also msys2, but best not
to even think of supporting anything other than cygwin on windows)

kini:
On 04/03/2017 06:42 PM, harshrc wrote:
> I wonder how system-books-dir is calculated; I am guessing everything
> depends
> on where the heap image resides or the directory is which it was first
> created.

From a bit of poking around in interface-raw.lisp it seems that first it
is read from the environment variable $ACL2_SYSTEM_BOOKS, but if that
isn't set, then it assumes it's in ./books relative to the current
directory.  (Correct me if I'm wrong...)

>> Dunno. I think the only case where the third option will actually be
>> used is when acl2 can't find any existing books directory without extra
>> help from cert.pl. So I think it can't hurt. Anyway, it was already
>> there before so I just left it in, as I said earlier.
> 
> 
> If it was there, best leave it. I thought it was added, because of the ++
> in the git diff.

Ah, that's because I moved that comment from a different position in the
file.  If you scroll up in the diff you'll see where I removed it.  (The
comment was also incorrect earlier -- it used to say "the directory
containing this script", presumably because cert.pl used to be directly
in books/ instead of books/build/ .)

> Ohh btw from my side, your changes are good on Cygwin. I didnt test the
> full regression, but I dont think testing that separately is necessary. (I
> ought to also test absolute path locations from different drives, i.e. D:
> vs C:, cygwin can act funny with its cygdrive discipline, but I dont have
> access to such a machine right now; and ohh theres also msys2, but best not
> to even think of supporting anything other than cygwin on windows)

OK, thanks for your help!

-Keshav

MattKaufmann:
I successfully ran the following on my Mac with the modified cert.pl, so I'm happy.

    make clean-books ; (time nice make -j 8 regression-everything USE_QUICKLISP=1 ACL2=/Users/kaufmann/acl2/acl2-git-scratch/sbcl-saved_acl2) >& make-regression-everything-sbcl-quicklisp-j-8.log&

MattKaufmann:
Harsh, since you asked:

>> I wonder how system-books-dir is calculated

That state global is set when ACL2 is built (specifically, in 
enter-boot-strap-mode, which is called by initialize-acl2).

-- Matt

MattKaufmann:
Following up on my previous reply: the following from Keshav is also
correct (that read happens in initialize-acl2 before passing the
directory to enter-boot-strap-mode).

>> From a bit of poking around in interface-raw.lisp it seems that first it
>> is read from the environment variable $ACL2_SYSTEM_BOOKS, but if that
>> isn't set, then it assumes it's in ./books relative to the current
>> directory.  (Correct me if I'm wrong...)

Note that this is all low-level stuff that, I hope, nobody on this
list needs to know -- I'm only commenting because it came up.

-- Matt

PR #712, opened 2017-04-11
Make patbind-assocs easily generalizable

kini:
This commit generalizes and factors out some of the guts of
patbind-assocs so that they can later be reused to create similar
patbind functions for other (possibly user-defined) alist-backed data
structures.

Suppose you have a datastructure which is backed by an alist but is
designed to be used via accessor and updater functions, which are
analogous to `assoc` and `acons` but do something slightly more
complicated.

Imagine for example an alist representing a polynomial by mapping
monomials to their coefficients, where only non-zero coefficients are
actually stored.  Then you might have

    (defun assoc-polynomial (x alist)
      (or (assoc x alist)
          (cons x 0)))

    (defun acons-polynomial (key datum alist)
      (if (equal datum 0)
          (delete-assoc key alist)
        (acons key datum (delete-assoc key alist))))

This commit will allow you to write a patbind function using
`def-b*-binder` so that you can grab multiple coefficients from a
polynomial in a single `b*` binding:

    (def-b*-binder polynomial-assocs
      :body (body-for-assocs 'assoc-polynomial
                              args forms rest-expr))

    (b* (((polynomial-assocs (a1 1) (a3 3) (a5 5))
          my-polynomial))
      (f a1 a3 a5))

### DIFF
--- a/books/std/util/bstar.lisp
+++ b/books/std/util/bstar.lisp
@@ -977,16 +977,29 @@ and @('list*') may be nested inside other bindings.</p>"
       `(patbind ,(car args) ,forms ,rest-expr)
     `(patbind-cons (,(car args) (list* . ,(cdr args))) ,forms ,rest-expr)))
 
-(defun assigns-for-assocs (args alist)
+(defun assigns-for-assocs (getter args alist)
   (if (atom args)
       nil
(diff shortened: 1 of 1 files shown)


solswords:
Seems good, thanks!

PR #713, opened 2017-04-13
Shnf: parameterized by subring of complex numbers

nadezhin:
Functions and theorems in `shnf/support.lisp`  are parameterized by subring of complex number.
`shnf\top.lisp` is almost without changes.

### DIFF
--- a/books/projects/shnf/support.lisp
+++ b/books/projects/shnf/support.lisp
@@ -2215,6 +2215,1006 @@
 ;;                            Instantiation Macro
 ;;*********************************************************************************
 
+(defn gen-basic (p list polyp evalp shfp evalh shf-normp shnfp
+                   closure-evalp closure-evalh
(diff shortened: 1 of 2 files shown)


ragerdl:
Nice, @russinoff What do you think?

russinoff:
@nadezhin I think it's great that you have found a use for the shnf book and that you have extended it to a more general class of rings.  It's important to me, however to retain the simpler version, and in particular, the original definition of evalh-witness without reference to the complex numbers, which is consistent with my provincial view of the ACL2 world and the very primitive theory that is the basis of all of my work.  Would it be possible to leave shine/top intact and include a separate function (in shnf/support) for the complex case?

nadezhin:
@russinoff, I returned your integer code. It is in the first part of `support.lisp`. The second part of `support.lisp` contains generic code with only `real/rationalp` and `acl2-numberp` instantiations.

nadezhin:
Work on rational polynomials became less important for me. I withdraw this pull request until I or  somebody need rational polynomials again.

PR #714, opened 2017-04-17
RTL: mbe implementation of rtl::expo .

nadezhin:
Logic implementation of `rtl::expo` multiply argument by `2` or `1/2` in loop
until it fits into `[1,2)`. Shift is computationally cheap operation, but
memory is allocated by each intermediate result. Amount of memory allocated
is `O( |e|^2 )` where `e` is a value of `(expo x)` .

New :exec implementation uses integer-length function to estimate `e` and then
performs comparison between shifted numerator and denominator to refine `e`.
Amount of memory allocated is `O( |e| )` .

`rtl/rel11/rel9-rtl-pkg` contains multiple copies of definition of `rtl::expo`.
All of them have been updated.

Thanks to Vladimir Hodukin for idea of this change.


ragerdl:
Hi @nadezhin ,

Neat.  There are a lot of new calls of :verify-guards nil.  In a way, this strikes me as backwards progress.  Maybe it's a "two steps forward, one step backward" situation.  However, is there a reasonable way to avoid the :verify-guards nil argument?

Thanks,
dLr

nadezhin:
Guards of all functions with `:verify-guards nil` in `rtl/rel11/support/definition.lisp` are verified in file `rtl/rel11/support/verify-guards.lisp` where proofs can use necessary lemmas from `rtl/rel11/support/float.lisp`.

This is assured by fact that there are no `:verify-guards nil` in public `rtl/rel11/lib` books.

ragerdl:
Ah, great!

@russinoff this is a second pull request that could use your feedback

russinoff:
I am quite ok with this.

PR #718, opened 2017-04-21
added rtl/rel11/support/interval.lisp

russinoff:


solswords:
It looks like this is an empty book -- am I missing something? Happy to merge if that's what's intended.

russinoff:
True, there's not a lot of content.  Dmitry asked for this because he wants
to add a book.  His legal department allows to edit files, but not create
them.  These guys don't have to make sense.

ragerdl:
We use whether a file is new as a heuristic to determine whether the contribution constitutes a bugfix or relatively minor improvement versus a more major contribution that requires management approval.  This scheme is a bit inconsistent with the "spirit of the deal," and I'm going to go ahead and remove the book that was created.  Our apologies to the community for the confusion.

On a broader note, it is critical to our community's viability that our ACL2 tool and books be able to be used without fear of legal action.  This is what motivates the stringent book-keeping.

PR #726, opened 2017-05-24
Improve git command for grabbing current git hash

kini:
From IRC:

    <kini> regarding the git invocation in :doc making-system-changes, I
           think a better one would be `git rev-parse HEAD`
    <kini> that will really just give you the commit ID and nothing else
    <kini> `git log` is subject to all kinds of user customizations
    <kini> for example, on my machine, here's what `git log -n 1
           --name-only` shows:
    <kini> $ git log -n 1 --name-only | cat
    <kini> commit 61f94d63a047bd8e5e7f59b99b30f707488d8ef1 (HEAD ->
           refs/heads/master, refs/remotes/upstream/testing,
           refs/remotes/upstream/master, refs/remotes/upstream/HEAD)
    <kini> Author: Alessandro Coglio <coglio@kestrel.edu>
    <kini> Date:   Mon May 22 11:07:24 2017 -0500
    <kini>     Add a few XDOC parent links.
    <kini> books/kestrel/utilities/defmacroq.lisp
    <kini> books/kestrel/utilities/defthmr.lisp
    <kini> books/kestrel/utilities/make-termination-theorem.lisp
    <kini> in contrast,
    <kini> $ git rev-parse HEAD
    <kini> 61f94d63a047bd8e5e7f59b99b30f707488d8ef1


ragerdl:
Nice idea to use rev-parse!  git rev-parse is certainly a more efficient way to get the hash of the most recent commit.  Matt may like having the extra information as a way to confirm that the hash is indeed what he thinks it is.  I'll leave this one up to him.

PR #731, opened 2017-05-30
Make a release-notes-books index doc topic

kini:


ragerdl:
Thanks for the original contribution @shigoel and the cleanup @kini

PR #732, opened 2017-05-30
Fix ld warning in regression and quicklisp workaround on Windows

harshrc:


ragerdl:
It's a bit unfortunate that there are so many feature names for windows, but this looks reasonable to me.  Merging.

PR #735, opened 2017-06-02
reuse mswindows feature in quicklisp raw files. Misc minor changes in acl2s books.

harshrc:


ragerdl:
Thanks for cleaning that up, @harshrc .  Merging.

PR #736, opened 2017-06-03
Adding asdf to quicklisp libs that are updated

ragerdl:
@solswords @kini does this look right to you?

### DIFF
--- a/books/centaur/quicklisp/update-libs.lsp
+++ b/books/centaur/quicklisp/update-libs.lsp
@@ -56,7 +56,8 @@
 ;;   "http://beta.quicklisp.org/dist/quicklisp/2015-03-02/distinfo.txt"
 ;;   :replace t)
 
-(ql:bundle-systems (list "bordeaux-threads"
+(ql:bundle-systems (list "asdf"
+                         "bordeaux-threads"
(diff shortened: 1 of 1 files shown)


ragerdl:
The main problem is that I'm trying to figure out how to get the latest version of asdf.  Even with this change, it appears I'm still only getting 2.26.  

```
==================================================
838 bytes in 0.01 seconds (94.87KB/sec)
; Fetching #<URL "http://beta.quicklisp.org/client/2017-03-06/quicklisp.tar">
; 250.00KB
==================================================
256,000 bytes in 0.09 seconds (2865.56KB/sec)
; Fetching #<URL "http://beta.quicklisp.org/client/2015-09-24/setup.lisp">
; 4.94KB
==================================================
5,054 bytes in 0.01 seconds (831.74KB/sec)
; Fetching #<URL "http://beta.quicklisp.org/asdf/2.26/asdf.lisp">
; 194.07KB
==================================================
...
```

ragerdl:
Actually, the asdf.cert failure was in the wrong part of the log.  It's still only getting 2.x instead of 3.x, but I'm not going to worry about it now.

Closing request without merging.

PR #740, opened 2017-06-17
Remove natp-listp from ordinals/lexicographic-book

kini:
There's a perfectly good nat-listp function built into ACL2 which could be used instead.  I also made one of the theorems local, since it's not a theorem about symbols introduced in the book and seems irrelevant to the purpose of the book (other than that it's used in the proofs of the main theorems in the book).

### DIFF
--- a/books/coi/defung/defung.lisp
+++ b/books/coi/defung/defung.lisp
@@ -385,7 +385,6 @@
 	    acl2::d<  ;; ??
 	    l<
 	    acl2::lexp
-	    acl2::natp-listp
 	    nfix
 	    not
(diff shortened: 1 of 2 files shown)


MattKaufmann:
The rule len-0 can be pretty useful, I think.  I wonder if anyone using the ordinals books relies on it and would have problems if it's made local.  I have no objection though to this pull request being accepted.  I've made a note to accept this pull request on Tuesday or perhaps late-ish Monday if nobody objects.

ragerdl:
@MattKaufmann Shall I accept this?

PR #742, opened 2017-06-20
Add book std/io/open-lemmas

kini:
Quoting from the docstring in the code:

      When programming with I/O, you may need to write functions that
      modify state in some way (perhaps by doing I/O) and exit
      returning a state which has some channels open, perhaps for
      further reading and writing later in your program.  Then, to
      prove guard theorems elsewhere, you'll need returns theorems
      about your function showing that those channels are open, either
      because they were open before the function ran and the function
      didn't close them, or because the function itself opened
      them.

      This book contains lemmas that can help you prove such returns
      theorems.  For each built-in I/O function, there is a lemma
      saying that, under appropriate hypotheses, it doesn't close your
      open input or output channels.

I also added this book to the std/io/top book, and added a note to the ACL2 developers' to-do list document pointing to this book, because improving this book was the motivation for one of the to-do list items in there.  I also fixed a couple of unrelated typos elsewhere.

Thanks to @airbornemihir for a lemma I stole from his file system project :P

Some review items that come to mind:

1) Is what I claim in the docstring right, or is there some better way of writing programs that do I/O with channels that are left open long-term?

2) Did I screw up the xdoc stuff?  I must admit I don't understand xdoc all that well, but loading the book into an ACL2 prompt and typing `:doc open-channel-lemmas` seemed to work as expected...

3) Did I miss any builtin I/O functions that could use similar lemmas about them?


kini:
Added a note to :DOC note-7-5-books. (Is this a sufficiently significant change to merit a mention there?)

ragerdl:
Nice.  I'll go ahead and merge, because I'm guessing this is strictly progress.  If it turns out that this isn't quite the idea of `std/io`, it'll be a good learning experience for us.  Anyway, I answer your questions below:

> Is what I claim in the docstring right, or is there some better way of writing programs that do I/O with channels that are left open long-term?

Really @jaredcdavis or @MattKaufmann are better qualified to answer this one.  I'm guessing what you've done is at least fine and probably good.

> Did I screw up the xdoc stuff? I must admit I don't understand xdoc all that well, but loading the book into an ACL2 prompt and typing :doc open-channel-lemmas seemed to work as expected...

I think the main thing to look for is warnings emitted from the ACL2 prompt while building the manual.  The next checks are how it looks in the emacs manual and in the web manual.

> Did I miss any builtin I/O functions that could use similar lemmas about them?

@MattKaufmann may know.

> Is this a sufficiently significant change to merit a mention there?

Yep, thanks!

MattKaufmann:
Looks nice.  I don't have anything to add about (1); on quick glance, your lemmas look good.  For (2): I think it's sufficient for you to include your book and then run :doc on each new topic at the terminal, to see that the output looks OK.  (That's what I do, and I think everyone should do that check on new xdoc strings.)  For (3): I don't know offhand, but for candidates, I'd just look at the :logic mode functions in the lists "Input Functions" and "Output Functions" in :doc io.

kini:
Thanks to both of you for the comments.  Matt: for (3), that is what I did.  Unfortunately it looks like I did screw up the xdoc, by creating two broken links (to be addressed in an upcoming PR). Running `:doc` at the prompt doesn't seem to detect broken links, while I guess building the manual would have, so that's something to keep in mind in the future.

PR #743, opened 2017-06-20
Fix XDOC errors

kini:
`open-input-channel-p1` and `open-output-channel-p1` have no doc
topics about them, so I changed the @(tsee)s to @({})s.

This should fix a build failure caused by #742.


ragerdl:
Thanks for following up

PR #748, opened 2017-06-27
Tweak some docs and some docs about docs

kini:
I added a couple sentences to the docs about XDOC to address something that I personally found confusing, in the hope that it will help others in the future.

### DIFF
--- a/books/doc/relnotes.lisp
+++ b/books/doc/relnotes.lisp
@@ -110,7 +110,7 @@
 
  <p>The @(see std/io) library now contains lemmas to help users prove
  that opened input and output channels remain open until closed, to
- aid guard theorem proofs.</p>
+ aid guard theorem proofs.  See @(see open-channel-lemmas).</p>
 
  <h3>Licensing Changes</h3>
 
--- a/books/std/io/open-channels.lisp
+++ b/books/std/io/open-channels.lisp
@@ -103,7 +103,7 @@
 ;;;; Open input channels stay open
 ;;;;;;;;;;;;;;;;
 
-  "<p>First are lemmas about @({open-input-channel-p1}).</p>"
+  "<p>First are lemmas about @('open-input-channel-p1').</p>"
 
   (defthm open-input-channel-p1-under-open-input-channel
     ;; Desired version:
@@ -182,7 +182,7 @@
 ;;;; Open output channels stay open
 ;;;;;;;;;;;;;;;;
 
-  "<p>Next are lemmas about @({open-output-channel-p1}).</p>"
+  "<p>Next are lemmas about @('open-output-channel-p1').</p>"
 
   (defthm open-output-channel-p1-under-open-input-channel
     (implies (open-output-channel-p1 channel type state)
--- a/books/xdoc/topics.lisp
+++ b/books/xdoc/topics.lisp
@@ -224,6 +224,11 @@ and \"preformatted,\" i.e., newlines and spaces should be preserved.</li>
 because they automatically escape special HTML characters like &lt; into
 &amp;lt; and also automatically add hyperlinks to documented functions.</p>
 
+<p>(Note that the @('<tt>') and @('<code>') tags in XDOC do not correspond to
+the @('<tt>') and @('<code>') tags in HTML.  In fact, @('<tt>') is rendered
+with an HTML @('<span>'), and @('<code>') is rendered with an HTML
+@('<pre>').)</p>
+
 <p>Whenever you include Lisp code fragments in your documentation, you should
 usually keep everything <b>indented one space</b> to prevent Emacs problems.
 For instance:</p>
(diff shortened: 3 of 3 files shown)


MattKaufmann:
I don't mind accepting this pull request, but I find the discussion of "span" unhelpful.  Even if <tt> is rendered with <span> (I'll take your word for that), the little I've googled about "span" tells me nothing useful.  I've been able to write tons of xdoc without knowing that <tt> isn't rendered as <tt>; so is it really helpful to mention "<span>"?  For one who is writing xdoc, that seems to me to be too low-level; instead it should, I think, suffice to say that "<tt>" renders with typewriter font inline, while "@({..})" creates displays in typewriter font, etc.

kini:
On 06/26/2017 10:44 PM, MattKaufmann wrote:
> I don't mind accepting this pull request, but I find the discussion
> of "span" unhelpful.  Even if <tt> is rendered with <span> (I'll
> take your word for that), the little I've googled about "span" tells
> me nothing useful.  I've been able to write tons of xdoc without
> knowing that <tt> isn't rendered as <tt>; so is it really helpful to
> mention "<span>"?  For one who is writing xdoc, that seems to me to
> be too low-level; instead it should, I think, suffice to say that
> "<tt>" renders with typewriter font inline, while "@({..})" creates
> displays in typewriter font, etc.

If a reader has no preconceived notions about tags named <tt> and
<code>, she certainly doesn't need this paragraph in order to write
XDOC.  But I wasn't confused because I didn't have an idea of what
<tt> or <code> do in XDOC -- I was confused because I had gotten into
my head an *incorrect* idea of what <tt> and <code> did in XDOC.  I am
familiar with HTML (as are many people, I would wager), and in HTML
there are tags named <tt> and <code> which do something different from
what <tt> and <code> do in XDOC.

So the purpose of this text is not to try to explain how XDOC works by
way of analogy to HTML.  Rather, it is to disabuse the reader of the
preconceived notion that the <tt> and <code> XDOC tags correspond to
the HTML <tt> and <code> tags -- a notion that is all the easier to
espouse when the XDOC markup language is described in :DOC
xdoc::markup as being "an XML markup language that is similar to a
subset of HTML".  The second sentence, about the <span> and <pre>
tags, is just an answer to the natural question that the reader might
ask after reading the first sentence, namely "Well if they don't
correspond to those HTML tags, what HTML tags *do* they correspond
to?"

I believe that many readers are familiar with HTML, as it is one of
the dominant computer languages today.  For readers who are not
familiar with HTML, the paragraph can be safely skipped, but I think
it can be useful to those readers who are.  If nothing else, the
arrogant reader (*cough*) who, upon seeing the sentence about how XDOC
is "an XML markup language that is similar to a subset of HTML",
blithely skims over the rest under the illusion that he already knows
the meaning of any token inside angle brackets may find that this
cautionary paragraph arrests his gaze and causes him to go back and
read a bit more carefully... :P

Technical aside: in actuality, the XDOC <tt> tag is semantically
similar to the HTML <tt> tag, but the XDOC <code> tag differs from the
HTML <code> tag in that the XDOC <code> tag is a block-level element
while the HTML <code> tag is phrase-level / inline.

-Keshav

MattKaufmann:
Yes, I understood your point, and I agree with its sense.  My concern
is only that the mention of the relatively obscure HTML tag, span,
might be off-putting to some.  (Maybe it's not as obscure as I think,
but even with 25 years of writing lots of vanilla HTML I didn't know
about it and I still don't understand it.  So that comment would
give me pause if I hadn't already written xdoc and seen it come
out OK.)

But it's not that big a deal, so I'll let this drop.  If someone
complains later we can revisit.  I guess I'm just nervous since rather
few of us have contributed xdoc so I want to minimize barriers.
For now I'll accept the pull request.

kini:
On 06/27/2017 03:24 AM, MattKaufmann wrote:
> Yes, I understood your point, and I agree with its sense.  My concern
> is only that the mention of the relatively obscure HTML tag, span,
> might be off-putting to some.  (Maybe it's not as obscure as I think,
> but even with 25 years of writing lots of vanilla HTML I didn't know
> about it and I still don't understand it.  So that comment would
> give me pause if I hadn't already written xdoc and seen it come
> out OK.)
>
> But it's not that big a deal, so I'll let this drop.  If someone
> complains later we can revisit.  I guess I'm just nervous since rather
> few of us have contributed xdoc so I want to minimize barriers.
> For now I'll accept the pull request.

Thanks!  I guess I mistook the thrust of your email -- if it's just
the sentence about <span> and <pre> in particular that bothers you, I
can remove it, or add a disclaimer saying "For those who are
interested, ", for example.

About the obscurity of <span> -- as I understand it, <span> and its
friend <div> are actually among most common HTML elements these days.
They represent semantically empty inline and block-level elements,
respectively, so when paired with the id and class attributes, they
form good building blocks for "from-scratch", javascript- and css-
based visual design which doesn't depend on meaning-carrying tags like
<p>, <h1>, etc.  Especially with the explosion of javascript in the
past decade, HTML and its traditional semantic tags are more and more
often shoved "under the rug" and made into a mere substrate that js
uses to build the page on the fly, often with no relationship to
HTML's original structured text document format of multi-level headers
and paragraphs.

-Keshav

MattKaufmann:
Hi, Keshav --

Thanks for the reply.  OK, I think we've converged; I plan to edit
that paragraph so that it starts with "Implementation Note."

-- Matt

PR #750, opened 2017-07-07
verify guards in std/util/generate-symbol.lisp

nadezhin:


nadezhin:
Close because forgot about acl2::testing

PR #753, opened 2017-07-09
Fix XDOC broken link

kini:
Thanks to Matt for the heads-up.  Sorry for the second XDOC-related breakage I've caused in the last couple weeks.  My excuse is that I'm unfortunately unable to test building the full manual since my laptop doesn't have enough memory...

### DIFF
--- a/books/centaur/fty/top.lisp
+++ b/books/centaur/fty/top.lisp
@@ -213,7 +213,8 @@ theorems mentioned in item 5, see the @(see std::returns-specifiers) feature of
 <p>Having unconditional return types and congruences is beneficial in and of
 itself.  But the main advantage of using the fixtype discipline is that in
 complex programs, program reasoning can be done while largely avoiding
-extensive @(see backchaining) involving proofs about type information.</p>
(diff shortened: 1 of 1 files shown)


ragerdl:
While we'd like our build to stay working, the purpose of the testing branch and jenkins is to catch stuff like this.  So, it's not the end of the world, so long as it gets patched up pretty quickly.  Thanks for fixing it.

MattKaufmann:
I agree.  And I don't blame Keshav at all -- it's awkward to test for
broken links.

With a modest effort, it may be possible to generate a file that's
maintained by github that defines a constant containing all the :doc
topic names.  Then the implementation of :doc at the terminal could
presumably be adjusted to check for broken links.  I realize that
generated files aren't generally under version control, but this might
be a useful exception.  I build the manual so often that I could
regenerate that file, keeping it reasonably up-to-date.  (For example,
I see that I've updated
http://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/index.html
eight times in the past month.)

But I don't know if anyone cares to spend the cycles to do such xdoc
implementation work.

-- Matt

PR #754, opened 2017-07-27
Generalize `why` and `why-explain`

kini:
Starting with a7ad6e8b7f80ca87777d6ff679975efd3d21118c (in 2015),
`:monitor` has been able to accept symbols and some other runic
designators instead of just runes, as before.  So `why` and
`why-explain` need no longer wrap symbol names with `(:rewrite ...)`.
This generalizes `why` and `why-explain` by allowing one to monitor
linear rules as well, for example.


ragerdl:
Great!  Now I don't have to worry about releasing why-linear and why-explain-linear; thanks!

PR #759, opened 2017-08-10
Fixing typo.

pennyan:
Fixing a typo that when trying to print two \\'s, we need to escape them using \\'s too.

### DIFF
--- a/books/xdoc/topics.lisp
+++ b/books/xdoc/topics.lisp
@@ -1672,12 +1672,20 @@ web-based viewer, allowing you to typeset basic formulas.</p>
 
 <h3>Basic Usage</h3>
 
+<box><p><b><color rgb='#ff0000'>ESCAPING WARNING</color></b>. LaTeX-style
+formulas may be especially hard to type in ordinary ACL2 string literals
+because you have to escape all the backslashes.  For instance, you have to
+remember to write @('\\\\sqrt{x}') instead of @('\\sqrt{x}').  You can avoid
+this headache by using the @(see acl2::fancy-string-reader). In the rest of
+this document, we will use the concrete xdoc syntax without the escapes for
+simplicity.</p></box>
+
 <p>To typeset block-style formulas, you can use the @('@([...])') @(see
 preprocessor) directive, e.g.,:</p>
 
 @({
     @([
-       c = \\\\pm\\\\sqrt{a^2 + b^2}
+       c = \\pm\\sqrt{a^2 + b^2}
     ])
 })
 
@@ -1691,7 +1699,7 @@ preprocessor) directive, e.g.,:</p>
 instance:</p>
 
 @({
-     The product @($a \\\\times b$) is even.
+     The product @($a \\times b$) is even.
 })
 
 <p>Produces text such as: the product @($a \\times b$).</p>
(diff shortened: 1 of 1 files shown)


kini:
I guess you mean \'s. Looks good to me 👍

pennyan:
Thanks for catching it! Fixed!

MattKaufmann:
There are actually several other occurrences of "\\" that could be changed the same way.  Perhaps, though, this issue is already covered by the following.

    <p>LaTeX-style formulas may be especially hard to type in ordinary ACL2 string
    literals because you have to escape all the backslashes.  For instance, you
    have to remember to write @('\\\\sqrt{x}') instead of @('\\sqrt{x}').  You can
    avoid this headache by using the @(see acl2::fancy-string-reader).</p>

I think a better fix is simply to move that paragraph forward in :doc katex-integration.  I don't think it's right to change just one line.

pennyan:
Hi @MattKaufmann , I failed in understanding your message. The other occurrences of "\" seems fine to me. What do you mean by move the paragraph forward? Like before the example?

MattKaufmann:
Actually I guess there is just one more occurrence of "\" that should
be changed the same way, not several (sorry).  It's this one:

    @({
	 The product @($a \\times b$) is even.
    })

It should, I think, be:

    @({
	 The product @($a \\\\times b$) is even.
    })

I guess the "Escaping Help" section (which I displayed previously) can
stay where it is.

jaredcdavis:
Hi,

This documentation is unfortunately confusing.  I think Matt's suggestion of moving something like the "escaping help" section to the beginning is a good idea, perhaps with some kind of large warning label.

I think the documentation as written, showing single backslashes when rendered, is correct.  But it is definitely only correct if the reader understands it to be describing the concrete xdoc syntax, instead of as a fragment of Common Lisp string literal that is going to become xdoc syntax.

This is similar to the problem of how to describe a Windows-style path in Lisp.  If you want to write a concrete path such as `C:\Windows`, you may need to write a string literal in your Lisp source file that instead looks like `"C:\\Windows"`.  Pointing out right away is probably a really good idea.

I'm afraid I can't currently make this change, but I'd definitely support changing the topic to make this more clear.

Cheers,
Jared

pennyan:
Hi @jaredcdavis , I now see what Matt meant and what you intended in the first place. I will see what I can do to amend this commit.

When using this feature, I found that the katex stuff is a bit out of date and doesn't support what's supported now, for example, \exists and \forall. I'm wondering if I should create another request for updating the Katex stuff in books/xdoc/fancy/lib/katex ? I'm thinking maybe replacing this directory with the new version would do it. Do you know if there are other dependencies I should take care of?

jaredcdavis:
Updating to the latest version of katex would probably be a fine idea.

I don't know what this would involve, but probably it would just require...
 - replacing xdoc/fancy/lib/katex with the latest
 - possibly tweaking the `renderMathFragments` code in `xdoc.js` if the KaTeX API has changed

The `books/xdoc/fancy/lib` files just get copied over when a manual is built.  So, to quickly beta-test, you could try out the above in an existing manual and avoid needing to certify any books.  It'd probably be a good idea to try it in a few browsers.

pennyan:
Hi @MattKaufmann and @jaredcdavis , I've updated with a new commit to address the Katex escaping issue with a big red warning. Let me know if you have other suggestions!

PR #762, opened 2017-08-14
[x86isa] Re-organization of memory-related functions and other misc. changes

shigoel:
Here's a summary of some notable changes to the x86isa books:

- Codewalker can now be used to reason about 64-bit user-level x86
  programs using the x86isa books --- see
  x86isa/proofs/codewalker-examples for two simple demos.

- Memory functions do not traffic in lists anymore, and consequently,
  their signatures have changed.  Instead of a list of canonical
  addresses, a contiguous linear memory region is now specified by:
  <n, lin-addr>, where n is the number of bytes to be read and
  lin-addr is the first address of the memory region.

  A non-exhaustive list of affected functions is as follows:

    * las-to-pas
    * rb
    * wb
    * program-at
    * all-xlation-governing-paddrs

  Note that this change does limit address translation using
  las-to-pas to contiguous linear memory regions at a time.  However,
  it's rare to want to do address translation for non-contiguous
  linear memory regions in one swoop, so this is probably fine.

  It'd be nice if las-to-pas could analogously traffic in physical
  memory regions --- that is, instead of trafficking in lists of
  physical addresses, <n, phy-addr> could specify a region of physical
  memory that corresponds to <n, lin-addr>.  However, it's not quite
  so straightforward because a contiguous linear memory region can map
  to multiple non-contiguous physical memory regions, depending on the
  page table configuration.

- In the programmer-level mode, disjointness of memory regions can be
  conveniently expressed using a function called separate (in
  proofs/utilities/disjoint.lisp).  All the proofs in the
  programmer-level mode have been updated to use this paradigm.

- Sped up symbolic simulation of programs in the programmer-level mode
  and non-marking system mode by making some rules more efficient ---
  reducing the number of hypotheses, writing better "opener" lemmas
  (e.g., for get-prefixes), having better bind-frees, and changing the
  definitions of functions when appropriate.

  An illustrative example is that of program-at. Previously,
  program-at was defined in terms of combine-bytes.  It is now defined
  in terms of a new function, combine-n-bytes.  This new function is
  analogous to bitops::part-select, but it traffics in numbers instead
  of lists. This change lets us define more efficient RHSes of rules
  like many-reads-with-rb-from-program-at, which is used to fetch
  bytes of an instruction given the (possibly symbolic) location of
  the program.  Previously, the RHS of
  many-reads-with-rb-from-program-at was absolutely grotesque:

  <pre>(equal (mv-nth 1 (rb n addr :x x86))
	               (logior (loghead 8 (logtail (ash (+ addr (- prog-addr)) 3) (combine-bytes bytes)))
                            (ash (mv-nth 1 (rb (+ -1 n) (+ 1 addr) :x x86)) 8)))</pre>

  This essentially caused many-reads-with-rb-from-program-at to be
  fired again and again till we ran out of n. Yes, grotesque.

  The RHS of many-reads-with-rb-from-program-at is much nicer now:

  <pre>(equal (mv-nth 1 (rb n addr :x x86))
	    (combine-n-bytes (- addr prog-addr) n bytes))</pre>

  During symbolic simulation of a program, we'd know the concrete
  value of bytes.  Even so, using combine-bytes instead of
  combine-n-bytes would have been expensive because the former would
  combine all program bytes whereas the latter only combines the
  necessary ones, i.e., n bytes.

  Two datapoints: proofs/popcount/popcount-general used to take ~900s
  to certify, and now it takes ~300s;
  proofs/zeroCopy/non-marking-mode/zeroCopy used to take ~170s to
  certify, and now it takes ~50s.

  There's still quite a lot that can be done to speed things up
  further by making the core unwinding rules better --- lots of
  low-hanging fruit here...

- Merge in some preliminary work from Kestrel (thanks, Eric and
  Alessandro!) to support 32-bit mode in the x86isa books.


kini:
Wow, codewalker integration! :+1: The other changes sound cool too!

shigoel:
@kini I didn't realize Github had emojis. Sort of cool, I guess. :)

ragerdl:
+1 for codewalker integration

PR #763, opened 2017-08-17
Improve :DOC compound-recognizer

kini:
The explanatory material after the table of types and primitive
recognizing expressions has been a bit off since the creation of the
type set for `1` some time ago.  Also, a restriction was referred to
as a weakening when in fact it is a strengthening, and I removed the
words "conjunction or", since a conjunction of primitive type
recognizers would just be false, which is an empty disjunction.


kini:
argh, supposed to be to testing

PR #765, opened 2017-08-18
Add paragraph about the format of patch.lisp

kini:
Based on a conversation on acl2-devel and IRC.

Also converted a bunch of occurrences of `--` to `&mdash;`. This looks better in the HTML XDOC manual, and gets converted to hyphens anyway in the text XDOC manual for `:doc` in the terminal or for the Emacs ACL2 doc browser.

Also fixed a typo in :DOC case-split.


MattKaufmann:
Looks good; I plan to accept this after I do a push in a few hours.
(ANd thanks for the perfect new paragraph for [I think] :doc
making-system-changes.)  If you have a minute before then, you could
make the following slight change.  For the following on line 49929

    - <p><i>dom</i> (``domain'') -- must be one of four symbols: @('INTEGERP'),
    + <p><i>dom</i> (``domain'') must be one of four symbols: @('INTEGERP'),

I think

    <i>dom</i>

should be:

    <i>Dom</i>

kini:
Done! (Also, yes, the new paragraph is in :DOC making-system-changes.)

PR #767, opened 2017-09-01
Remove duplicate rule in bitops library

kini:
The enabled rule `bitops::unsigned-byte-p-when-unsigned-byte-p-less`
is an duplicate of the disabled rule `bitops::unsigned-byte-p-incr`.
To simplify this state of affairs, I removed the former and enabled
the latter.

### DIFF
--- a/books/centaur/bitops/ihsext-basics.lisp
+++ b/books/centaur/bitops/ihsext-basics.lisp
@@ -2438,13 +2438,11 @@ off looking at the source code.</p>")
                     :pattern (unsigned-byte-p bits x)
                     :scheme (unsigned-byte-p-ind bits x))))
 
-  (defthmd unsigned-byte-p-incr
+  (defthm unsigned-byte-p-incr
     (implies (and (unsigned-byte-p a x)
                   (natp b)
                   (<= a b))
              (unsigned-byte-p b x)))
-
-  (local (in-theory (enable unsigned-byte-p-incr)))
 
   (defthmd unsigned-byte-p-logcons
     (implies (and (unsigned-byte-p (1- b) x)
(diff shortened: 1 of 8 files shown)


ragerdl:
These types of consolidations are really healthy for our community.  Thanks for finding this opportunity, @kini !  @solswords , I'm going to let you merge this, since it has the potential to break some Centaur books.

BTW, fwiw, the name with "less" seems more intuitive to me.  But, maybe I'm just missing the meaning of "increase" or "increment"

solswords:
I'm of course fine with this in principle, but it'd cause less churn to delete the disabled rule, plus I like the more explicit name better -- can we please do it the other way?

kini:
Hi Sol, thanks for commenting. I'm a bit surprised that you like the other name better, since there are many similar looking theorems (i.e. `(implies (and (foo a x) (<= a b)) (foo b x))`) with the naming convention `foo-incr` in the Centaur books in this repo. That's one reason I chose to keep that one and not the other.

Also the `-incr` theorem was included in a ruleset while `-when-unsigned-byte-p-less` wasn't, so I figured it was more "planned" than the other one, which seems more ad-hoc.

Also there is a `signed-byte-p-incr` but no corresponding `signed-byte-p-when-signed-byte-p-less`. (Maybe `signed-byte-p-incr` should be enabled too.)

I'm not averse to switching to the other rule, but I wonder what your response to the above is.

solswords:
OK, seems fine -- I'll merge.  Apparently I don't have any consistency from one day to the next on what kinds of names I like.

kini:
OK, thanks :) (Should `signed-byte-p-incr` be enabled?)

solswords:
I'm open to it if you want to check what it breaks.  It shouldn't be too expensive since its first hyp having a free variable will limit its application.

kini:
OK, I'll run a regression and see what happens.

PR #775, opened 2017-09-29
Make unsigned-byte-p-unsigned-byte-p a disabled rule

kini:
This way you can use it by just enabling it, rather than by needing to write a `:use` hint.

### DIFF
--- a/books/ihs/logops-definitions.lisp
+++ b/books/ihs/logops-definitions.lisp
@@ -157,12 +157,11 @@
                   (< i (expt 2 bits))))
     :rule-classes :forward-chaining)
 
-  (defthm unsigned-byte-p-unsigned-byte-p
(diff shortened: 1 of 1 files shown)


ragerdl:
Seems good to me, but I'm a little worried there's something subtle here that I'm missing.  At first glance, it looks like this rule could fire (once enabled) without having to be manually instantiated.  In case I've missed something, I'll wait for @MattKaufmann  or @solswords  to chime in/accept it.

ragerdl:
Regardless, thanks for the attempt at cleaning it up!

kini:
Well yes, the point of this PR is to make it such that this rule can fire, once enabled, without having to be manually instantiated. So your first glance seems accurate to me.

PR #776, opened 2017-09-29
Replace GPL XDOC icons with MIT/X11 and Public Domain

ragerdl:
Maybe I should make the "download from the cloud" icon a lighter blue.
At least it's electrifying.  Plus, it involves clouds, which I hear
are super in right now.


ragerdl:
See http://www.cs.utexas.edu/users/ragerdl/acl2-manual3/?topic=ACL2____TOP for a preview of the new icons

PR #781, opened 2017-10-11
 Make defpointer support package-external symbols

kini:
Previously, defpointer took its arguments and ignored what package
they were in, incorrectly using the symbols with the same names in the
current package instead.  Now it respects the package in which its
arguments lie.

In making this change, I also had to decide what the link text of the
pointer should look like, depending on what package the linked doc
topic's symbol was in.  I decided that if the `from` and `to` symbols
were both in the same package, we should leave off the package
qualification in the text, and otherwise we should include it.

I also added a guard to the macro to ensure that only symbols are
accepted.

----

Something's still kind of weird here -- if you look at the generated HTML manual, XDOC's notion of what the "base package" is for both the "acl2::counter-example-generation" and the "acl2::cgen" topics is actually "ACL2S". But at least the build still works, and the doc topics are still readable.

### DIFF
--- a/books/acl2s/cgen/top.lisp
+++ b/books/acl2s/cgen/top.lisp
@@ -239,6 +239,9 @@ package.</p>
 "
  )
 
+#!ACL2
+(defpointer counter-example-generation cgen)
+
 (defxdoc cgen::flush
   :parents (acl2::cgen)
   :short "Flush/Reset the Cgen state globals to sane values."
--- a/books/xdoc/top.lisp
+++ b/books/xdoc/top.lisp
@@ -470,20 +470,30 @@
       (concatenate-symbol-names (list ,@args))
       mksym-package-symbol)))
 
-; Moved from acl2-doc.lisp to make it more widely available
 (defmacro defpointer (from to &optional keyword-p)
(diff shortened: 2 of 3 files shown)


MattKaufmann:
Great!  Did you test whether things look OK at the terminal (or in acl2-doc)?

kini:
Things look fine in the terminal. Only the HTML manual has this "base package" concept, it seems -- look at the current manual page for `acl2::cgen`, and you'll see that in the upper right hand corner there's a little box saying "ACL2S Package", and the title at the top of the page includes "ACL2::" at the beginning: http://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/?topic=ACL2____CGEN



I didn't check acl2-doc since off the top of my head I don't know how to generate a partial manual suitable for use by acl2-doc and I didn't want to try to build the full manual on my laptop (too slow and I don't have enough memory to do it anyway).

MattKaufmann:
Thanks, Keshav -- if it looks OK at the terminal then I'm pretty sure
it will also look OK in acl2-doc.  I'll accept the pull request.

-- Matt

PR #785, opened 2017-10-23
X86isa

nadezhin:
Fix SAR instruction.
ELF reader can read ELF32.


shigoel:
Thanks, @nadezhin!
As far as I'm concerned, this PR is totally okay.

PR #786, opened 2017-10-25
Updating Katex to be Katex-0.8.3

pennyan:
Updating Katex to be most recent release Katex-0.8.3.


solswords:
Anyone have opinions on this? I don't know anything about katex -- is it
all in-browser or is it something that needs to be built in order to build
the manual?  I suspect the former because I've never had to think about it
before.

pennyan:
Hi @solswords , here's the link to the issue #759 @jaredcdavis once commented in. It seems to me that when building the manual, the files are just copied over to the manual directory.

ragerdl:
I guess my main thought is that I don't know how to evaluate this without downloading katex and thinking a bit about where the katex files should go.  I wonder a bit why there are so many new files, but that may simply be because Katex has grown since its initial import.  We could be conservative and wait until after the next ACL2 release to upgrade Katex.  But, such waiting is probably unnecessary.  So long as Yan's tested it reasonably well, I'm okay with merging it.

ragerdl:
@pennyan says she built the manual and confirmed that her katex formula was rendered correctly in it.  Assuming that's the case, it seems reasonable to me to assume that this pull request is positive progress.  Thus, I am going to go ahead and merge it.

PR #791, opened 2017-11-04
Remove ACL2_SAVE_QUICKLISP hack in Makefile

kini:
The comment says that if this hack is no longer needed by December
2015, it should be removed.  Clearly, it is already past December
2015, so we should remove it.

---

Or so confidently declares my commit message, but what do you think?

### DIFF
--- a/books/GNUmakefile
+++ b/books/GNUmakefile
@@ -529,25 +529,6 @@ quicklisp_clean:
 	@echo "Cleaning up Quicklisp files"
 	cd $(QUICKLISP_DIR); ./clean.sh
 
-# The following hack was useful briefly in late January, 2014, when
-# there were issues with quicklisp that caused us not to want to clean
-# quicklisp files.  Rager likes this hack (because his company-wide
-# ACL2 installation isn't writable by those that use it and he doesn't
-# always want to clean quicklisp since proxies have to be configured
-# and because it's good to preserve the
-# quicklisp/inst/cache/asdf-installs directory), and so he prefers
-# that the "hack" be preserved.
-
-# May 2015 update: since the quicklisp installation procedure has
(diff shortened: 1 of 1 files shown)


ragerdl:
Thanks for the `clean`up.  Did you test this with a fresh checkout?  I agree that based on my comments that it should be reasonable to remove.  If not, maybe I'll wipe out the leeroy workspace before it gets sent over to leeroy.

kini:
I didn't test it, no -- sorry about that. I can test it if you want to avoid slowing down leeroy by clearing the workspace. (Though probably it's good to clear out the workspace from time to time.)

ragerdl:
Alright, I've had time to think about this now.  It seems to me that the dependency indicated by `clean: quicklisp_clean` needs to remain.  Can you put back that one line and then update this pull request stating as such?  It should be ready to merge at that point.

books/GNUmakefile
```diff
@@ -529,25 +529,6 @@ quicklisp_clean:
...
-# quicklisp/inst/cache/asdf-installs directory), and so he prefers
-# that the "hack" be preserved.
-
-# May 2015 update: since the quicklisp installation procedure has
-# changed, the conditional part of this code is hypothetically dead.
-# We leave it commented for now, but if December 2015 comes around,
-# and it's still unnecessary, the conditional part of this (and the
-# above comments) should be deleted.
-#ifndef ACL2_SAVE_QUICKLISP
-clean: quicklisp_clean
```
ragerdl: Please put back this dependency

kini: Done.

ragerdl:
Thanks for the cleanup, I went ahead and "Squashed and merged" your pull request.

PR #793, opened 2017-11-05
Updated quicklisp; added Babel; removed external-program

ragerdl:
This builds the manual from a fresh clone of the clone.  So, I'm pretty sure it's working.

You may notice that `asdf` is no longer included by us.  I can add it back if we're using it directly via our quicklisp bundle, but I suspect we're not.  CFFI manually includes one of the `asdf` files now, so I suspect that's why we don't need to install it.

I added `Babel`, because `Alexandria` had a build error that indicated `Babel` was missing.  That dependency should have been fulfilled via some sort of quicklisp dependency mechanism.  But, it seems not to work at this moment.  Rather than chase down this quicklisp issue, I've simply included `Babel`, which seems to resolve our problem.  It looks to me like `Babel` is released under 3-clause BSD.

Making @solswords the reviewer for this change, because if anyone's still using `external-program`, it'd be him.  If anyone else can legitimately surprise me by informing me of their use of `external-program`, please do so.


solswords:
This works for me.  Anyone know what external-program is or was used for?  What was the problem with it?

kini:
Basically it has an odd license (Franz Inc. "LLGPL2.1"), so we'd like to remove it if we're not using it for anything. (Also generally it makes sense to remove quicklisp packages that are not being used, I would say.) See the recent acl2-books thread about USE_QUICKLISP.

ragerdl:
It may have been used in `xdoc` or `vl` a long time ago.  Not sure either, but it's good to prune it.

ragerdl:
Well shoot, I submitted the pull request into `master` instead of `testing`.  Sorry about that.  I ran (two) clean regressions before creating the pull request, so I don't expect there to be problems.  But, lmk if anyone runs into something.

jaredcdavis:
Hi,

I'd be really surprised if external-program was ever used for
anything.  I'm sure I added it somewhere along the line, but I can't
at all remember why.

Cheers,
Jared

On Tue, Nov 7, 2017 at 12:31 PM, David L. Rager
<notifications@github.com> wrote:
>
> Well shoot, I submitted the pull request into master instead of testing. Sorry about that. I ran (two) clean regressions before creating the pull request, so I don't expect there to be problems. But, lmk if anyone runs into something.
>
> —
> You are receiving this because you are subscribed to this thread.




-- 
11410 Windermere Meadows
Austin, TX 78759
http://www.cs.utexas.edu/users/jared/

PR #794, opened 2017-11-07
Remove drat-trim

kini:
I noticed that we're using a 2.5-year-old version of `drat-trim`. If there isn't any particular reason that we're using an old version, we might as well upgrade to the latest version since Marijn has made many improvements over the last 2.5 years.


ragerdl:
This is a good idea in principle.  However, it should be tested before we merge it to make sure it doesn't need new flags (or something else surprising).  If anyone tests the new drat-trim, lmk, and I'll merge this PR.

kini:
Now that I look more closely, nothing in our build process seems to actually use `drat-trim.c`. What's the workflow for using `drat-trim.c`? How should I go about testing it? Should I just compile `drat-trim.c` to an executable `drat-trim` and put it in my `$PATH` along with `glucose`, then run a full regression?

MattKaufmann:
Or maybe drat-trim doesn't belong in the ACL2 repository, and instead,
a README should point to Marijn's repository.  Does anyone mind this
change?  (Maybe someone at Oracle or Centaur uses it?)  I'm pretty
sure that the regression doesn't depend on it, since I've run
regressions recently but the directory books/tools/drat-trim/ only has
the .c file.

-- Matt

kini:
Yeah, I agree. We should avoid duplicating code if possible, and it seems there's no reason to have `drat-trim.c` in this repository specifically. I'll change this PR to delete `drat-trim.c` instead of updating it.

MattKaufmann:
I'd like to keep the directory and just put a README there pointing to:

https://www.cs.utexas.edu/~marijn/drat-trim/

Feel free to do that.  If you don't and nobody objects, then maybe I
will.

-- Matt

kini:
Done.

ragerdl:
Hi, I object.  As part of the repos, this is one less component that we have to download to get it to work.  I could make a test file for it, if we really want a regression.  But, it's not worth the trouble.  Can we please leave it?

ragerdl:
`drat-trim` seems a lot like bundling quicklisp to me.  That is, it needs to be updated every once in awhile.  One of the nice things about updating quicklisp is that there are effectively tests in the repo that obtain some amount of minimal coverage.  I'd be in favor of [someone not me] writing such a test for `drat-trim` and adding that test to the repo.  In the absence of such a test, I still find it more convenient to bundle `drat-trim` with acl2.

~my $0.02~

kini:
DRAT-trim is only useful in combination with a SAT solver, which one has to download separately anyway, so I don't really see what the issue is, in terms of the "ACL2 should provide full functionality without requiring external components" argument. We could also put a SAT solver into this repository, but we're not doing that.

Also, I don't think the question of whether we have tests for it is important. It's not like we'd be removing it because of a lack of tests, or keeping it because it did have tests. If the `drat-trim.c` file were actually used in the build process that might be a different story, but as it stands, there's no actual connection between this file and the other stuff in the repository. The only reason I can see why this file is in the repository is as a convenience for users so that when they download the repository, they also get this file, which they would otherwise have had to download separately.  (And I think that's exactly what you're saying you like about the current arrangement.)

However, I don't think that's a good reason to have a file in a repository. It might be a good reason to have a file in a distribution tarball, but not in a repository. Think about the manual for example. We could well bundle it with ACL2 tarballs, but we certainly wouldn't want to commit it into the repository.

MattKaufmann:
I think I agree with Keshav.  David raises an interesting analogy
between drat-trim and quicklisp.  I think the difference is that we
need quicklisp to build the manual, which is something that people
want to do, and some find critical.  Maybe a better analogy for
drat-trim is glucose, except that glucose is actually more tied into
the regression suite than drat-trim as far as I know, since I don't
think anything in the regression suite currently depends on drat-trim.
(Some of my books under books/projects/sat/lrat/ definitely tie into
the use of drat-trim, but not in the regression suite.)

I'm concerned about setting a precedent in the direction of
maintaining copies of external tools not necessary for the regression
suite, and deciding when to update.

But maybe I'm missing something.  In particular: David, if you use
drat-trim and find it onerous to fetch it from the web, that could
change my opinion.

-- Matt

jaredcdavis:
Hi,

I don't have any strong feelings about this.  FWIW, I think this could be
easily tested in a very basic way using, e.g.,

cd [books]/centaur/satlink/solvers cert.pl test-glucose-cert

As described in

http://www.cs.utexas.edu/users/moore/acl2/current/combined-manual/index.html?topic=SATLINK____UNSAT-CHECKING

I'm sure this isn't part of the build system today just because it would
require some Makefile mumbo jumbo for compiling drat-trim, putting it
somewhere usable, etc.  It would probably be fine to include that sort of
thing as part of satlink, but I agree that this seems a little odd if we're
not including, e.g., glucose or some other sat solver.

Cheers,
Jared

ragerdl:
At this point, I'm thinking that in an ideal world, `drat-trim` and `glucose` would both be submodules of our repository.

Given:

(1) Submodules add another level of complexity that some would probably object to (though, in these two cases, the submodules could be safely ignored by people who wanted to maintain and use their own copies of glucose/drat-time) and

(2) Glucose isn't currently on github (though, that's something I'm willing to fix)

Can we compromise and leave `drat-trim.c` in the repository?  If someone wants to update `drat-trim.c` to the latest, that sounds good to me.  Given we're not picking the "ideal" (in my view) solution, can we settle for something practical?

kini:
I chatted a bit with @ragerdl about this today. I really disagree that `drat-trim` and `glucose` should be submodules of our repository, partly because of the caveats that David stated, but mainly because they're not dependencies of ACL2. ACL2 and the community books are just designed to work with them if they happen to be present.

What's more, they're designed to work with them through publicly documented, very stable, simple APIs (the DIMACS SAT text file format for SAT problems and the DRAT text file format for UNSAT proofs). In terms of coupling strength ( https://en.wikipedia.org/wiki/Coupling_(computer_programming) ), this is a very weak coupling, making the argument for submodulization weaker as well. (A major reason for using submodules is to pin versions of your dependencies at specific commits to avoid the breakage that commonly occurs when tightly coupled dependencies are updated -- but in fact, submodules force you to pin versions, even if you don't want or need to.)

It's true that in comparison to the prospect of making `drat-trim` a submodule, I'd much rather just leave it in the repository, but that doesn't mean I would call it a "compromise", as David did. I protest this Overton window stretching :P

But anyway, this PR has been stagnant for many months so I'm closing it. I will, however, at least update `drat-trim.c` to the latest version, which was my original proposal and which David asked me to do when we chatted today.

PR #795, opened 2017-11-10
Xdoc: Add a hack to preprocessing to make save/render faster.

solswords:
This change speeds up the manual build (certification of doc/top.lisp) by around 3-5 minutes by adding a semi-ugly hack to xdoc's get-event, which it uses for expanding preprocessor directives like @(def ...) and @(thm ...).  The idea is that before we start, we traverse the world once and collect a fast alist binding every function/theorem/macro name to the event form we'd get from get-event*.  (Previously, each call of get-event would traverse the world linearly, which took a lot of time -- around 90 seconds in save-rendered with doc/top loaded.) This fast alist is generated and bound in a state global at the beginning of the save/render routine.  If it isn't bound, everything works as before, which is important for using :doc/:xdoc from the terminal.

Submitting this as a pull request in case anyone wants to weigh in -- @MattKaufmann @ragerdl @jaredcdavis

### DIFF
--- a/books/system/doc/render-doc-base.lisp
+++ b/books/system/doc/render-doc-base.lisp
@@ -236,8 +236,18 @@
     (mv (cons first rest) state)))
 
 (defun render-topics (x all-topics state)
-  (b* ((topics-fal (topics-fal all-topics))
-       ((mv ans state) (render-topics1 x all-topics topics-fal state)))
+  (b* ((prev-event-table-binding
+        ;; save the previous binding (or lack) of xdoc-get-event-table, set it
+        ;; to one generated from the current world, and restore at the end
+        (and (acl2::f-boundp-global 'xdoc-get-event-table state)
+             (list (f-get-global 'xdoc-get-event-table state))))
+       (state (f-put-global 'xdoc-get-event-table (make-get-event*-table (w state) nil) state))
+       (topics-fal (topics-fal all-topics))
+       ((mv ans state) (render-topics1 x all-topics topics-fal state))
+        (- (fast-alist-free (@ xdoc-get-event-table)))
+        (state (if prev-event-table-binding
+                   (f-put-global 'xdoc-get-event-table (car prev-event-table-binding) state)
+                 (makunbound-global 'xdoc-get-event-table state))))
     (fast-alist-free topics-fal)
     (mv ans state)))
 
--- a/books/xdoc/preprocess.lisp
+++ b/books/xdoc/preprocess.lisp
@@ -208,6 +208,46 @@
       (otherwise
        event))))
 
+
+;; (defun get-event-types (wrld acc)
+;;   (declare (xargs :mode :program))
+;;   (b* ((wrld (acl2::scan-to-event wrld))
+;;        ((when (atom wrld)) acc)
+;;        (form (acl2::access-event-tuple-form (cddr (car wrld))))
+;;        (kind (and (consp form) (car form))))
+;;     (get-event-types (cdr wrld)
+;;                      (if kind (cons kind acc) acc))))
+  
+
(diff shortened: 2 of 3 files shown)


MattKaufmann:
Looks like a cool idea to me!  I totally trust that you did it right (and besides, if not then we'll probably find out ;).  If nobody else comments soon, please feel free to accept your own pull request, as far as I'm concerned (or I will).

jaredcdavis:
Hi,

Sounds awesome.  Thanks!

Cheers,
Jared

On Fri, Nov 10, 2017 at 4:29 PM, MattKaufmann <notifications@github.com> wrote:
>
> Looks like a cool idea to me! I totally trust that you did it right (and besides, if not then we'll probably find out ;). If nobody else comments soon, please feel free to accept your own pull request, as far as I'm concerned (or I will).
>
> —
> You are receiving this because you were mentioned.




-- 
11410 Windermere Meadows
Austin, TX 78759
http://www.cs.utexas.edu/users/jared/

ragerdl:
Super-excited you found a reasonable way to speed-up our build bottleneck!  Merging, since everyone's chimed in.

MattKaufmann:
By the way, I ran an experiment, building the manual using
an SBCL-based ACL2 executable, both before and after
updating.  The difference was noticeable but small, as shown
below.  I can easily imagine that CCL would produce a bigger
difference because of static consing.

    ;;; old
    Built [...]/books/doc/top.cert (489.394s)
    443.548u 14.716s 8:24.49 90.8%  0+0k 2736+531360io 1pf+0w

    ;;; new
    Built [...]/books/doc/top.cert (473.852s)
    428.656u 14.960s 8:08.77 90.7%  0+0k 5960+531344io 0pf+0w

solswords:
Hi Matt,
Just curious - how long does CCL take to certify that book on your system?
-Sol

MattKaufmann:
Hi, Sol --

I'll have to do some runs to figure that out.  I'll do so and report
back when that's done.

-- Matt

MattKaufmann:
Hi, Sol --

Here are results with CCL.  The "git pull" I did updated from
something pretty recent, so we can reasonably attribute all of the
time difference to your change.

;;; old
Built /v/filer4b/v11q002/acl2space/acl2/acl2/books/doc/top.cert (979.578s)
921.244u 8.436s 16:56.12 91.4%	0+0k 1307424+531824io 219pf+0w

;;; new
Built /v/filer4b/v11q002/acl2space/acl2/acl2/books/doc/top.cert (841.411s)
801.364u 7.752s 14:15.73 94.5%	0+0k 7960+532016io 0pf+0w

Not bad!

A possible explanation for why there isn't more speed-up is that
get-event isn't terribly slow -- there is some world indexing that's
supposed to help it, I think.  I haven't thought this through or even
absorbed what you did, so please ignore this remark if it doesn't make
much sense.

-- Matt

solswords:
That's actually about in line with what I see with CCL now, too.  My
initial estimate of 3-5 minutes savings must have come from a slower
machine.  I wonder why SBCL is that much faster on this book -- is that in
the ballpark of a normal ratio for SBCL versus CCL?

MattKaufmann:
Hi, Sol --

I don't know about the ratio, but I think I've often seen SBCL run
faster than CCL.  Here are my most recent "regression-everything"
times:

CCL:
87413.460u 1125.316s 3:34:31.26 687.8%	0+0k 3262064+7011688io 1392pf+0w

SBCL:
74489.952u 1555.880s 3:03:19.36 691.3%	0+0k 3825432+7757496io 1344pf+0w

I vaguely recall mentioning that once to this list (or some
acl2-related list), and getting a reply that CCL is still a lot faster
for applications that make heavy use of HONS.

-- Matt

PR #798, opened 2017-11-13
Remove Y86 GPL2 file

ragerdl:
The file `books/models/y86/LICENSE` clearly states that the Y86 is
licensed under 3-clause BSD.  As such, I'd like to remove this reference
to the GPL.

@MattKaufmann , is this okay with you/UT?


MattKaufmann:
Good catch -- merged and closed.

PR #799, opened 2017-11-14
Make 1.x release notes use usual naming convention

kini:
Each ACL2 version 1.x's release notes are listed under the doc topic "NOTEx", rather than the more customary "NOTE-1-x". This pull request fixes that. I also changed the parents of the "books" topic, since it had "note1" (now renamed to "note-1-1") as a parent for some reason.

It will probably be hard to review this pull request with a standard diff viewer, in particular the second commit, feae5dd. I suggest individually finding the ten corresponding forms in the the 9613bc1 version and the feae5dd version of acl2-doc.lisp and comparing them -- that is, if you don't believe me when I say that I merely moved a contiguous block of lines up in the file :)


MattKaufmann:
Trivial to compare with Emacs meta-x compare-windows; looks fine.

kini:
Thanks!

PR #800, opened 2017-11-15
Replace a defunct link

airbornemihir:
Somehow, through dozens if not hundreds of C-t g keystrokes, I never noticed that one of the first hyperlinks is broken! This pull request tries to fix that.


ragerdl:
I actually noticed this yesterday too, thanks for the cleanup!  Rebasing and merging

PR #802, opened 2017-11-17
Provide hint for deffixequiv const-thm event

kini:
The fty::deffixequiv macro produces three theorems, namely fix-thm,
const-thm, and cong-thm.  Of these, const-thm is a syntaxp-conditioned
reversal of fix-thm; while fix-thm rewrites (F (FOO-FIX X)) to (F X)
under the equivalence relation out-equiv, const-thm rewrites (F X)
to (F (FOO-FIX X)) under out-equiv when X is a constant that doesn't
satisfy (FOO-P X).

The ideal way to prove const-thm is to apply fix-thm to
rewrite (F (FOO-FIX X)) to (F X) under out-equiv, but the surrounding
theory when the event is submitted may cause ACL2 to try to prove it
in a different way, possibly failing or taking far longer than it
should to prove such a trivial corollary of fix-thm.

This commit provides an :in-theory hint to the const-thm event so that
it is carried out in a theory that should allow it to always succeed,
and succeed instantly.

(I made this commit because I came across a situation where
fty::deffixequiv failed because FOO-FIX was enabled.)

### DIFF
--- a/books/centaur/fty/fixtype.lisp
+++ b/books/centaur/fty/fixtype.lisp
@@ -253,7 +253,6 @@
        (hints (getarg :hints nil kwd-alist))
        (skip-const-thm (or (getarg :skip-const-thm nil kwd-alist)
                            (not (fixtype->executablep fixtype))))
-       (out-equiv (getarg :out-equiv 'equal kwd-alist))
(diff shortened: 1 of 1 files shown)


solswords:
Sounds good, thanks!

PR #815, opened 2018-01-30
GCD self-timed circuit synthesizers

myall86:
Added the modeling and verification of GCD self-timed circuit synthesizers


ragerdl:
Nice, Cuong!

myall86:
Thanks, David.

PR #817, opened 2018-02-06
Improve some comments in cert.pl (fixes #749)

### LINKED ISSUE #749: Inconsistent comments in cert.pl about where the books are
In early section of the comments in cert.pl, we read:

```
# This script assumes that the ACL2 system books directory is where it
# itself is located.  Therefore, if you call this script from a
# different directory, it should still be able to resolve ":dir
# :system" include-books.  It also scans the relevant .acl2 files for
# each book for add-include-book-dir commands.
```

Then later at line 636, we read:

```
# If $acl2_books is still not set, then try the following in order
# until one succeeds:
#   - set it based on the location of acl2 in the path
#   - set it by running acl2 and dumping (@ system-books-dir)
#   - set it to the parent of the directory containing this script.
```

I haven't fully read the code but it seems to me that the second 
section of comments is more true than the first section. According
to my experience, $acl2_books will only gets set before the second
section of comments if the option --acl2-books is set or if the
environment variable $ACL2_SYSTEM_BOOKS is set. Otherwise,
it follows whats described in the second comment.

If that's really the case, should we remove the first comments and
move the second to the beginning with below changes?

```
# If $acl2_books is set in below order until one succeeds:
#   - set it based on cert.pl option --acl2-books
#   - set it based on environment variable $ACL2_SYSTEM_BOOKS
#   - set it based on the location of acl2 in the path
#   - set it by running acl2 and dumping (@ system-books-dir)
#   - set it to the parent of the directory containing this script.
```

kini:
See #749 for discussion. Sorry I took so long to do this.

### DIFF
--- a/books/build/cert.pl
+++ b/books/build/cert.pl
@@ -38,11 +38,26 @@
 # dependencies of all files needed.  When it is finished, it writes
 # out a file Makefile-tmp which can be used to make all the targets.
 
-# This script assumes that the ACL2 system books directory is where it
-# itself is located.  Therefore, if you call this script from a
-# different directory, it should still be able to resolve ":dir
-# :system" include-books.  It also scans the relevant .acl2 files for
-# each book for add-include-book-dir commands.
+# This script chooses which ACL2 to use by the following methods, in
+# decreasing order of priority:
+#   - use the ACL2 specified by the -a or --acl2 command line argument
+#   - use the ACL2 specified by the environment variable $ACL2
+#   - use whatever "acl2" exists in the $PATH
+
+# This script chooses which community books directory to use by the
+# following methods, in decreasing order of priority:
+#   - use the directory specified by the -b or --acl2-books command line argument
+#   - use the directory specified by the environment variable $ACL2_SYSTEM_BOOKS
+#   - look for a "books" directory where the ACL2 we're using is located
+#   - run ACL2 and check if (@ system-books-dir) evaluates to a directory that exists
+#   - use the parent of the directory containing this script
+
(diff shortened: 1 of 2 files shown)


pennyan:
Looks good to me! Think you for cleaning it up! I found one typo! "A third option is tell" -> "A third option is to tell" ?

kini:
Good catch! Fixed.

kini:
Thanks for the review! I made one more change -- I noticed that I had removed some text about .acl2 files, so I added that back in with some further clarification. To avoid more back-and-forth I'll just merge the PR now, but feel free to back out or edit the new paragraph I added (starting "This script also scans the [...]") if you don't like it.

The reason I put the "incidentally" in at the beginning of the last paragraph is that the previous paragraph, which says "At this point, `cert.pl` should be configured properly and ready to use.", is referring back to (and wrapping up) the entire doc topic, whose purpose is to help the user set up the environment in which they should be running `cert.pl`. In contrast, the last paragraph is about what happens when you don't follow the directions in the doc topic, and let `cert.pl` figure things out for itself; furthermore, it's only relevant to the books directory section and not the whole doc topic. Ideally I would have liked to put it before the "At this point [...]" sentence, but then it's weird because the "if `cert.pl` cannot determine [...]" paragraph is not an instruction to the reader, so one loses the sense of a "timeline of actions" while reading that paragraph, making "At this point" seem incongruous afterwards. So I figured the easiest thing would be to put that out-of-place paragraph at the end but with an "Incidentally," in front, to mark it as a postscript and make it seem like less of a non-sequitur. Oh well, maybe I'm overthinking it, lol.

PR #819, opened 2018-02-08
For SMT, make python call more generic

ragerdl:
Hi Yan,

Does this look okay to you?  Can you test such a change before I commit it?

Thanks,
David


pennyan:
Hi David, I've successfully tested this change on my machine when David asked me.

PR #826, opened 2018-02-23
Add ifdef feature to support conditional dependence on external tools…

### LINKED ISSUE #824: Make two manual build targets, one for z3+glucose and one for neither
Some are frustrated that there's no way to build the manual without `glucose`.  There's also discussion of making the current situation worse by also requiring `z3`.  This issue proposes that we create two manuals:

(1) One that requires neither `glucose` nor `z3`, and
(2) One that requires both `glucose` and `z3`

If you'd like (1), then feel free to record that here.  Also, if interested, please include whether you have the time and inkling to implement it.

Related to #789

solswords:
… etc.  Fixes #824 

Changes:
 - added a book that has ifdef and ifndef macros (see the added documentation in build/doc.lisp)
 - added support for ifdef/ifndef in certlib.pl dependency tracking
 - messed around in books/GNUmakefile a lot to make sure the relevant environment variables are visible to both ACL2 and the cert.pl invocation (surprisingly awkward).

I tested this with a couple different manual builds and I think it should be adequate.  Let me know if you have feedback.  Might be good to document somewhere how to add an optional external tool.

### DIFF
--- a/books/GNUmakefile
+++ b/books/GNUmakefile
@@ -135,12 +135,21 @@
 ### Section: Preliminaries
 ##############################
 
+# GNU Make exports exported variables to the shell inside recipes, but
+# not inside $(shell ...).  To work around this, we'll add every
+# variable we export to a list and manually provide the environment we
+# want when we need it.
+EXPORTED_VARS :=
+
+
 # For cygwin, we need to do something special to get a Unix-like
 # pathname.
 ifneq (,$(findstring CYGWIN, $(shell uname)))
 export ACL2_SYSTEM_BOOKS := $(shell cygpath -m $(CURDIR))
+EXPORTED_VARS += ACL2_SYSTEM_BOOKS
 else
 export ACL2_SYSTEM_BOOKS := $(CURDIR)
+EXPORTED_VARS += ACL2_SYSTEM_BOOKS
 endif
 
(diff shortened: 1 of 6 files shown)


books/GNUmakefile
```diff
@@ -135,12 +135,21 @@
...
+# variable we export to a list and manually provide the environment we
+# want when we need it.
+EXPORTED_VARS :=
+
+
 # For cygwin, we need to do something special to get a Unix-like
 # pathname.
 ifneq (,$(findstring CYGWIN, $(shell uname)))
 export ACL2_SYSTEM_BOOKS := $(shell cygpath -m $(CURDIR))
+EXPORTED_VARS += ACL2_SYSTEM_BOOKS
```
ragerdl: Want to move this outside the if-then-else?

ragerdl:
I'm assuming that Sol's changes are simpler and more easily maintained than the approach Matt was thinking of pursuing.  Thanks to both Matt and Sol for getting this particular issue resolved in a way that makes the largest number of people happy.  I haven't looked at the technical details, but the spirit of it sounds good to me.

BTW, is this sound?  E.g., does changing an environment variable after a book has been certified, and then including that book, provide a hole?

solswords:
Either way is fine.  I had each addition to EXPORTED_VARS right next to
where it's exported, so the correspondence is obvious.

solswords:
It's sound as long as make-event is sound :)

PR #829, opened 2018-03-21
Add a new model with a file allocation table matching the Microsoft specification

airbornemihir:
There's some work still to be done on this model - details can be found in the commit messages. However, it's going to be a while and the bulk of the work is done.


MattKaufmann:
What level of testing has been done with these changes?  For example, did you run "make all" in the books directory?

airbornemihir:
Hi Matt,
When I try to run "make all" in books/, I get a failure on account of the smtlink examples, as shown below. Can you suggest an alternative make command that would leave those out? Possibly "make basic", as suggested by :doc books-certification?

--
**CERTIFICATION FAILED** for /home/mihir/git/acl2/books/projects/smtlink/examples/examples.lisp

MattKaufmann:
Hi, Mihir --

No: unfortunately (or more accurately: fortunately), we want all the
books to certify, even with your changes.  One reason is that leeroy
will be running something like "make all", as I understand things; but
a deeper reason is that we simply want the books to certify.  We
should figure out why you are getting that failure.  Perhaps you'd be
willing to post a log of the certification attempt, which is probably
file:

    /home/mihir/git/acl2/books/projects/smtlink/examples/examples.cert.out

Perhaps the failure isn't related to your change, but we should
confirm that, and it would be good to figure out in that case what
needs to be done to fix books/projects/smtlink/examples/examples.lisp.

-- Matt

airbornemihir:
I should note that when I ran "make -k -j8 all" after my last commit, I got only the failure I mentioned above with the smtlink examples. "make -k" continues to build targets even after one or more targets fail, so I feel confident these changes work aside from the Z3 stuff which I'm not set up for.

I'll get in touch with you privately about the rest.

MattKaufmann:
I have contacted David Rager and Yan Peng, with a CC to Mihir, to try to get this resolved.

airbornemihir:
"make all" exits with status 0 in the books/ subdirectory as of the last commit (3ad67d394abeca46b6f104358b90fc59b2797a4c). Again, thanks everyone for helping me through the build issues.

PR #830, opened 2018-03-23
Remove duplicate function in set-theory.lisp

kini:
The function `packn-in-pkg` in this file duplicates the system function `packn-pos`, so this commit removes the function definition and replaces calls of it with calls of `packn-pos`.

One reason to reject this pull request might be that this book's didactic purpose makes it desirable to define `packn-in-pkg` in the book so readers can easily see what it does, whereas `packn-pos`'s definition is elsewhere, in the ACL2 sources.

Another reason to reject this pull request might be that because it is the companion file to J's paper "Finite Set Theory in ACL2", it shouldn't be changed if possible, lest it diverge from the exposition in the paper.

As an aside, there are many books in the repository which are companion files to publications, such as most of the books in the `workshops` directory. Some of these books haven't been changed for many years. Are we to consider them potential targets for improvement, or as static "historical" files? I know that they are sometimes changed when they break the build, but I don't think I've seen anyone making other non-trivial improvements to the workshop books.

### DIFF
--- a/books/finite-set-theory/set-theory.acl2
+++ b/books/finite-set-theory/set-theory.acl2
@@ -1,6 +1,6 @@
 (defpkg "S"
   (set-difference-equal
-   (union-eq '(PACK
+   (union-eq '(PACKN-POS
                ORDINARYP
                <<
                <<-IRREFLEXIVITY
--- a/books/finite-set-theory/set-theory.lisp
+++ b/books/finite-set-theory/set-theory.lisp
@@ -11,8 +11,8 @@
 ; The original kernel of this book was created with support from
 ; Compaq Systems Research Center, Palo Alto, CA.  A report documenting
 ; the main ideas in this book is available from the ACL2 home page.
-; Visit the link "Papers about ACL2 and its Applications" and then
-; visit "Finite Set Theory".
+; Visit the link "Publications about ACL2 and its Applications" and
+; then visit "Finite Set Theory".
 
 (in-package "S")
 
@@ -1020,11 +1020,6 @@
 ; including ``def'' in front of them and still search for ``(def...''
 ; when looking for things.
 
-(defun packn-in-pkg (lst pkg-witness)
-  (declare (xargs :mode :program))
-  (acl2::intern-in-package-of-symbol (coerce (acl2::packn1 lst) 'string)
-                                     pkg-witness))
-
 (defmacro defx (&rest args)
   (let ((temp (member :strategy args)))
 
@@ -1035,7 +1030,7 @@
     (cond
      (temp
       `(,(if (keywordp (cadr temp))
-             (packn-in-pkg (list (symbol-name (cadr temp))) 'defx)
+             (packn-pos (list (symbol-name (cadr temp))) 'defx)
            (cadr temp))
         ,@(butlast args (length temp))
         ,@(cddr temp)))
@@ -1056,7 +1051,7 @@
 
 (defun genname1 (name n avoid)
   (declare (xargs :mode :program))
-  (let ((new-name (packn-in-pkg (list name n) name)))
+  (let ((new-name (packn-pos (list name n) name)))
     (cond ((acl2::member-equal new-name avoid)
            (genname1 name (+ 1 n) avoid))
           (t new-name))))
@@ -1080,13 +1075,13 @@
            (vars (cdr expr))
            (x (nth (- n 1) vars))
            (name1
-            (packn-in-pkg (list fn "-SET-INSERT") fn))
+            (packn-pos (list fn "-SET-INSERT") fn))
            (name2
-            (packn-in-pkg (list fn "-CANONICALIZE") fn))
+            (packn-pos (list fn "-CANONICALIZE") fn))
            (vars1 (acl2::remove-eq x vars)) ; Matt K. mod from delete1, v2-9-4
            (e (genname 'e vars1))
            (a (genname x vars1))
-           (x-equiv (packn-in-pkg (list x "-EQUIV") x)))
+           (x-equiv (packn-pos (list x "-EQUIV") x)))
       `(encapsulate
         nil
         (local (in-theory (enable scons scar scdr ur-elementp)))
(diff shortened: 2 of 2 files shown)


ragerdl:
Personally, since we're on a version control system and will likely be on git for a long time, I think it could be reasonable for us to expect users to time-warp to previous versions of files via `git`.  Also, Matt still puts up tarballs with every release.  And, it's easy to determine the correct tarball by looking at workshop dates and release dates.

Thus, it seems, to me, okay to improve workshop files.  But, I don't have a strong opinion and can certainly believe that others might feel strongly the other way.

I'll email J to see if he blesses these changes :)

airbornemihir:
I was reflecting on an adjacent topic with my own recent pull request. Suppose I actually want to include a book from workshops, locally or otherwise. Suppose further that the book has a screwy definition of len-of-nthcdr, a reasonably useful defthm from std/lists/nthcdr.lisp, that works for the book's own purposes but does not fit ACL2's precise definition of a "redundant definition." Would I be justified in changing the workshops book, so I could include it and include the nthcdr.lisp book as well in the same development?

ragerdl:
Hi Mihir,

Good question!  In this case I think the answer is clear: make a copy of the workshop book somewhere in your `projects` space and then modify to suit your current project :).  

~David~

kini:
Thanks for the comments, David and Mihir.

David, I disagree that the answer to Mihir's question is clear; I think it's quite tied up in my original question (which, of course, I also think doesn't have a clear answer, which is why I asked it :P). Actually I also disagree with your answer itself -- I think that if the books can be brought more in line with each other and integrated nicely, we'll have a stronger and more cohesive base library to build further ACL2 code on, which is a good thing. If, when we want to use two books that are incompatible with each other, our default action is to split off one of the books and create a variant copy, we're just encouraging needless divergence of our toolset.

On the other hand, some workshop books are designed to be used as components of other projects, and some workshop books are intended to be stand-alone projects themselves. I imagine that when facing a situation like Mihir's, the decision of what to do will depend on the details of the situation.

By the way, I found the following paragraph on the [ACL2 Workshops and UT ACL2 Seminar page](http://www.cs.utexas.edu/users/moore/acl2/workshops.html) on the ACL2 website:

>ACL2 input files (certifiable books) from the preceding workshops are available from the links above. **WARNING**: The above links point to the original versions of those books. In order to obtain up-to-date versions of those books that will certify in the [latest version of ACL2](http://www.cs.utexas.edu/users/moore/acl2/), you should [download workshops.tar.gz](http://code.google.com/p/acl2-books/downloads) to the `acl2-sources/books/` subdirectory of your ACL2 distribution, and then gunzip and extract it. On a Unix/Linux system you can then certify all the books by standing in the `acl2-sources/` directory and issuing the command `make regression`.

This paragraph is probably a little out of date, since typically people download ACL2 and the community books together in one tarball or one git repository clone, and get the workshops books along with that. Also, it links to Google Code, which doesn't really exist anymore.

Still, though, it seems to indicate that one officially should expect that the "current" workshop books might have changed since the articles were published; yet it also seems to imply that the current books differ from the old mainly in that they are "up-to-date" and "will certify in the latest version of ACL2", so that doesn't seem to explicitly bless other kinds of changes.

MattKaufmann:
I'll weigh in on the recent discussions about changing books,
including J's finite-set-theory book and also (as Mihir discussed)
changing workshops/ books.  I discussed all this with J today, and I
think he and I are on the same page, but I'm CCing J here (since he is
not on acl2-books) in case he wants to add (or subtract) anything.

It seems fine to change any non-workshop book to make it more useful.
That applies to the particular change from Keshav, which replaced J's
custom function packn-in-pkg with a built-in ACL2 function, packn-pos.
(J is fine with that.)  However, one could quibble with that
particular change, because there is, in principle, risk associated
with using a built-in ACL2 function (like packn-pos) that does not
have a :doc topic: J and I feel free to change such unadvertised
functions, because we introduced them for our system, and we do not
want to be inhibited when we choose to change them.  A nice resolution
in this case would be for Keshav to contribute documentation for
packn-pos, or to ask me to do so.  By coincidence (except it probably
isn't a coincidence), he and I are having a related discussion
already; Keshav, let's discuss adding such documentation.

Now, turning to workshops/ books....

These are not generally used elsewhere, and they form archival support
for corresponding ACL2 Workshop papers.  I plan to fix the misleading
comment found on
http://www.cs.utexas.edu/users/moore/acl2/workshops.html (thanks,
Keshav!).  In particular, most recent workshops do not include links
to supporting materials, which weakens Keshav's argument (my fault for
the misleading comment on the workshops page, not his): there isn't
really a static version of a given set of supporting materials.
(Well, I suppose the first appearance in a release is such a version
-- in fact I see that David has pointed this out -- but that's not
always particularly easy to access.)

So my opinion (and J's) is that books under workshops/ should not be
changed, even to improve them, unless either it is necessary because
something has changed (either the ACL2 system or supporting books) or
the original author of the book approves of the change (or can't
easily be contacted).  If a workshops/ contribution is generally
useful for other things (and maybe one has been identified by Mihir),
then perhaps it's worth considering to move it elsewhere (e.g.,
somewhere under projects/ or tools/ or std/), with permission of the
author, and with a README file in the workshops/ directory stating the
new location.

Of course, J and I are just two members of the community.  Speaking
for myself: the sense of the community overall is more important than
what J and I think.  (But of course, we're right... ;)

-- Matt

MattKaufmann:
Hi everybody.  I agree with what Matt said about
workshop books except for one thing (which is probably
a mis-statement by Matt!): he said that if a workshop
book is thought to be generally useful it should
perhaps be MOVED elsewhere and I would have said COPIED
elsewhere.

J

MattKaufmann:
Hi, J and acl2-books --

[Adding acl2-books to the CC, because I'm not sure that my original
"To" address actually worked:]

Actually I did mean "move", but I think I see the advantage of "copy":
the old version is still under workshops/ to serve as archival
material, but the new version can evolve.  Yeah, that seems good to
me.

-- Matt

MattKaufmann:
Thanks, @kini, for adding CONTRIBUTING.md to workshops/.

PR #833, opened 2018-04-05
l3-to-acl2 translator tweaks

kini:
### Remove ASSERT! macro from l3-to-acl2 translator

@nadezhin discovered that the definition of ASSERT! in the l3-to-acl2 translator conflicted with an identically named macro in the misc/assert book (which is also documented at :DOC assert! in the ACL2+Books manual).

This commit solves that problem by doing away with the ASSERT! macro in the l3-to-acl2 translator.  It was only being used in the pattern (ASSERT! NIL \<something\>), which I was able to replace everywhere with the pretty much identical expression (PROG2$ (IMPOSSIBLE) \<something\>).

### Expand ATOM in l3-to-acl2 translator's EQUAL-LEN-0

There are more than two dozen different versions of this theorem throughout the community books.  Most of them rewrite (EQUAL (LEN X) 0) to (NOT (CONSP X)), some rewrite it to (ATOM X), and some rewrite it to (ENDP X).  I'm not going to try to unify them all right now, but @nadezhin wanted the one in the l3-to-acl2 translator to at least be compatible with the one in the projects/apply/apply-lemmas book, which happens to be the kind which rewrites to (NOT (CONSP X)).  Hence this commit.

### Use :dir :system paths in l3-to-acl2 translator

As @nadezhin pointed out, the reason absolute paths were used in the l3-to-acl2 translator is probably that originally the translator was not part of the ACL2 community books.  Now that it is, we can do away with the absolute paths, which this commit does.


books/projects/translators/l3-to-acl2/examples/mips/mips.lisp
```diff
@@ -8,7 +8,8 @@
                                                ("DATA" . IS_DATA)
                                                ("gpr" . GPREGS)))))
 
-(INCLUDE-BOOK "/Users/kaufmann/projects/l3-to-acl2/translator/l3")
+(INCLUDE-BOOK
```
nadezhin: It can be replaced by relative path or by `:dir :system`

nadezhin: Maybe it is possible to change definition of `*tranlator-dir*` in line 1281 of `l3-to-acl2.lisp` by `projects/translators/l3-to-acl2/translator/` and add `:dir :system` in line 1293 ?

nadezhin: I think that absolute path remains from time when l3-to-acl2 was not in the repository.
Now it is in the repository and `:dir :system` can be used`. Let's ask Matt.

kini: OK, I think you're right. I'll make a commit which does that.

MattKaufmann:
I'll accept the pull request.  The changes look good to me; I confess
that I haven't looked at this stuff in years, but I don't think
anything else depends on it, and if everything certifies, that's
probably sufficient.  That said, normally I annotate changes with
comments like "Matt K. mod".  You could consider adding one or two
such comments, or maybe just a sentence in the README, though the
changes seem minor so it's probably fine if you don't bother.
By the way, I don't know why I didn't use :dir :system, but the use of
it seems to me to be an improvement.

PR #834, opened 2018-04-07
Extend `impossible` to take an argument

kini:
### Extend `impossible` to take an argument

I renamed the nullary utility function `impossible` in the book std/basic/defs to `impossible-fn`, and created a macro named `impossible` which optionally accepts one argument, `retval`, which, when provided, determines the logical value of `(impossible retval)`.

I also documented this in `:doc impossible`, and improved the hard error message that is shown when running `(impossible-fn)` from program-mode code or with guard checking set to `nil` or `:none`.

I also added a defpointer in the system docs from `:doc guard-checking` to `:doc set-guard-checking`.

### Move NOTE defpointers to end of acl2-doc.lisp

In commit 6aea004330e463b48616d46143c185c2f944cc1e I renamed the ACL2 1.x release note doc topics so that they use the usual naming convention for release note doc topics.  For backwards compatibility, I created defpointers from the old names of those topics to the new. Now I realize that there apparently was a convention that all defpointers in the file are in a big alphabetized list at the end, so this commit moves those defpointers I created to where they should be.

### Fix typos


MattKaufmann:
Oops, please ignore my preceding comment.  I see now that the release notes were already renamed in a previous commit, which I probably didn't notice, and is probably fine.  I'll just accept this pull request (the "impossible" changes look OK to me on a quick scan).

MattKaufmann:
Sigh, testing failed because a dependency is missing.  I think the solution is to add the following line in the multi-line comment in books/projects/translators/l3-to-acl2-deps.lisp:

    (include-book "std/basic/defs" :dir :system)

My git ignorance prevents me from doing this with confidence.  If someone tells me how, or better yet just fixes this, then we can un-stick testing on leeroy.

My plan is that once this all gets into master, I'll check whether any other dependencies are missing and I'll consider updating a README file about maintaining dependencies.

ragerdl:
I'll make the change directly, and we can see what happens.  Thanks,
@mattkaufmann.

kini:
First, thanks to both of you for looking into this while I slept in on
a Saturday morning ;)

I did run a full regression-everything on this PR yesterday before
submitting it, so I don't think there was any real dependency problem;
I guess that, as you discovered, it was an issue with the custom
Makefile stuff in the l3-to-acl2 directory, which didn't clean out all
the l3-to-acl2 books when l3-to-acl2-deps was rebuilt.  That is, I
think Matt's solution below is the best fix:

On 04/07/2018 07:52 AM, Matt Kaufmann wrote:
> I *think* I've figured out a change that will prevent this problem
> from happening in the future.  I plan to make this change when the
> dust settles.  Read on only if interested....
>
> .....
>
> In books/GNUmakefile, I am tempted to change these lines
>
> projects/translators/l3-to-acl2/target.cert: \
>   projects/translators/l3-to-acl2-deps.cert
> 	cd $(@D) ; $(STARTJOB) -c "$(MAKE) -j 1"
>
> to:
>
> projects/translators/l3-to-acl2/target.cert: \
>   projects/translators/l3-to-acl2-deps.cert
> 	cd $(@D) ; $(STARTJOB) -c "$(MAKE) -j 1 clean"
> 	cd $(@D) ; $(STARTJOB) -c "$(MAKE) -j 1"
>
> It will continue to be the case that books/GNUmakefile only remakes
> this stuff when projects/translators/l3-to-acl2-deps.cert is out of
> date.  The difference is that when this happens, all f l3-to-acl2/
> will be remade.

Matt, when you make this change, could you please also remove the
(include-book "std/basic/defs" :dir :system) event from
l3-to-acl2-deps?  The std/basic/defs book is already included in that
file via ihs/logops-lemmas, so it's a redundant event.

> "David L. Rager" <ragerdl@gmail.com> writes:
>> Matt, I don't think you want to know how I made the fix.  But, I'll
>> state it here, so Keshav knows what I did (since he's already more
>> familiar with the github interface).  @Keshav, I edited the file in
>> the deps file in the testing branch directly in the github interface.
>> I committed it directly, because the branch is already broken.

Thanks for the explanation.  By the way, I try to avoid using the
github interface directly nowadays because it creates subtly malformed
commit objects in the git store, as the maintainer of magit explained
to me in 2016:

  https://github.com/magit/magit/pull/2822

All git tools I've used (including the `git` command line tool and
magit) are sufficiently tolerant that they can handle this
malformation, but still, it doesn't seem like a great thing to rely
on.

Linus Torvalds also ranted about on-the-web, github-made commits back
in 2012, lol (not that that's necessarily important, but it makes for
good reading anyway):

  https://github.com/torvalds/linux/pull/17#issuecomment-5659933

>> BTW, I suspect that this error might not show up if we cleaned the
>> workspace,  It's non-deterministic which one of std/basic/defs  and
>> the l3 books gets built first.  I'm kind of surprised that this has
>> ever worked reliably.
>>
>> Follow-up: Matt and I just discussed.  We got around this error by
>> having me "ssh" into the machine and doing a make clean in the
>> l3-to-acl2 directory.  I've restarted the build, and we suspect it'll
>> go through.  We also suspect that cleaning the workspace and
>> re-running the build would have worked.

Yes, I agree that cleaning the workspace would have made this build
failure go away, since I did run a full regression on the PR with no
problems.

>> I'm hopefully off to do a bunch of yard work today, so I'll probably
>> be afk.  But, it may be too cold... we shall see!

Have fun!

Thanks,
    Keshav

P.S. I'm replying to acl2-testing but also CCing the github ticket, in
case anyone's reading there.  I'm doing this by copying and pasting
the (apparently uniquely generated) Reply-To address from the
particular notification email I got from github when the latest
comment on PR #834 was made.  Plus, I'm sending this email from a
different address than the one that notification was sent to.  Let's
find out what chaos ensues! :P

PR #836, opened 2018-04-11
add xdoc feature to support books that contain some documentation wit…

solswords:
…hout the events it's based on

Notes: Pull requesting instead of just pushing because it changes the internals of the xdoc preprocessor somewhat & I thought might warrant some discussion -- see the changes to xdoc/preprocessor.lisp for details.  Mainly I restructured the code a bit and added an escaping mechanism for @({ }) and @(' ') constructs.

The feature supports isolating the xdoc produced by some local events (up to and including include-books), partially preprocessing them so that they don't refer to definitions that won't exist anymore after exiting the encapsulate.  (That assumes that the definitions do exist at the time the xdoc is partially preprocessed, so some care needs to be taken to choose the set of definitions to use this on.)

In principle, this could be used to create a book containing just the VL documentation (without VL), just the Kestrel documentation (without the Kestrel books), etc, etc, which could make it faster to build the manual.  In practice this is a little complicated due to the interconnectedness of a lot of our libraries and I haven't tried to actually do any of it yet. (I did check that the feature works, though.)


MattKaufmann:
This looks great -- what a fine idea!

I like the idea of using it in books/xdoc/top.lisp, though
that will remove the forced compatibility checking we have
now, where different definitions of the same function symbol
in two books included in xdoc/top would now cause an error.

Perhaps we'll want to arrange that "make all" certifies a
version of top.lisp that uses archive-xdoc, while "make
everything" certifies a version of top.lisp that does not
use the new feature.  That way, "make everything" will
continue to enforce the compatibility that is currently
enforced.

I wonder if the version of top.lisp that uses archive-xdoc
will wrap it around each include-book separately, to keep
the memory footprint manageable.  If so, then it would be
good to avoid redundant include-books, perhaps only
including "maximal" ones.  (Perhaps some code in the
implementation of include-book-paths could help with that,
though I haven't thought about it.)

I'd accept the pull request now, but since Sol has suggested
that this change might warrant discussion, I'll leave that
to someone else -- perhaps Sol, when he feels enough time
has passed.

solswords:
I agree that it would still be good to check for name clashes, maybe with a separate top book that tries to include as much as possible.

I think the fastest way to build a manual (assuming some degree of parallelism) would be for several libraries (maybe just the largest ones, or maybe more) to have individual top-doc books where the library is locally included and the xdoc extracted.  Then to include the xdoc for (say) VL, we just include that book, which doesn't need to include VL.  That way we both save on memory usage of doc/top.lisp (because it doesn't have to include all the big libraries simultaneously) and also are able to do the most expensive parts of building the documentation (mainly just including the books) in parallel.  We might need more tools to do this effectively, too -- e.g. it might be better to have a tool that filters the documentation that was created within a given directory and saves only that, or something.

I think I'll merge this after a few more hours unless anyone has more feedback/objections.

PR #839, opened 2018-04-17
Rename theorems for consistency

airbornemihir:
This is a minor change to William Bevier's book, but it does make the theorems a little better by removing an unnecessary call of nfix. Normally I would want to push this along with a more substantial update to the books, but I'm going to have to roll back to an old version in order to help fix a Makefile and it's just easier to get this out of the way.

If you're merging this request, let me know if you'd prefer me to bundle changes in a larger pull request next time.


books/data-structures/list-defthms.lisp
```diff
@@ -929,18 +929,19 @@
...
-           (nth (- (nfix n) (len a)) b)))
-  :hints (("Goal" :induct (revappend a b))))
+;; Renamed for consistency with std/lists/nth.lisp
+(defthm nth-of-append
+  (equal (nth n (append x y))
+         (if (< (nfix n) (len x))
+             (nth n x)
+           (nth (- n (len x)) y))))
+
+;; Renamed for consistency with books/std/lists/nth.lisp
```
kini: I suggest changing this to read "std/lists/nth.lisp", or changing the comment on the previous theorem to read "books/std/lists/nth.lisp". Actually I suggest putting these comments in your commit message body rather than in the file -- a future reader won't care that the names of the theorems used to be different (and your comment doesn't tell them what the names used to be anyway).

ragerdl:
w.r.t. "let me know if you'd prefer me to bundle changes in a larger pull request next time."  My (strong) vote is to make change lists (pull requests) as small, independent, and modular as possible.  This way we can focus on as few issues as reasonably possible within a single pull request.

Also, in the case that a change needs to be rolled back, it's much easier to keep the parts that aren't problematic because they're more likely to be part of a different set of commits.

airbornemihir:
@solswords I looked at your examples of disabling nth-of-append and defining your own version, and no doubt if you had to do that then it isn't as clear-cut as it seemed initially. Yet, I'm puzzled by a slight contradiction in what you said - you said the nfix was unnecessary, but you also said it allowed rewriting n with any term that's equivalent to it w.r.t. nat-equiv. Did you mean the latter, rather than the former?

Also, if you did mean the latter, couldn't we add a congruence rule allowing the first argument of nth to be re-written under nat-equiv?

solswords:
@airbornemihir By "the nfix was unnecessary" I just meant it doesn't need to be there (in the RHS of the rewrite rule, in the else branch of the if) in order for the theorem to be true.  Whether it's desirable or not is another question -- I think it's good to have it there, because it keeps N in a nat-equiv context everywhere.

We do indeed have the congruence rule you mention (in std/basic/arith-equivs, for example).  The problem with the RHS of that rule is that while (- n (len a)) as a whole can be rewritten in a nat-equiv context (when that congruence rule is loaded), there's no such congruence rule about binary-+, so we don't get to rewrite n under that nat-equiv context.  In fact, the following is a theorem, but can't be a congruence rule because ACL2 doesn't allow congruence rules to have hypotheses other than the one equivalence --
```
(defthm nat-equiv-congruence-of-plus-non-positive
    (implies (and (integerp b) (<= b 0)
                  (nat-equiv a1 a2))
             (nat-equiv (+ a1 b) (+ a2 b))))
```

airbornemihir:
This is interesting. I tried proving this:
`(defthm nat-equiv-implies-nat-equiv-+-1
      (implies (nat-equiv y y-equiv)
               (nat-equiv (+ y (ifix x)) (+ y-equiv (ifix x)))) 
      :rule-classes (:congruence))`
and it failed, with a subgoal suggesting this might be useful:
`(thm (implies (integerp x) (equal (integerp (+ x y)) (integerp y))))`
and this failed as well. I'm actually surprised this isn't a built-in theorem in ACL2, but either way, this pull request may need to be shelved.

kini:
That subgoal is false - consider x = 0 and y = nil. The original theorem is
also false - consider x = (complex 1 0), y = (complex 0 1), y-equiv = nil.

2018年4月19日(木) 9:40 airbornemihir <notifications@github.com>:

airbornemihir:
All right then.

PR #842, opened 2018-05-03
Add a lemma

airbornemihir:
This is handy while reasoning about unsigned-byte arrays within stobjs.

### DIFF
--- a/books/std/typed-lists/unsigned-byte-listp.lisp
+++ b/books/std/typed-lists/unsigned-byte-listp.lisp
@@ -114,7 +114,7 @@
     (implies (and (unsigned-byte-listp width (take n x))
                   (unsigned-byte-listp width (nthcdr n x)))
              (unsigned-byte-listp width x)))
-  
+
   (defthm unsigned-byte-listp-of-update-nth
     (implies (and (unsigned-byte-listp n l)
                   (< key (len l)))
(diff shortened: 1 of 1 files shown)


ragerdl:
@kini and @acoglio -- y'all are more qualified than I to review this

airbornemihir:
@acoglio I noticed you have a separate doc topic
THEOREMS-ABOUT-UNSIGNED-BYTE-LISTP for some lemmas you keep in a different
file. Would this be more appropriate to add there?

kini:
Looks good to me. I do wonder why this book even exists, though, given that `std::deflist` exists. I see a comment in the xdoc string for the `unsigned-byte-listp` section: "BOZO consider switching this book to use deflist.". Maybe we should do that at some point; it doesn't look like this book has been touched since 2014. Once that's done, the theorem you're adding in this pull request can probably be replaced by the theorem that `std::deflist` will produce from the following abstract theorem in `:/books/std/lists/update-nth.lisp`:

```lisp
(def-listp-rule element-list-p-of-update-nth
  (implies (element-list-p (double-rewrite x))
           (iff (element-list-p (update-nth n y x))
                (and (element-p y)
                     (or (<= (nfix n) (len x))
                         (element-p nil)))))
  :hints(("Goal" :in-theory (disable update-nth-when-atom)
          :induct (update-nth n y x))))
```

However, I'm not suggesting that you (or anyone) do this right now. For now I like the theorem you're adding in this PR.

airbornemihir:
@kini I noticed that comment too, but I'm not sure how that could be 
done with std::deflist. I mean... (unsigned-bytep n) is a function only 
if we think in currying terms, which ACL2 does not support (although I'm 
not sure if this is still true after the introduction of apply$). 
Anyway, so, std::deflist needs a function name provided to it as a 
parameter, and (unsigned-bytep n) isn't a function name.

acoglio:
@airbornemihir I'll take care of moving that theorem, but thanks for offering.

@kini I agree that at some point it would be good to use std::deflist but, as you say, it's not urgent.

PR #848, opened 2018-05-22
Make an inclusion local

airbornemihir:
The non-local inclusion of "std/lists/take" in std/typed-lists/unsigned-byte-listp.lisp was causing a number of very permissive lemmas from the former to be included in every book that included the latter. This meant the output of (show-accumulated-persistence :useless) was full of these lemmas in the books I'm working on, and probably in others as well. The proposed change could potentially have been very broad, but I've tried to choose the smallest set of books that allow the regression go through.

(There are two commits in this request, one of which reverts the other's changes.)


kini:
Seems reasonable to me, but:

>(There are two commits in this request, one of which reverts the other's changes.)

Why? Also, this doesn't seem to be accurate as far as I can tell (the second commit doesn't revert the first commit's changes).

airbornemihir:
So, if you look at the commit diffs in isolation, it'll seem like the
changes are disjoint because the second is a merge commit (cleaning up some
stuff for pull request#846 which was superseded by Matt's changes) and
GitHub shows its diff with its first parent. But you could locally checkout
the pull request (like we discussed in the office today) and diff it with
its second parent to see the reversion - or else, if you want to stay on
the GitHub website, you could look at the diff for the whole pull request.

kini:
OK, I see what's happened. The merge commit e7f2ed6342808c3dfac065af6cbf1e3fddc13648 is what is known as an "evil merge" (see `git help glossary`), and the other commit in this PR, 8b3bceb9cf21a83c619cd115c4bade7ad39cf098, is superfluous because the evil merge right after it also undid its effects. Please create a single commit containing your changes, descended from the testing branch, and force push it to this PR. You'll have to use your own fork's testing branch to do this, since this PR is specified as a PR from airbornemihir:testing to acl2:testing; in the future it might be a good idea to use named branches for pull requests you make, to avoid issues like this.

PR #849, opened 2018-05-26
Fix crash in EXPLAIN-FN

kini:
Previously, EXPLAIN-FN assumed that the only way that a rule could
have failed to apply was that one of its hypotheses failed to be
relieved.  Perhaps that's because EXPLAIN-FN was originally only meant
to deal with rewrite rules, before I changed WHY and WHY-EXPLAIN to
handle other rule types in #754.  (However, even rewrite rules can
sometimes fail for other reasons, such as because of backchain
limits.)

This commit makes EXPLAIN-FN a bit more robust; now it checks if the
failure reason is of the correct form before trying to extract a
term (the term the unrelieved hypothesis rewrote to) from it.  This
also means it's now unnecessary to run :OK-IF (BRR@ :WONP) before
calling EXPLAIN-FN.

### DIFF
--- a/books/std/std-customization.lsp
+++ b/books/std/std-customization.lsp
@@ -84,25 +84,50 @@
     (defun explain-fn (state)
       (declare (xargs :stobjs (state)
                       :mode :program))
-      (mv-let (clause ttree)
-        (clausify-type-alist (get-brr-local 'type-alist state)
-                             (list (cddr (get-brr-local 'failure-reason state)))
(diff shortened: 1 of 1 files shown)


ragerdl:
Always happy to see improvements to the explain utility.  I haven't examined your changes in detail, but I know you use the utility in the same way I do, so I trust that you've been able to test it well enough to warrant the merge.  Thanks!

PR #850, opened 2018-05-26
Make an inclusion local

airbornemihir:
This pull request supersedes  #848, the description is reproduced below, sans a comment which no longer applies.
--
The non-local inclusion of "std/lists/take" in std/typed-lists/unsigned-byte-listp.lisp was causing a number of very permissive lemmas from the former to be included in every book that included the latter. This meant the output of (show-accumulated-persistence :useless) was full of these lemmas in the books I'm working on, and probably in others as well. The proposed change could potentially have been very broad, but I've tried to choose the smallest set of books that allow the regression go through.

### DIFF
--- a/books/unicode/utf8-decode.lisp
+++ b/books/unicode/utf8-decode.lisp
@@ -43,6 +43,7 @@
 (local (include-book "tools/mv-nth" :dir :system))
 (local (include-book "std/lists/append" :dir :system))
 (local (include-book "std/lists/revappend" :dir :system))
+(include-book "std/lists/take" :dir :system)
 (set-verify-guards-eagerness 2)
 (set-state-ok t)
 
--- a/books/projects/x86isa/tools/execution/exec-loaders/mach-o/mach-o-reader.lisp
+++ b/books/projects/x86isa/tools/execution/exec-loaders/mach-o/mach-o-reader.lisp
@@ -11,6 +11,7 @@
 (include-book "std/io/read-file-bytes" :dir :system)
 (include-book "mach-o-stobj"
               :ttags (:include-raw :syscall-exec :other-non-det :undef-flg))
+(local (include-book "std/lists/take" :dir :system))
 
 ;; ======================================================================
 
--- a/books/std/io/take-bytes.lisp
+++ b/books/std/io/take-bytes.lisp
@@ -37,6 +37,7 @@
 (local (include-book "base"))
 (local (include-book "arithmetic/top" :dir :system))
 (local (include-book "tools/mv-nth" :dir :system))
+(local (include-book "std/lists/take" :dir :system))
 (set-state-ok t)
 
 (defsection take-bytes
--- a/books/std/typed-lists/unsigned-byte-listp.lisp
+++ b/books/std/typed-lists/unsigned-byte-listp.lisp
@@ -34,7 +34,7 @@
 (in-package "ACL2")
 (set-verify-guards-eagerness 2)
 (include-book "std/lists/repeat" :dir :system)
-(include-book "std/lists/take" :dir :system)
+(local (include-book "std/lists/take" :dir :system))
 (include-book "arithmetic/nat-listp" :dir :system)
 
 


books/unicode/utf8-decode.lisp
```diff
@@ -43,6 +43,7 @@
 (local (include-book "tools/mv-nth" :dir :system))
 (local (include-book "std/lists/append" :dir :system))
 (local (include-book "std/lists/revappend" :dir :system))
+(include-book "std/lists/take" :dir :system)
```
ragerdl: @airbornemihir How come you added a non-local include here?

ragerdl:
Seems like a good idea in general.  Since some people will probably have books that implicitly used `std/lists/take`, it'd probably be good to add a comment to the books :doc topic.  Also, have a question about the non-local inclusion of std/lists/take in one of the books.  Thanks!

airbornemihir:
@ragerdl Sorry about the late reply - I responded through email a few days ago, but Gmail ate the comments. One of these new commits addresses your concerns about the non-local include I added earlier, and the other adds some documentation - I tested these changes locally and on a UTCS computer.

airbornemihir:
Also, I'm happy to discuss the proposed documentation updates before they are merged.

ragerdl:
Looks good -- could you please add a sentence to note-8-1-books saying that you changed the book to a local include?  It can be found in https://github.com/acl2/acl2/blob/master/books/doc/relnotes.lisp.  Thanks!

kini:
Please don't add nontrivial extra changes in merge commits that didn't already exist in either of the parents, as you have done in commits e3fea747172912f62a390b87e2ac561cc09b79e1 and f116ffdb223b4fcaf35c89db78f9a8c61133c720
. See "evil merge" under `git help glossary`. Also, in ea1b6e23edff8d116c763adcc8f1062dff6eb864, you made a simple merge without making any further changes, but you changed the commit message so that it seemed like there might be some changes involved. Please don't do that either.

Ideally a merge commit should just have the default commit message, and that default commit message should be accurate, i.e. it should be "just a merge". This makes it much easier to review changes and to see what has happened in the git history. Also, please consider using `git rebase` (or `git pull --rebase` as the case may be) in order to avoid needless merge commits when you can.

kini:
I think it's fine to wait for leeroy and see if it fails -- that's what it's there for :) You changed several books which have a lot of reverse dependencies, so I'm not surprised the rebuild time is long. This time it seems like leeroy fell over or something, though, since the build you linked to doesn't seem to exist (anymore?). I guess it will get up again eventually.

kini:
Yup, it's trying again now: http://leeroy.defthm.com/job/acl2-testing/3043/console

Actually, now I realize that since you changed some documentation in the `defrule` book in 606ab0bf608a836844b40bd1229b26d4ea49f355, it's even less surprising that a large portion of the community books will need to be rebuilt. It would be nice if documentation changes in a book didn't require a rebuild of the book, but as far as I can tell (and based on a conversation with Matt I had a while ago) it would be pretty hard to arrange xdoc in such a way that that could be the case, unfortunately.

MattKaufmann:
I guess I don't know why it's all that important to include
the :doc with define and other macros.  I realize that this
has the "literate programming" advantage of keeping the
documentation in sync with the function, but I wonder if
that isn't outweighed by the current issue, namely, slowing
down builds.  In principle, at least, I'd expect it to be
possible to gather up formals with types and do other fancy
stuff that DEFINE and other macros might do, by including
the original book in the corresponding documentation book.
Of course, even if this is feasible, there's the issue that
someone would need to step up and make such changes.

airbornemihir:
@kini I almost forgot to write back about this. I started writing a defence
of evil merges, but on reflection I think you're right - GitHub has the
ability to separate and show only the part of the diff that makes a merge
evil, but it's not easy to see on the command line, nor do tools for
sending email with commit diffs do sane things with evil merges in general.
I think it's sensible to make separate merge commits, when merge conflicts
need to be resolved (and GitHub's pull request UI makes it easy to seen
when this is the case.)

However, I'm still not in favour of "git push -f", in conjunction with "git
rebase" or otherwise, because of the principle (which I mentioned to you
once before) that commit history made visible to others should not be
altered.

PR #853, opened 2018-06-15
Add theorems about rotate-right and rotate-left

airbornemihir:
The hypotheses about positivity of width probably can't be avoided since
they are also hypotheses for the theorems unsigned-byte-p-of-rotate-left
and unsigned-byte-p-of-rotate-right.

Thanks are owed to @kini for the idea.

### DIFF
--- a/books/centaur/bitops/rotate.lisp
+++ b/books/centaur/bitops/rotate.lisp
@@ -29,7 +29,7 @@
 ; Original author: Jared Davis <jared@centtech.com>
 
 (in-package "BITOPS")
-(include-book "std/util/define" :dir :system)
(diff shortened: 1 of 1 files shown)


airbornemihir:
Trying to weaken the hypothesis of unsigned-byte-p-of-rotate-left causes it to create tens of thousands of subgoals, which presumably means that the theorem prover is getting confused. This is just what I learned from examining the .cert.out file though - let me see what I can figure out in the read-eval-print loop.

airbornemihir:
I didn't fare any better in the read-eval-print loop. Feel free to make changes if you can see how to get it to work.

solswords:
This looks fine to me.

PR #858, opened 2018-07-11
VL Patches

jaredcdavis:
Several VL patches.  Mostly preprocessor/parser bugfixes/extensions.  Also some efficiency tweaks, linting improvements, and misc updates.

(Doing this as a pull-request so heavy VL users can test it before merging.)


kini:
It's a pleasant surprise to see a PR from you! I see from the commit log that it's been brewing for some time :) As you suggest, I'll let it sit open so VL users can comment.

airbornemihir:
Yep, it sure is! I hope this means Apple has become more relaxed re: legal
stuff.

Le mer. 11 juil. 2018 11:18, Keshav Kini <notifications@github.com> a
écrit :

solswords:
Looks like an interesting grab bag of new support for awkward language constructs :)  I'm tickled that you added support for bind, in particular.

Looks like you ran into the problem with very wide wires in sv-use-set.  I tried to address that a while ago by representing masks and sometimes values as sparseints, so you might check whether it works for you now (if you have a small example where it still goes out to lunch I might be interested).  Anyway I'm fine with having a no-sv-use-set linter option but would prefer it default to NIL.  Only nitpick so far from browsing through the diffs.

jaredcdavis:
Hi Sol,

Sure, we can definitely change the defaults so that sv-use-set is enabled unless the user explicitly says to turn it off.  I'll investigate whether I still need to do that to work around any issues.

IIRC, the bind support is extremely limited.  I thought about implementing something more general, but I think it would have required performing brain surgery on elaborate, and I definitely didn't want to go there...

Cheers,
Jared

solswords:
Is there actually a way to turn off a Boolean defoptions option if its default is T?

jaredcdavis:
Huh.  I guess not.  We should definitely change the default to `nil` in `kit/lint.lisp`.  If this blows up `linttest/trunc` we might hack it by just turning off sv-use-set in the Makefile there, for now.

solswords:
I'm good with this provided there's at least a way to enable sv-use-set. Anyone else still looking it over?

ragerdl:
Other than being pleased to see it, I haven't looked at this PR in detail.  That being said, maybe it should be closed and re-opened against acl2:testing?

kini:
I can just merge it into testing; then when testing is merged into master
by leeroy, this PR will be marked as merged. That seems simpler.

2018年7月13日(金) 12:05 David L. Rager <notifications@github.com>:

solswords:
OK, I'm good with that, @kini -- will patch the no-sv-use-set thing after you merge

kini:
I've merged this PR into the testing branch. Once leeroy is finished testing it, it will get merged into master automatically and this PR will be marked as merged. Thanks for the updates, Jared, and thanks to Sol for reviewing them!

jaredcdavis:
Thanks guys, sounds great.

On Mon, Jul 16, 2018 at 11:29 AM, Keshav Kini <notifications@github.com> wrote:
>
> I've merged this PR into the testing branch. Once leeroy is finished testing it, it will get merged into master automatically and this PR will be marked as merged. Thanks for the updates, Jared, and thanks to Sol for reviewing them!
>
> —
> You are receiving this because you authored the thread.




-- 
11410 Windermere Meadows
Austin, TX 78759
https://www.kookamara.com/jared/

PR #859, opened 2018-07-14
ipasir updates to make guards executable and optional extended functions to the interface

robsumners:
These updates do two primary things:

1. Add enough structure to the ipasir$c stobj to allow "tracking" of the underlying ipasir state to support executable guards for the ipasir logic state.

2. Add support for a couple of new function interfaces intended for MINISAT-based sat solvers.. get-stats and bump-activity-vars which allow the user to gather stats from the running ipasir (most importantly, the number of current unresolved clauses), and for bumping the activity of certain var.s to make them more likely to be chosen as decisions.

### DIFF
--- a/acl2-fns.lisp
+++ b/acl2-fns.lisp
@@ -1230,13 +1230,16 @@ notation causes an error and (b) the use of ,. is not permitted."
                       (if (fboundp 'interface-er)
                           'interface-er
                         'error)
-                      "When the ACL2 reader tries to read #\\x, then ~
-                       x must either be a single character followed ~
-                       by a character in the list ~x0, ~
-                       or else x must be one of Space, Tab, Newline, ~
-                       Page, Rubout, or Return (where case is ignored). ~
-                       However, ~s1 is none of these."
+
+                      "When the ACL2 reader tries to read #\\x, then x must ~
+                       either be a single character or one of Space, Tab, ~
+                       Newline, Page, Rubout, or Return (where case is ~
+                       ignored).  What follows x must be a character in the ~
+                       list:~|~X01.~|However, ~s2 is neither a single ~
+                       character nor one of Space, Tab, Newline, Page, ~
+                       Rubout, or Return (where case is ignored)."
                       *acl2-read-character-terminators*
+                      nil
                       x)))))))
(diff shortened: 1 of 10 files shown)


ragerdl:
Just in case someone misses it, since this PR involves "acl2 system" files, @MattKaufmann is the only person who should accept this pull request.  I presume Matt is talking offline with Rob about it.

ragerdl:
Hmmm, actually, looking at the commit log history, maybe this is an issue with merging.  I'd think that Matt shouldn't show up in the commit log history.  @robsumners do you have an idea what happened?  Was this a collaboration between you and Matt, or is this an artifact of how you did the merges?

solswords:
It actually doesn't -- I don't know why it looks like that, but if you fetch it you'll see it's just Rob's one commit (at the end) on top of a previous commit that was already in testing/master.

ragerdl:
I'd prefer to remove the extra commits, as they may require Matt resetting his purity script, and they're also not really indicative of "what we mean."  @robsumners @kini , would one of you mind helping clean this up?

airbornemihir:
I was puzzled by this too - it seems Rob's one commit
(99632db5a6559a3ac9cc4cb3c91ad66445516cfb) has the parent
8388ac10289d5cab791953238057294604af6d60 ,which Matt created in the course
of his regular development. Possibly Rob did a rebase and force-push.

airbornemihir:
The dates of the commits actually don't line up with the rebase/force-push
hypothesis I suggested above - it looks like Matt did his commit at 10:07
CT, Rob did his commit at 11:22 CT, and then the pull request email went
out at 11:28 CT (all on Saturday the 14th.) I'm inclined to conclude GitHub
is being glitchy; there will be no extra commits when Matt (or J) should
get around to merging this.

robsumners:
This is some bad merge issue. My commit should only have the ipasir directory file changes. None of Matt's changes showed up in my status and push to my version of testing, so I am not sure how this happened (doesn't look like a ships passing in the night type of thing either).. So I will somehow cancel my pull request (or one of you could kill it). I will create a clean clone of testing and go through the process again and send (hopefully) a clean commit. It should only be the 6 ipasir files that are being changed (well, 4 changed and 2 added).

PR #863, opened 2018-07-19
Prep

airbornemihir:
This is to move some things into the community books ahead of a pull request for my filesystem books themselves.

### DIFF
--- a/books/kestrel/utilities/characters.lisp
+++ b/books/kestrel/utilities/characters.lisp
@@ -233,7 +233,12 @@
   (defrule chars=>nats-of-append
     (equal (chars=>nats (append chars1 chars2))
            (append (chars=>nats chars1)
-                   (chars=>nats chars2)))))
+                   (chars=>nats chars2))))
+
+  (defrule nth-of-chars=>nats
+    (iff (integerp (nth n (chars=>nats chars)))
+         (< (nfix n) (len chars)))
+    :induct (nth n chars)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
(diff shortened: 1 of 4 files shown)


acoglio:
This looks fine to me w.r.t. the changes to the Kestrel Books, but I'll let @jaredcdavis review it w.r.t. the changes to STD.

Also, the theorems nth-of-chars=>nat and nth-of-string=>nat look somewhat specific: have you considered, instead, proving that chars=>nat and string=>nat return integer-listp (with more-return) and having a more general theorem (implies (integer-listp x) (iff (nth n x) (< (nfix n) (len x)))? The latter is generated by (std::deflist integer-listp ...), or could be perhaps added to std/typed-lists/integer-listp.lisp. These two more general theorems should subsume the somewhat specific ones.

jaredcdavis:
Hi,

This looks fine -- just two minor quibbles:

  - I would rename `nth-when-unsigned-byte-listp` to `unsigned-byte-p-of-nth-when-unsigned-byte-p`, as described in http://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/?topic=ACL2____NAMING-REWRITE-RULES

  - The `nth-when-unsigned-byte-listp` rule could perhaps be strengthened to avoid `(< key (len l))` as a hypothesis, by moving it into the rhs.  On the other hand, the weaker rule may be preferable sometimes anyway, so I don't propose doing anything about this in particular unless you want to.

Cheers,
Jared

airbornemihir:
@jaredcdavis I'll do the rename. IIRC, when I first came up with the other rule a while ago, I tried moving the hypothesis to the RHS as you suggested, and there was some reason it didn't work... but maybe I can try again.

ragerdl:
@airbornemihir are you ready for this to be merged?

acoglio:
I would advocate for first exploring the suggestion I made above, about the more general theorems that should subsume the more specific ones.

PR #864, opened 2018-07-25
Abbreviate output from osets pick-a-point strategies

kini:
The function defined by automate-instantiation computes a hint, prints it as a "suggestion" to the user, and then immediately provides that same hint to the theorem prover.

Since the hint is being provided to the prover anyway, there doesn't seem to be a need to also tell the user about it, and these hints are often very large and contribute a lot of noise to the proof output when trying to prove a large theorem that uses pick-a-point subset reasoning a lot.

So, this commit greatly abbreviates the message that the computed hint prints.  The new message is only one or two lines long, and still tells the user how to disable the hint (in case the pick a point strategy is ruining the proof), but doesn't supply any further information.

---

If you don't want to abbreviate the message, I could also roll the PR back to the previous commit in the sequence, which simply makes the message more accurate without abbreviating it.

### DIFF
--- a/books/std/osets/computed-hints.lisp
+++ b/books/std/osets/computed-hints.lisp
@@ -136,12 +136,12 @@
 
 (defun rewriting-goal-lit (mfc state)
   (declare (xargs :stobjs state)
-	   (ignore state))
+           (ignore state))
   (null (mfc-ancestors mfc)))
 
 (defun rewriting-conc-lit (term mfc state)
   (declare (xargs :stobjs state)
-	   (ignore state))
+           (ignore state))
   (let ((clause (mfc-clause mfc)))
     (member-equal term (last clause))))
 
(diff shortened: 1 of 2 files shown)


jaredcdavis:
FWIW, this is completely fine with me.  Thanks!

MattKaufmann:
Shall I merge this pull request?  I don't use this work myself, but I see no reason to delay further.  Looking for advice here....

ragerdl:
Seems okay to merge to me.  Though, I was going to leave it to Jared or Keshav (probably Keshav).

MattKaufmann:
Good idea, David.  Since Keshav has merge privileges and Jared has said "OK", I'll leave it to Keshav to merge when ready.

kini:
Thanks for the review, @jaredcdavis! I'll merge the PR now, then.

PR #865, opened 2018-08-01
Fix 3 typos

ericsmithkestrel:


ragerdl:
It took me a moment to confirm that `books/build/doc.lisp` isn't the meta-file that Matt frequently updates.  It's only 1.2k LoC, so that's not it.  Thanks for the fixes -- merging.

MattKaufmann:
In case anyone is interested, I thought I'd clarify here about "the
meta-file that Matt frequently updates" that David mentioned....

The ACL2 system documentation, which anyone in the acl2-books group is
welcome to update, resides in books/system/doc/acl2-doc.lisp.  Part of
my process involves generating, from that file, a file doc.lisp in the
top-level ACL2 directory.  That doc.lisp file is what supports typing
:doc at the terminal to get system documentation, even before
including any books.  After evaluating the form (include-book
"xdoc/top" :dir :system), the :doc command is hijacked so that it uses
all documentation defined by xdoc, rather than merely what is in
doc.lisp.

ragerdl:
Thanks for the clarification @MattKaufmann .  I suspected it was a top-level file but wasn't sure -- good to know and an easy way to remember!

PR #866, opened 2018-08-01
Minor VL patches

jaredcdavis:


solswords:
I changed it to merge into testing rather than master, but otherwise looks good.  Will merge in a bit if no one else weighs in.

jaredcdavis:
Cool, thanks.  I'll try to remember to use testing in the future.

PR #867, opened 2018-08-05
Expand the manual page about Z3 installation.

acoglio:
These additional notes provide slightly alternative instructions, for installing Z3 machine-wide on Mac.

I'm making a pull request instead of pushing because this is @pennyan's file and so she has to agree to the change. Alternatively, I can create a separate file somewhere else with :doc z3-installation as :parent.

Also, if any Python experts have corrections or improvements in regard to the prefix/package issue, please let me know.

### DIFF
--- a/books/projects/smtlink/doc.lisp
+++ b/books/projects/smtlink/doc.lisp
@@ -455,6 +455,71 @@ sat
 result.</p>
 </li>
 </ul>
-")
+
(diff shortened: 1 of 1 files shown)


pennyan:
Hi @acoglio , this looks good to me! Thank you for helping fulfilling the document!

PR #870, opened 2018-08-28
GL cleanup

solswords:
This pull request changes GL in a potentially backward-incompatible way: mainly, symbolic objects no longer contain a format for bit-level encoding of all complex rationals, just integers.  This makes sense given typical usage of GL, especially since rewrite rules can now be used to manipulate rationals/complexes at the term level with symbolic integers inside.  I am making this pull request in order to give anyone who uses GL a chance to check that these changes don't break their proofs.

Basically, the g-number symbolic object constructor (which used to take 4 lists of bits: numerator/denominator of the realpart/imagpart) no longer exists; instead, there is a g-integer symbolic object constructor that just takes a single list of bits.

To avoid breaking everyone's proofs immediately, g-number is still supported as a shape specifier, i.e. it can still exist in the g-bindings of a def-gl-thm.  It can now only take one list of bits (the numerator of the realpart), which likely was all anyone used.  G-integer is also supported as a shape-spec and works the same way, except for the syntactic difference:
```
(g-number (list (list 1 2 3)))  --> (:g-number (1 2 3))
(g-integer (list 1 2 3))        --> (:g-integer 1 2 3)
```
This replaced a different shape-specifier called g-integer, and I also removed another one called g-integer?.  I doubt anyone used these, but they could probably be replaced with term-level constructs if needed.

I also stripped out hand-coded symbolic counterparts for code-char, coerce, and intern-in-package-of-symbol, and replaced them with a (hopefully comprehensive) set of rewrite rules.

If you use GL, please check whether these changes break your proofs and let me know in the comments on this pull request.


ragerdl:
Glad to see you being willing to break stuff to clean it up.  Do you print warnings in the case that a soon-to-be-unsupported feature is used?  I'll leave it to @kini to chime if we've got any preference to hold off, but I'm guessing you're good to merge.

solswords:
At the moment nothing's really deprecated, it's just either there or gone.  The g-number shape spec thing is a wart but I don't think I want to deprecate it because it's used in too many places and it doesn't complicate things much.  (At worst I could handle it as a preprocessing step to turn the g-numbers into g-integers, so I don't feel any compelling need to drop it.)

PR #876, opened 2018-09-17
Workshop related updates

airbornemihir:
One of these commits touches the Kestrel books (making a rewrite rule more general) so it would be great if @acoglio or @ericsmithkestrel could review and merge.

### DIFF
--- a/books/kestrel/utilities/strings/chars-codes.lisp
+++ b/books/kestrel/utilities/strings/chars-codes.lisp
@@ -146,7 +146,9 @@
            (repeat n (char-code char)))
     :enable repeat)
 
+
   (defrule nth-of-chars=>nats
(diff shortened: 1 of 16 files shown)


MattKaufmann:
The rule indeed matches more generally, but it also eliminates integer-range-p and
can introduce a case split.  (I don't know if that's good or bad, but it does change
the library other than just generalizing the rule.)  So it's good that one of the Kestrel
guys has been asked to vet this change.

ericsmithkestrel:
I think the change to nth-of-chars=>nats is a good one.  You want the case split if i may be too big.  Also, this is not my rule, but I never liked the use of integer-range-p in the hyp.

acoglio:
This change is fine with me.

PR #882, opened 2018-09-22
Generalise lemmas and add lemmas

airbornemihir:
It would be cool if @ericsmithkestrel or @acoglio could review and merge this.


acoglio:
Thanks for using and contributing to this library.

nth-of-nats=>chars looks good, and consistent with nth-of-chars=>nats (which you generalized earlier in the same way). Also thanks for removing the extra blank line before the latter.

Ideally the inversion theorem chars=>nats-of-nats=>chars could be phrased as the unconditional (chars=>nats (nats=>chars nats)) = (unsigned-byte-list-fix 8 nats), which should be true if both functions fix their arguments to their guards (if they don't, we should change them to do that), but I don't think there's an unsigned-byte-list-fix in the libraries yet.  The version with fix-true-list that you wrote is "in between" (as far as fixing goes) this ideal one and the one in the corollary. It may be better to replace fix-true-list with STD's list-fix, since there may be more theorems about this -- these string utilities are meant to extend the STD string library. Perhaps @jaredcdavis could comment on this particular issue of fix-true-list vs. list-fix. Finally, since the inversion theorem involves both conversions chars=>nats and nats=>chars, I would make a separate defsection with it, which eventually can contain the other inversion theorem too, as done for example in [books]/kestrel/bitcoin/base58.lisp, defsection base58-val<=>char-inversion-theorems.

I would rename nth-of-nats=>string to nth-of-explode-of-nats=>string for clarity. Also, it feels like this theorem is combining things that could be more modular, like an nth-of-explode theorem to rewrite (nth i (explode s)) to (char s i) (with any needed hyps) and a separate char-of-nats=>string in this library. I guess the theorem could be added as is (with the suggested name change above) and re-modularized later, but perhaps it's easy to do now.

chars=>nats-of-make-character-list looks good.

I would rename length-of-nats=>string to len-of-explode-of-nats=>string for clarity. Similar observation to the theorem above, namely that maybe it could be made more modular, via a len-of-explode and a length-of-nats=>string.

Maybe @ericsmithkestrel has additional comments.

MattKaufmann:
Regarding:

> Perhaps @jaredcdavis could comment on this particular issue of
> fix-true-list vs. list-fix.

Yes, I agree that this would be interesting.

I did a little experiment, and I'm wondering if it would be good to
replace, with permission, fix-true-list in the ACL2 sources by
list-fix.  I did this little experiment using ACL2 8.1 built on CCL.

ACL2 !>(include-book "std/lists/list-fix" :dir :system)
[[.. output omitted ..]]
ACL2 !>(defconst *c* (make-list 10000000))
[[.. output omitted ..]]
 *C*
ACL2 !>(time$ (null (list-fix *c*)))
; (EV-REC *RETURN-LAST-ARG3* ...) took 
; 0.06 seconds realtime, 0.06 seconds runtime
; (32 bytes allocated).
NIL
ACL2 !>(time$ (null (fix-true-list *c*)))

***********************************************
************ ABORTING from raw Lisp ***********
********** (see :DOC raw-lisp-error) **********
Error:  Stack overflow on value stack.
While executing: FIX-TRUE-LIST
***********************************************

[[.. etc. ..]]

-- Matt

acoglio:
Yes, I received an email from GitHub about your new commit, which looks good.

I agree with expressing things with built-in functions as much as possible. In this particular case, list-fix is already present in the book, and my concern was about the interplay of this new rewrite rule with other rewrite rules that generally use list-fix. It would be great if list-fix and fix-true-list are unified into one; in general, I think it's unfortunate to have synonymous or quasi-synonymous functions, also because rules about the different functions may not interplay.

I hadn't look at char closely, but given that it has that definition and is enabled (I always disable my functions, so I tend to think of every function as disabled) than I agree that the rule I proposed would be a bad idea! So scratch that suggestion. It looks like length is also enabled, so the same applies to my related suggestion about the other rule.

acoglio:
It looks like fix-true-list and list-fix won't be unified right away, so we should go ahead and make a choice about these theorems. I'm okay with leaving them with fix-true-list for now, but if we see that they don't play well with other STD theorems, we may have to replace that with list-fix.

As a data point, I did text searches for '(fix-true-list' and '(list-fix' in [books]/ and it is used in relatively few books and there are many more uses of the latter than the former.

airbornemihir:
Another, mostly unrelated, thing that occurred to me - I'm not sure an 
unsigned-byte-list-fix function could be defined, at least, not in a way 
that would play well with the common notion of fixing a list type. Since 
the :guard is t both for unsigned-byte-p and unsigned-byte-listp, 
non-natp values are possible for n/bits, and the following are theorems.
--
(thm (not (unsigned-byte-p -1 x)))

(thm (implies (unsigned-byte-listp -1 x)
           (equal x nil))
      :hints (("Goal" :in-theory (enable unsigned-byte-listp))))
--
Thus, if we wanted the following to be a theorem...
--
(thm (unsigned-byte-listp n (unsigned-byte-list-fix n x)))
--
... I imagine we would have to define the fixing function like so.
--
(defun unsigned-byte-list-fix (n x)
   (declare (ignore n x))
   nil)
--

So I don't think this would be a good solution for the fixing we're 
trying to do here.

Mihir.

ericsmithkestrel:
The current set of changes looks pretty good to me.  I would suggest splitting out the corollary of chars=>nats-of-nats=>chars into a separate rule so that each of the two rules can be easily enabled/disabled/used.  But I have my own version of all this machinery, so it's not a big deal for me.

acoglio:
Unsigned-byte-list-fix can be defined analogously with integer-range-list-fix in [books]/kestrel/utilities/integer-range-lists.lisp. The theorem asserting that unsigned-byte-list-fix returns an unsigned-byte-listp must have a hypothesis about the argument n (otherwise we would run into the problems that you point out), similar to the hypotheses on the lower and upper arguments in [books]/kestrel/utilities/integer-range-lists.lisp.

If (un)signed-byte-p (and integer-range-p) were changed to fix their size/bound arguments so that the predicates are never empty, then we could avoid those hypotheses too. For example, [books]/kestrel/utilities/digits-any-base.lisp has parameterized recognizers and fixers for digits in arbitrary bases (the base being the parameter) that fix the parameter, and the theorems generally need no hypotheses on the parameter.

airbornemihir:
Right, we can definitely take this up later. Re: the suggestion from @ericsmithkestrel about splitting off the corollary into a separate rule, I think that has the potential for problems if we should come up with a nice fixing function for unsigned-byte-listp - we would want to replace both the theorem and its corollary with the fixed version, and having two rules with two different names (which people would be using for different purposes) would complicate this. If Eric agrees with this assessment, I'm OK with this being merged. Thanks again, Alessandro and Eric.

ragerdl:
Not sure where this PR is, but I'll point out that the conflicts in `books/doc/relnotes.lisp` need to be resolved before we can do anything more with it.  Thx!

airbornemihir:
@ragerdl I fixed the merge conflict just now. Unfortunately, because 
I've had to merge thrice for this same reason, the changes in this pull 
request have become kinda unwieldy to review. Still, I've discussed this 
above with Eric and Alessandro, and they have indicated they are OK with 
this being merged - so anyone should be able to do it. Thanks!

ragerdl:
Merging per Eric and Alessandro's transitively communicated approval.

PR #885, opened 2018-09-25
VL: linter tweaks

jaredcdavis:
- Suppress truncation and extension warnings when the RHS has the form {n{0}}.
   In either case you're just getting zeroes anyway and this seems fine.

 - Suppress extension warnings when the RHS has the form {'0,...}.  This will
   get zero extended as intended so all is well.

 - For fussy size-warnings, treat integer-valued system calls like
   $countones(...) more like plain integer constants.

 - Suppress qmark-width warnings for (foo ? bar : baz) where foo refers to a
   plain, untyped parameter.  This is a bow to, e.g., `parameter WITH_FOO = 1`
   where the user intends the parameter to be 0 or 1 but has not bothered to
   give a type.

 - Treat all expressions within parameter declarations as "indexy" for the
   purposes of left-right checking.  This should suppress warnings about
   constructs like `parameter foo = size + size`, which are unlikely to be
   errors.

 - JSON format: increase the wrap-column for the text of each warning to a
   larger number.  We previously word-wrapped the warnings to 80 columns, which
   was resulting in ugly wrapping in many cases.  (It probably made good sense
   originally, i.e., until the pretty-printer started preserving newlines.)

 - Add some test cases to validate the above.

### DIFF
--- a/books/centaur/sv/vl/trunc.lisp
+++ b/books/centaur/sv/vl/trunc.lisp
@@ -92,8 +92,19 @@ We basically don't want to complain about things like</p>
   ;; (:gateport ((portname stringp)
   ;;             (expr vl-expr-p))))
 
-
-
+(define vl-expr-is-{n{0}}-p ((x vl-expr-p))
+  (vl-expr-case x
+    :vl-multiconcat
+    (and (equal (len x.parts) 1)
+         (let ((first (car x.parts)))
+           (vl-expr-case first
+             :vl-literal
+             (vl-value-case first.val
+               :vl-constint (equal first.val.value 0)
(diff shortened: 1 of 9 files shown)


jaredcdavis:
Sounds great, thanks Sol!

On Tue, Sep 25, 2018 at 2:26 PM, solswords <notifications@github.com> wrote:
> Seems fine. Will merge in a few hours unless anyone else wants to weigh in.
>
> —
> You are receiving this because you authored the thread.



-- 
11410 Windermere Meadows
Austin, TX 78759
https://www.kookamara.com/jared/

PR #886, opened 2018-09-27
Update filesystem books

airbornemihir:
This pull request touches only my own books, with one exception. I have reworked some of @MattKaufmann 's edits to the 8.2 release notes from earlier today, to correct the unintended implication that the std/io byte-combining libraries are part of the Kestrel books - so it would be neat if he could review and merge.

### DIFF
--- a/books/doc/relnotes.lisp
+++ b/books/doc/relnotes.lisp
@@ -150,7 +150,7 @@
 
  <h4>Filesystem Books</h4>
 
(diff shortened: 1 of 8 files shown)


MattKaufmann:
The changes to :doc note-8-2-books look fine to me.  I'll leave it to someone else to do the merge, since there are lots of other changes and I haven't been following the relevant discussion.

PR #887, opened 2018-09-27
Updates to ACL2s

pmanolios:
Added a proof builder macro that repeats instructions until all goals have been proved.

Updated the base-theory to include the utilities book and updated the doc file since the new utilities file has new documentation.


MattKaufmann:
I see no problem with merging, since whatever Pete wants to do with ACL2s seems to me to be his business.  So I'll merge (without having checked anything, though I did look very briefly at the new proof-builder command and I saw no obvious problems with it; seems like a nice contribution).

ragerdl:
For what little it's worth, I tend to agree.  Quite glad to see additional contributions from the ACL2(s) folk!

PR #891, opened 2018-10-05
VL dependency reduction and support for alternate asdf-home locations

jaredcdavis:
- Reduce dependencies on OSLIB books such as "file-types" and "ls" throughout
   the VL books by just including the "logic" versions of these books instead.
   This generally reduces the use of trust tags, and also reduces the number of
   books that need to be recertified when changing QUICKLISP_ASDF_HOME,
   described below.

 - The quicklisp/base book can now be told, by way of a new QUICKLISP_ASDF_HOME
   environment variable, to write compiled files to a location other than
   books/quicklisp/asdf-home.  For this to work, this environment variable must
   be set consistently (1) when the books are built, and (2) when they are
   included.  Most users should have no interest in this new feature, but in
   some cases it may provide a more reliable alternative to hacks like the
   centaur/misc/sharedlibs book.


MattKaufmann:
Thanks, Jared!

Since I know almost nothing about quicklisp, it seems prudent for me to
leave it to someone who does (e.g. Sol, David Rager, maybe Alessandro) to
accept it.

ragerdl:
Looks great to me - thanks!

PR #892, opened 2018-10-08
Add a documentation topic

airbornemihir:


MattKaufmann:
I'll accept this, and then probably change the primary parent to io (but keep princ$ as a secondary parent, I guess).

PR #896, opened 2018-10-14
Adding supporting materials in workshops/2018/kwan-greenstreet/

kwancarl:


MattKaufmann:
I believe that many of these books depend on ACL2(r), but I didn't find the
following in any of them:

    ; cert_param: (uses-acl2r)

I'd be happy to accept this pull request if I'm mistaken, or if those
directives are added to the books (or at least, the "leaves" of the
graph of these books under the include-book relation).

kwancarl:
Yes, you're correct. Thank you for letting me know. I've added them now.

PR #931, opened 2018-12-17
Updated rtl/rel11, projects/rac, and projects/arm.

russinoff:


ragerdl:
Thanks for the submission!  As a reminder, please setup pull requests to merge into `testing` instead of `master`.

airbornemihir:
@ragerdl Is there any way you can change the default merge target to be testing? I know I've slipped on this particular banana peel more than once with my own pull requests.

airbornemihir:
@ragerdl There's a warning at the bottom of the page about Subversion (I don't know if any of our regular contributors use it for ACL2) and git ls-remote (the small number of folks who know about this command would probably not be confused.)

So the only unintended consequence we should be concerned about, I think, is what branch people are going to get when they clone the repository. Per @kini 's comment above, we are going to have trouble with that. So I guess we'll just have to agree among ourselves to keep making pull requests against testing - it's too much work to ask people to merge errant pull requests into testing on the command line, rather than use the GitHub GUI.

MattKaufmann:
> So I guess we'll just have to agree among ourselves to keep making
> pull requests against testing - it's too much work to ask people to
> merge errant pull requests into testing on the command line, rather
> than use the GitHub GUI.

I second that!  I'd be dismayed if "git clone" took from testing
instead of master, and I'm pretty sure that others could be, too.

-- Matt

PR #939, opened 2018-12-29
Factor out encapsulated lemmas

airbornemihir:
As I've noted in the commit messages, the rules remain the same, so no one should be affected by this change one way or the other. Still, since this is a Kookamara book, perhaps @jaredcdavis would like to review.


jaredcdavis:
Looks good, no objections here.  Thanks!

MattKaufmann:
I'll accept (merge) this pull request, since Jared has replied saying that he has no objections.

PR #947, opened 2019-01-20
RTL: Move old crt proofs from rel9-rtl-pkg to project/srt/support

nadezhin:


MattKaufmann:
As is typical with mods to the rtl library, I've gotten an OK from Russinoff before accepting the pull request.

PR #958, opened 2019-02-13
 Changed the package used by the enumerators of symbol and var to ACL2S from ACL2

pmanolios:


ragerdl:
Huzzah for moving more symbols out of the default package!

PR #965, opened 2019-03-12
Trim an STD rule

airbornemihir:
This pull request trims an unnecessary function call from an STD rule, fixes a regression which arises, and incorporates the new rule into the filesystem books.


acoglio:
The change to STD looks good to me but I'll let @solswords look at the change to aignet before merging.

solswords:
Looks fine, thanks?

PR #968, opened 2019-03-15
Add manual-fast and rendered-doc-combined-fast.lsp to .gitignore files.

ericwhitmansmith:


ragerdl:
Looks like there's precedence for ignoring these files, as they appear to just be additional versions of their originals.  Rebasing and merging.

PR #970, opened 2019-03-20
Number theory

ericwhitmansmith:
Make some include-books local and adjust proofs accordingly.  See the item I added to the relnotes.  This change was approved in principle by Russinoff, but he might want to review the details.  I also deleted some trailing whitespace (my editor does this automatically).


russinoff:
Thanks for doing this, Eric.

ragerdl:
That seems to be an approval from @russinoff so I'll rebase and merge.

PR #974, opened 2019-03-23
Update filesystem books and add IO lemmas

airbornemihir:
I'm not sure who should review the books/std/io/base.lisp changes, but I think it would suffice for @ericsmithkestrel to do so, since he responded to my question on Slack. I decided to follow the (apparent) convention that lemmas about built-in I/O functions go in the base.lisp file while community-defined functions each get their own file in that directory.

### DIFF
--- a/books/projects/filesystems/LICENSE
+++ b/books/projects/filesystems/LICENSE
@@ -1,0 +1,3 @@
+Copyright (C) 2017, Regents of the University of Texas
+Written by Mihir Mehta
+License: A 3-clause BSD license.  See the LICENSE file distributed with ACL2.
--- a/books/projects/filesystems/block-listp.lisp
+++ b/books/projects/filesystems/block-listp.lisp
@@ -1,7 +1,3 @@
-; Copyright (C) 2017, Regents of the University of Texas
-; Written by Mihir Mehta
-; License: A 3-clause BSD license.  See the LICENSE file distributed with ACL2.
-
 (in-package "ACL2")
 
 ;  block-listp.lisp                                  Mihir Mehta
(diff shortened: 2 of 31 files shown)


ericwhitmansmith:
Hi, I don't own std, and can't merge pull requests.  But I did take a look.  I  would suggest removing all hyps except the last one of stringp-of-read-file-into-string1.  Otherwise, the changes to std seem good to me.  (You could also make stringp-of-read-file-into-string1 a :type-prescription rule, but I don't feel strongly about that.)  Thanks, Mihir, for contributing this improvement.

ericwhitmansmith:
Responding to Mihir's note about hyps: It may be the case that no hyp individually could be removed (you have to remove them as a group, because they support each other inductively).  Perhaps remove-hyps doesn't handle that case.

airbornemihir:
@ericwhitmansmith : I haven't checked, but this seems plausible to me. I've made @MattKaufmann aware of this discussion about remove-hyps, but I'm tagging him here so he can take your hunch above into account.

MattKaufmann:
@ericwhitmansmith  I like your theory.  In fact I tried removing just one hyp at a time and the proof failed (even without the step-limit restrictions imposed by remove-hyps).  I suppose remove-hyps could be made more powerful by attempting removal of more than one hyp at a time, but that would also slow it down.

@airbornemihir  I've updated :doc remove-hyps by adding a paragraph about this limitation.  The bleeding-edge manual (http://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/index.html) is up-to-date with this change.  If you think it's not clear, feel free to change it further or to send me email to work out an improvement.  As far as I'm concerned, case closed. ;)  Of course, feel free to add options to remove-hyps that make it more aggressive.

MattKaufmann:
Oops, I clicked the button prematurely... manual update and git push should happen in a few minutes....

MattKaufmann:
OK, done; sorry about the noise.

airbornemihir:
No problem. Thanks for looking into this, Matt.

PR #978, opened 2019-04-03
[std/io & std/basic] Factor out bytep.

acoglio:
Making a PR since it affects Std. Please see commit message for details.

If there are no objections, I'll merge in a few days.

### DIFF
--- a/books/doc/relnotes.lisp
+++ b/books/doc/relnotes.lisp
@@ -388,6 +388,9 @@
  <p>Added new lemmas to the <see topic='@(url combine-functions)'>std/io
  byte-combining libraries</see>.</p>
 
+ <p>Factored out the @(tsee bytep) predicate into a new file under @(see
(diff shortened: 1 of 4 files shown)


MattKaufmann:
I have no objection, but here's an observation about the :exec tweak
mentioned in the commit message.  I did each of the following in raw
Lisp (after starting ACL2 and :q), and the results were identical, both
using CCL and (after ignoring differences in location) SBCL.

    (disassemble (defun foo (x) (natp x)))
    (disassemble (defun foo (x) (and (integerp x) (<= 0 x))))

That's probably because of the

    (declaim (inline ...))

form in axioms.lisp, which declaims natp to be inline.

acoglio:
Thanks @MattKaufmann for running these tests. That all makes sense. Anyways, happy to change it back to what it was, or for anyone who cares to do that.

PR #981, opened 2019-04-07
Doc update

airbornemihir:
I'm open to suggestions as far as the doc-string is concerned. I'm trying to convey that proof-by-arith works just as well for encapsulate or make-event as it does for defthm.


MattKaufmann:
I don't mind accepting this pull request, but the commit comment seems to suggest that I should maybe wait for someone to make comments (I'm not planning to proofread this :doc).

airbornemihir:
Hi Matt,

Given that no one's brought up any issues so far, it'd be nice if you could accept.

Thanks,
Mihir.

PR #987, opened 2019-04-18
Update developers' guide

airbornemihir:
This clears up a part of the developers' guide that I found a little
hard to follow. It also, while keeping the end result the same,
slightly changes the order of operations in the process of merging
a contributor's branch, in order to reduce the number of merge commits
produced. Since this pull request follows personal discussions as well as list discussions with @MattKaufmann , I'd prefer that he should merge.

### DIFF
--- a/books/system/doc/developers-guide.lisp
+++ b/books/system/doc/developers-guide.lisp
@@ -2027,10 +2027,9 @@
  <p><b>IMPORTANT</b>: Before reading this topic, be sure to read the topic,
  @(see developers-guide-maintenance).  The present topic assumes that you have
  followed the process there to make changes in your copy of ACL2 and the @(see
- community-books), including testing and documentation. In particular, you
- should have completed the devel-check process laid out in the topic, @(see
- verify-guards-for-system-functions).  Here are the steps for contributing your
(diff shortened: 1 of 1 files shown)


MattKaufmann:
@airbornemihir Thanks for the git instructions modification.  I'll take your word for that!

Regarding the new sentence about devel-check: I don't think that a devel-check is always necessary, but it's a bit difficult for me to characterize exactly when.  If it's a heuristic change, for example, then probably I wouldn't feel it's necessary.  But when a definition changes, as in your case, then it seems impoortant.

Also, it seems to me that any such addition really belongs in :doc developers-guide-maintenance, since that's the location of testing-related information.

You're welcome simply to remove this addition, leaving it to me to deal with the wording as something is added to :doc developers-guide-maintenance.  That might be less work for both of us.  But of course you're also welcome to take a crack at that yourself.  If you do, then you could suggest that it might be necessary, depending on the nature of the change, and one could check with an author (see the wording for that in the next-to-last step in :doc developers-guide-contributing).

MattKaufmann:
@airbornemihir I forgot to add that whatever is said about devel-check should probably also be said about doing "make proofs" (also in :doc developers-guide-maintenance).

PR #992, opened 2019-04-29
Add an STD book

airbornemihir:
This book is currently not used, but is likely to be used in a future pull request.


acoglio:
Thanks.

Would you be willing to include this file into `[books]/std/lists/top.lisp`, as is (I believe) customary for Std?

Also, for consistency with other similar Std files, `remove1-equal` should be disabled: would you be willing to try that and see what needs to be adjusted in the community books? Functions like these are typically disabled in `[books]/std/lists/top.lisp`, but I would argue that they should be disabled in the individual files that provide theorems about them, in case users include less than `top.lisp` -- this can be discussed more broadly with the community.

You might also consider making the second rule both a rewrite and a type prescription rule.

PR #995, opened 2019-05-03
[Std/lists] Remove TAKE-REDEFINITION and TAKE-INDUCTION.

acoglio:
[I'm making a PR, instead of pushing directly, to give the community a chance to review, since this affect Std and indirectly many books.]

Now that the built-in definition of TAKE has been changed to be essentially like
TAKE-REDEFINITION in Std, there is no need for this rule to remain, which in
fact was slowing down some proofs. The companion rule TAKE-INDUCTION, an
induction rule based on TAKE-REDEFINITION, has also been removed, because it
should be now equivalent to the built-in induction rule generated by the new
definition of TAKE.

Several books had to be adapted to this change. Occurrences of TAKE-REDEFINITION
(e.g. in hints) have been replaced by TAKE. It was also necessary to enable TAKE
in several proofs that were otherwise failing. An :EXPAND hint also had to be
added to one proof, to force expansion of TAKE in the inductive case of the
proof. In a few places, it was necessary to enable or disable other TAKE-related
rules, such as TAKE-WHEN-ATOM.

While the changes to the books were minimal and all the proofs work, book
authors may want to double-check their files and perhaps tweak things.

### DIFF
--- a/books/acl2s/base-theory.lisp
+++ b/books/acl2s/base-theory.lisp
@@ -64,7 +64,7 @@
 (sig acl2::rev ((alistof :a :b)) => (alistof :a :b)
      :suffix alistof)
 
-(sig nth (nat (listof :a)) => :a 
+(sig nth (nat (listof :a)) => :a
      :satisfies (< x1 (len x2)))
 ; PETE: I added and removed the sig below because it caused the
 ; theorem prover to get super slow.
@@ -100,6 +100,7 @@
 (sig first-n-ac (nat (alistof :a :b) (alistof :a :b)) => (alistof :a :b)
      :satisfies (< x1 (len x2))
      :suffix alistof)
+(local (in-theory (enable take)))
 (sig take (nat (listof :a)) => (listof :a)
      :satisfies (<= x1 (len x2))
      :hints (("Goal" :cases ((equal x1 (len x2))))))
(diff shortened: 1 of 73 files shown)


MattKaufmann:
@acoglio Thanks for tweaking the release note on this in acl2-doc.lisp.  I think you should get some credit too!  (Up to you though.)
You and I agreed that I could accept this pull request this morning, but maybe I'll wait till this afternoon in case anyone else has feedback.

acoglio:
@MattKaufmann Sounds good to wait a bit. I was out for a few hours last night and so I made this PR later than I planned to.

acoglio:
Just merged an edit to a readme file from @airbornemihir. This does not affect Std or `take`.

MattKaufmann:
I've timed the pull request version against master, and as far as I'm concerned, the difference is trivial.  So I'll probably accept this pull request after lunch so that I can start final Version 8.2 release preparations.  Results are below (on a 4-core, 8-thread machine using -j 8).

    ;;; old
    129875.952u 1990.960s 5:09:42.93 709.6%     0+0k 3740160+8736328io 1286pf+0w

    ;;; new
    130730.996u 1864.796s 5:10:40.27 711.3%     0+0k 3810184+8468264io 1328pf+0w

PR #997, opened 2019-05-16
Changing startjob from bash alias to env variable

ragerdl:
I don't know how the bash alias ever worked, but changing it to
an environment variable should do the trick.


ragerdl:
Here's some fun documentation on aliases not working the way it looks like they were meant to be used.

https://unix.stackexchange.com/questions/1496/why-doesnt-my-bash-script-recognize-aliases

PR #1002, opened 2019-06-06
Improve books makefile

ragerdl:
I started down the route of changing GNUmakefile but stopped, because I wasn't sure how to go about testing all of the changes in a reasonable amount of time.  Do I need to test it on every Lisp?  Do I need to test it with every version of ACL2?  Do I need to test all of the different targets?  I did test it on CCL with the default (hons) version of ACL2.

Also, I'd feel more comfortable if someone were to check the logic changes for removing the slow books.  I know it is a change, but is it a reasonable one, and did I implement it correctly?  Please keep in mind that the primary goal is to make the build more maintainable, and having all sorts of random functionality isn't all that important relative to that first goal, imho.

Please don't merge this directly.  I've used "WIP:" as a prefix, which arguably prevents a merge from occurring on gitlab (but this is github).  There's a handful of commits to cherry-pick.

I renamed the regression/pull-request target to "pull-request".  I just figured we should call a smurf a smurf (and only once), but if the community wants to call it "regression" instead, that's quite fine by me.

Thus, I submit this to the community, as a draft, and hope that someone else will pick it up.  There are some obvious changes that still need to occur, like change the documentation to accommodate the new targets (both the documentation in GNUmakefile and the contribution xdoc).  And, I think the community can decide how much testing they think is needed.

Maybe after some feedback I'll be convinced that it's not too much work to just finish it myself.  But, I'm also happy to have someone else finish it :).

### DIFF
--- a/books/GNUmakefile
+++ b/books/GNUmakefile
@@ -50,18 +50,19 @@
 #                 and other miscellaneous tools and macros.  Many users may
 #                 find "basic" to be a convenient starting place.
 #
-#   all           Certifies most books that are not horribly slow, including,
+#   regression    Certifies most books that are not horribly slow, including,
 #                 for instance, most of workshops, projects, the centaur
 #                 books, the jvm models, etc.  Usually, committers to
-#                 acl2-books should run "make all" first.
+#                 acl2-books should run "make regression" first.
 #
-#   almost-everything
-#                 An alias for "all"
-#
-#   everything    A very full build, including very slow books.  Most users
+#   regression-everything
+#                 A very full build, including very slow books.  Most users
 #                 will not want to use this target.  It is useful for, e.g.,
 #                 regression testing before releases.  Note that
-#                 EXCLUDED_PREFIXES may not work with the `everything' target.
+#                 EXCLUDED_PREFIXES may not work with the
+#                 `regression-everything' target.
+#
+#   all           alias for regression-everything
 #
 #   <dirname>     There are targets for many top-level directories, e.g., you
 #                 may run "make coi" to build everything in coi/ directory,
@@ -185,7 +186,7 @@ export ACL2
 .SUFFIXES:
 .SUFFIXES: .cert .lisp
 
-.PHONY: default basic all
+.PHONY: default basic regression
 
 # Keep this before any other target so that default will be the default target
 default: basic
@@ -422,9 +423,9 @@ ifneq ($(ACL2_HAS_REALS), )
 $(info Excluding non-ACL2(r) books: [$(CERT_PL_NON_ACL2R)])
 OK_CERTS := $(filter-out $(CERT_PL_NON_ACL2R), $(OK_CERTS))
 
-all: nonstd/workshops/1999/calculus/book/proof-outline.cert
-all: nonstd/workshops/1999/calculus/book/tree.cert
-all: nonstd/workshops/1999/calculus/book/tree/proof-outline
+regression: nonstd/workshops/1999/calculus/book/proof-outline.cert
+regression: nonstd/workshops/1999/calculus/book/tree.cert
+regression: nonstd/workshops/1999/calculus/book/tree/proof-outline
 
 # Some additional special instructions for generated ACL2(r)-only books.
 nonstd/workshops/1999/calculus/book/proof-outline.cert: nonstd/workshops/1999/calculus/book/tree.cert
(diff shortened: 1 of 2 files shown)


ragerdl:
Also, I couldn't figure out what we're currently doing with USE_QUICKLISP.  It seems to me that USE_QUICKLISP is actually set (by cert.pl or something like it) when building the `all` target.  So, the comment about USE_QUICKLISP and the `all` target is perhaps outdated.  Personally, I think we should get rid of USE_QUICKLISP, and act as if it were `1` whenever the chosen lisp supports it.  However, it's important to legal departments and defense contractors to be able to disable Quicklisp.   Any thoughts, @TheBeaNerd ?

MattKaufmann:
Hi, David and acl2-books --

I have little to contribute regarding modification of
books/GNUmakefile -- as long as nothing breaks then I think I'll be
happy, and I plan to modify the top-level GNUmakefile to match the new
books/GNUmakefile when it's installed.

Since the scope of my original plan to remove "certify-books" has been
expanded, I plan to leave it to someone else to do that removal --
it's off my calendar now.

Regarding quicklisp: I'm pretty sure that GCL doesn't work with
quicklisp, so whatever the change, I suggest running a GCL regression
of some sort before changing books/GNUmakefile.  FYI, here's what I
actually run for GCL on an 8-thread machine, in the top-level
directory ("regression-fresh" invokes the "all" target of
books/GNUmakefile):

time nice make -j 8 regression-fresh ACL2=<path_to_acl2> GCL_MEM_MULTIPLE=0.25

-- Matt

airbornemihir:
@ragerdl I was looking at this pull request in git, and I noticed that a rebase seems to have gone wrong somewhere, causing five commits to be repeated. For instance, 05cce4d76748037438c301c10378fe4eb1b04e76 and 3f4cf3a72e9ace6308d72ebc45fbdd47ee83b6bf have the same commit message and seem to have the same contents.

kini:
I took the liberty of converting this PR and #1118 to "drafts", which are a new-ish GitHub feature that formalize the notion of putting "WIP" in the title: https://github.blog/2019-02-14-introducing-draft-pull-requests/

ragerdl:
I'd like to revisit this PR, which now has updated code.  My goal with this PR is two-fold:

(1) Make the names of the books build targets easier to understand.  I believe this PR achieves this for two reasons: (a) "all" now builds all of the books, and (b) the names of the new targets are now consistent with the targets in the ACL2 Theorem Prover GNUmakefile.

(2) Simplify the ADDED_BOOKS/SLOW_BOOKS thing.  I believe this is achieved, because this PR removes the ADDED_BOOKS part.  This should make it easier for us to understand what the build commands are actually doing.

This PR _does not_ address the question of whether USE_QUICKLISP should be automatically set by the `regression`, `regression-everything`, or `all` targets.  I'd like to leave that discussion for another day.  Comments on that question are welcome, but I don't expect to incorporate them in this PR.  

After confirming that there is no strong objection from @MattKaufmann , I plan to document these changes in the books release notes and also in the git instructions, as appropriate.  If you know of another place I should put a note, please lmk.  As discussed earlier in this PR, @MattKaufmann would then probably update the ACL2 Theorem Prover's GNUmakefile targets (or I can do it if he's okay with updating the purity-script mechanism).

I've tested the default, basic, regression, regression-everything, and all targets on CCL, checking that each, at least approximately, builds the right set of books.  I've got runs going for SBCL and other variants of ACL2 too.

Thanks,
David

MattKaufmann:
@ragerdl I'll try to look at this soon (though you hit me at an unusually busy time).

ragerdl:
There's something going on in SMTlink, but I suspect it's a jenkins-only issue.  The error is:

```
HARD ACL2 ERROR in SMT-PROVE=>MAKE-FNAME:  Error: Generate file error.
```
And can be viewed more extensively at http://leeroy2.defthm.com/job/acl2-multiple-flavors-and-lisps-for-new-makefile/ACL2_PAR=no_p,ACL2_REAL=no_r,LISP=ccl/3/console

That being said, I don't hit this problem when I run outside of jenkins, so I'm guessing it's a jenkins build configuration issue.

books/GNUmakefile
```diff
@@ -1181,23 +1173,18 @@ ifeq ($(shell ls workshops 2> /dev/null), )
...
 
-all: $(OK_CERTS)
-
-.PHONY: almost-everything
-almost-everything: all
+regression: $(filter-out $(SLOW_BOOKS), $(OK_CERTS))
 
-# Starting 8/15/2014, everything does not automatically set USE_QUICKLISP=1,
-# even in the HONS case.  In particular, "make everything" no longer makes
+# Starting 8/15/2014, all does not automatically set USE_QUICKLISP=1,
```
ragerdl, line 1178: I should probably change this to `regression-everything`

MattKaufmann:
I've taken a look at all the changes, and they look reasonable as far as I can tell.  As I mentioned in my "Jun 6, 2019" comment, "as long as nothing breaks then I think I'll be happy."  I prefer to leave books infrastructure maintenance to others when that seems feasible.

I'd like to add temporary lines as shown below to books/GNUmakefile right now, in which case I'd update the top-level GNUmakefile accordingly right now as well. That way I wouldn't need to modify the top-level GNUmakefile when your changes go in -- everything would work right away.  Your changes would throw away these temporary additions.

    .PHONY: regression regression-everything
    regression: all
    regression-everything: everything

@ragerdl, is that OK with you?

By the way, I've made a note to myself so that after the dust settles, I'll update installation/installation.html where it says "make all".

ragerdl:
@MattKaufmann Thanks for taking a look!  I like your idea about adding those temporary lines so the top-level makefile can get "ahead" of this commit.  Please proceed.  Thanks!

ragerdl:
I took a look at the git quickstarts, and the instructions assume one is in the top-level directory (as opposed to the books directory), so they're already using the top-level `regression` and `regression-fresh` targets.  Thus, the git quickstarts do not need to be updated.

MattKaufmann:
@ragerdl What are "the git quickstarts"?

ragerdl:
Ah, sorry, I could have been more precise.  The git quickstart :doc topics.  E.g. http://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/index.html?topic=ACL2____GITHUB-COMMIT-CODE-USING-PULL-REQUESTS

ragerdl:
I've merged in Matt's latest changes and am running a test on leeroy2 before I accept this pull request and merge it into `testing` (at which point it will run again on leeroy2, but in a more official manner).

ragerdl:
Everything-regression passed according to http://leeroy2.defthm.com/job/acl2-refactor-makefile/9/console so merging (via fast forward merge)

PR #1011, opened 2019-07-25
[crypto] add keccak/sha-3 spec and tests

bendyarm:


ragerdl:
Neat, thanks for adding a release note.  Normally I'd await @acoglio's approval for modifying the kestrel directory.  However, since you're from Kestrel, I think this is good to go and will merge.

PR #1030, opened 2019-09-04
WIP: Update to defcong

pmanolios:
Defcong doesn't work the way I'd like when one is not in the ACL2
package. The problem involves symbol names.  If I add a hint of
the form :induct (f x x-equiv), when in say the ACL2s package,
then I will not get the results I intended because chances are
that I should have written :induct (f acl2::x acl2::x-equiv).
Also, the names of the theorems generated will often be in the
ACL2 package. It would be a lot more convenient if defcong
interned symbols in the current package. I modified defcong to do
that, but when I tested it, I got errors. The issue is that the
only way I can see to do that in ACL2 is to use make-event, but
defcong is used in the boot-strap process and make-event is not
allowed inside an encapsulate during boot-strap. The current
definition of defcong is a compromise. First, instead of using
the package of equiv1, I use the package of (car fn-args),
because I expect that this will be more strongly correlated to
the current package than equiv1.  Second, defcong now includes
the use-current-pkg keyword arg which by default is nil. If it is
nil, then defcong works as before (modulo using the function
symbol to pick the package), but if it is t, then it uses
make-event to generate symbols in the current package.

It would be nice to have some standard way of generating symbols
in the current package that doesn't require all of this pain.

### DIFF
--- a/defthm.lisp
+++ b/defthm.lisp
@@ -11647,6 +11647,7 @@
      ,@(if otf-flg (list :otf-flg otf-flg) nil)
      ,@(if doc (list :doc doc) nil)))
 
+#|
 (defmacro defcong (&whole x
                           equiv1 equiv2 fn-args k
                           &key (rule-classes '(:CONGRUENCE))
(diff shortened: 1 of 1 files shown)


MattKaufmann:
Please don't anyone accept this pull request, since only J and I should commit to master.  I've emailed Pete about this.  (I could just "Close and comment" but I'm not sure if that's the best way to proceed.)  I do plan to look into these proposed changes reasonably soon, by the way.

ragerdl:
I've added "WIP" as a prefix to the subject line.  I thought this would keep the PR from being mergeable, but I guess that's just a gitlab feature.  So, nevermind.  

But, I'll leave the WIP tag since it partially implies that the PR isn't ready to be merged immediately by just anyone with push privileges.

pmanolios:
I closed the request as per the discussion with Matt.

PR #1031, opened 2019-09-05
Update documentation

airbornemihir:
Feel free to tweak the free variable example or replace it altogether afterwards. It's the simplest I could think of.


MattKaufmann:
Mihir, would you produce a tiny example illustrating what you call "indispensible"?  Offhand I think that rewriting could handle the sort of scenario you mentioned.  The tiny example would be good to include in the documentation; at the moment I, at least, find it confusing.

airbornemihir:
Well, I get that an example does a lot more than talking, but let me give talking one more try.

I gave an example in my pull request of a rule, (implies (p1 x) (p2)). If the term (p2) is to be proved, then it is insufficient to have a rewrite rule which proves (p1 A) . To be precise, it is insufficient in the sense that :monitor on the rule (implies (p1 x) (p2)) will report no suitable bindings found for x because the rewriter will not even think that A might be a suitable binding for x. The scenario changes when there is a forward-chaining rule, because (p1 A) is now going to be in the type-alist. That's a good enough reason for the monitored rule above to try A as a binding for x, succeed, and prove (p2).

I hope that helps. Perhaps it's the 0-arity stuff that is confusing...?

MattKaufmann:
Ah, I was missing the idea that (p1 A) is known from a rewrite rule, as opposed being in the context.  But I still don't understand.  Below are some sequences of events that show my attempt to understand.  Can you come up with something along the lines of one of these that succeeds when using a :forward-chaining rule but fails when instead one uses a rewrite rule?

    (defstub p1 (x) t)
    (defstub p2 () t)
    (defaxiom p1-implies-p2
      (implies (p1 x)
               (p2)))
    (thm ; succeeds
     (implies (p1 a)
              (p2)))

    ..... fresh session .....

    (defstub p1 (x) t)
    (defstub p2 () t)
    (defaxiom p1-implies-p2
      (implies (p1 x)
               (p2)))
    (defaxiom p1-holds
      (p1 x))
    (thm ; fails
     (p2))

    ..... fresh session .....

    (defstub p1 (x) t)
    (defstub p2 () t)
    (defaxiom p1-implies-p2
      (implies (p1 x)
               (p2))
      :rule-classes :forward-chaining)
    (defaxiom p1-holds
      (p1 x))
    (thm ; fails
     (p2))

    ..... fresh session .....

    (defstub p1 (x) t)
    (defstub p2 () t)
    (defaxiom p1-implies-p2
      (implies (p1 x)
               (p2)))
    ; The following fails because x is not in the trigger.
    (defaxiom p1-holds
      (p1 x)
      :rule-classes ((:forward-chaining :trigger-terms ((p2)))))

MattKaufmann:
Ah -- Mihir, is what's below the sort of thing you had in mind?  Feel
free to improve it if you had something simpler in mind.  I think this
is slightly tricky stuff, so an explicit example like this should go
into the documentation that you added.  But if you'd prefer that I add
it, then I can accept your pull request as is and add it myself -- let
me know in that case.

    (defstub p0 (x) t)
    (defstub p1 (x) t)
    (defstub p2 () t)
    (defaxiom p0-implies-p1
      (implies (p0 x)
               (p1 x))
      :rule-classes :forward-chaining)
    (defaxiom p1-implies-p2
      (implies (p1 x)
               (p2)))
    (thm ; succeeds, but not if p0-implies-p1 is a rewrite rule
     (implies (p0 x)
              (p2)))

airbornemihir:
Yep, I think your most recent example, with the forward-chaining rule 
p0-implies-p1, exactly captures the spirit of what I was saying. I'm OK 
with that example being used verbatim if you're OK with it.

MattKaufmann:
As discussed with Mihir, I'll accept this pull request and then add something like my example.

PR #1034, opened 2019-09-06
[crypto] convert ecdsa to xdoc constructors and improve

bendyarm:
[crypto] convert ecdsa to xdoc constructors and improve


ragerdl:
Does anyone know what happened here?  It looks in retrospect like leeroy automatically merged a PR made against master.  I'm skeptical that's actually what happened for two reasons:

(1) Bendyarm would have had to make the PR against acl2:master instead of acl2:testing - possible but not likely.
(2) The merging happened 10 hours after the comment occurred, suggesting the PR was alive for at least 10 hours before the merge occurred.  That could be explained by the build server being down, or maybe 10 hours being required to finish a test build before merging.

ragerdl:
It looks like #1033 contains two the three commits, which were merged in by Alessandro, so that explains two of them.  I don't see the defpointer xdco one in there though.

PR #1045, opened 2019-10-03
Small xdoc changes

airbornemihir:
I saw a couple of built-in functions, first-n-ac and make-list-ac, which seemed like they should have documentation topics even if only one line long.


MattKaufmann:
Thanks for the mods.  It seems to me that if (defpointer first-n-ac take) is to be added, then the definition of first-n-ac should be added to :doc take.

PR #1059, opened 2019-11-20
Remove TABs from XDOC strings.

acoglio:
I'm making a PR, instead of pushing directly, because this affects books written by others, in particular @MattKaufmann, @jaredcdavis, @solswords, @shigoel, and @bendyarm. If nobody objects, I'll merge this in a few days.

This is a follow-up  to an email sent by @MattKaufmann on the acl2-books mailing list.


shigoel:
Thanks a lot for removing the tabs, @acoglio! Definitely no objection from me.

solswords:
I sent my one concern about this to the mailing list, regarding the makefile patch in the building-an-ipasir-solver-library topic.

PR #1063, opened 2019-12-04
Improve the documentation of Quicklisp.

acoglio:
[I'm making a PR, instead of pushing directly, so that anybody interested in this can review.]

Mention that OpenSSL may need to be installed, as recently discussed on the ACL2
Help and ACL2 Books mailing lists.


ragerdl:
This is good.  Thanks.  Feel free to merge when comfortable.

PR #1067, opened 2019-12-15
Move some lemmas to STD

airbornemihir:
Some lemmas are added to books/std/lists/prefixp.lisp. Filesystem books are also updated.


MattKaufmann:
I'm not up on books/std/, so probably someone other than I (Matt) should accept this pull request.

airbornemihir:
Hi Matt - that is fine. I'll note that this change does cause many books to rebuild, but I can confirm that "make regression-everything" at the top level works.

acoglio:
The Std lemmas about prefixp look fine to me. Maybe we can wait another day to see if others have comments on them, before merging?

PR #1068, opened 2019-12-16
[Std/lists] Add theorem about UNION-EQUAL.

acoglio:
Since this affects Std/lists, I'm making a PR instead of pushing directly. I'll merge in a few days if there are no objections.


ragerdl:
Looks pretty innocuous to me... not that you should be waiting on my feedback.

PR #1071, opened 2019-12-18
Factor two functions out of pseudo-good-worldp.lisp.

acoglio:
@MattKaufmann Can you please quickly review and either merge or ask me to make changes?

This was discussed with Matt and J.

The functions PSEUDO-EVENT-FORMP and PSEUDO-EVENT-FORM-LISTP in
[books]/system/pseudo-good-worldp.lisp have been moved to two separate
files. The comment block just after 'EVENT-LANDMARK [GLOBAL-VALUE]' has been
moved to the new file pseudo-event-formp.lisp, and slightly tweaked, since it
essentially referred to event forms.

A :GUARD T declaration has been added to both functions, since they are no
longer under the VERIFY-GUARDS-EAGERNESS 2 setting of pseudo-good-worldp.lisp.
This way, they are still guard-verified.

Now pseudo-good-worldp.lisp includes pseudo-event-form-listp.lisp (which
obviously includes pseudo-event-formp.lisp). A comment has been added just
before this book inclusion in pseudo-good-worldp.lisp.

### DIFF
--- a/books/system/pseudo-event-form-listp.lisp
+++ b/books/system/pseudo-event-form-listp.lisp
@@ -1,0 +1,17 @@
+; Copyright (C) 2013, Regents of the University of Texas
+; Written by Matt Kaufmann and J Strother Moore
+; License: A 3-clause BSD license.  See the LICENSE file distributed with ACL2.
+; Contributions by Alessandro Coglio
+
+(in-package "ACL2")
+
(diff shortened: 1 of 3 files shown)


MattKaufmann:
Thanks @acoglio for improving the comment that you moved.  This all looks fine to me; I'll merge.

PR #1079, opened 2020-01-28
Minor changes in rp and svl books

temelmertcan:
Moved some files around and added lemmas as per some requests from Anna.


solswords:
@temelmertcan -- looks like we need DT_SB4_HC_64_64_multgen.sv to certify projects/rp-rewriter/lib/mult2/demo.lisp -- can you add it?

temelmertcan:
@solswords Ugh I didn't realize the file was missing. Added now..

PR #1082, opened 2020-02-07
Add a defpointer and corresponding content

airbornemihir:
This is for the function subsequencep.


MattKaufmann:
Mihir, please add subsequencep to the list under the comment, ``Symbols below should probably be added to *acl2-exports*.'', in books/misc/check-acl2-exports.lisp.  Otherwise I fear that that book will fail to certify.  (This is what I was trying to say on Slack.)

MattKaufmann:
Thanks, Mihir, for the update.  I just now noticed that your defpointer points to subsetp; but I think it should instead point to system-utilities.  Please either make that change or comment on why you think subsetp is correct.

PR #1090, opened 2020-03-13
Without subumption

TheBeaNerd:
Add without-subsumption macro; minor documentation updates.


ragerdl:
Thanks for including a license @TheBeaNerd .  There appear to be some edits to acl2 system files, so I'll let @MattKaufmann handle this one.

MattKaufmann:
Normally, only files under books/ should be changed by users other than me (or J).  But I'll accept this pull request (the files in question look OK to me) and then adjust the script bin/purity.sh accordingly.

PR #1095, opened 2020-04-04
enhanced quantification support, functional instantiation TODO

TheBeaNerd:


MattKaufmann:
Thanks (@TheBeaNerd) for adding that clearly-explained to-do item!  As usual, anyone on the acl2-devel list is welcome to note in the item that he or she is claiming it, and then ultimately follow the procedure in :doc developers-guide-contributing (and more generally, as discussed in :doc developers-guide).

PR #1098, opened 2020-04-07
Add theorems to STD prefixp book

airbornemihir:
Filesystem books are also updated.


ericwhitmansmith:
Regarding nthcdr-when-prefixp, I suggest instead proving a stronger iff rule for nthcdr, like this:

(defthm nthcdr-iff
  (iff (nthcdr n x)
       (if (< (nfix n) (len x))
           t
         (if (equal (nfix n) (len x))
             ;; If we know true-listp, this simplifies to nil and get merged
             ;; with the nil branch below.
             (not (true-listp x))
           nil)))
  :hints (("Goal" :in-theory (enable nthcdr))))

and then adapting nthcdr-when-prefixp to talk about comparing the lengths of two lists when one is a prefix of the other (and also changing the name since the new rule won't mention nthcdr).

Thanks for this contribution.  It inspired me to formulate the above rule about nthcdr, which I'll add to my nthcdr library shortly.

airbornemihir:
@ericwhitmansmith You're welcome. What would you say to making the following change to the existing len-when-prefixp theorem, subject to a check whether it breaks anyone's books?
(defthm len-when-prefixp
    (implies (prefixp x y) (and
                            (equal (< (len y) (len x))
                                   nil)
                            (equal (< (len x) (len y))
                                   (not (list-equiv x y)))))
    :rule-classes ((:rewrite)
                   (:linear :corollary (implies (prefixp x y)
                                                (<= (len x) (len y)))))
    :hints(("Goal" :in-theory (enable (:induction prefixp)))))

ericwhitmansmith:
Seems ok (I'd make many stylistic changes, but this is not my library).  I'd also probably put a backchain limit of 0 on that rule, to avoid introducing prefixp reasoning out of nowhere when comparing two calls of len, but that's also kind of just personal preference (in that case I might add an unlimited variant of the rule but leave it disabled).

airbornemihir:
I made some changes to backchaining as you suggested. The books seem to 
be certifying OK, but it might be best for me to wait until the 
regression can complete.

airbornemihir:
@ericwhitmansmith How important is the backchain limit to you? Setting it to 0 causes my own books to fail certification, and I'm not sure there's a clean solution for the proofs that fail.

ericwhitmansmith:
It's not very important to me (I have my own book about prefixp).  Failures like the ones you mention can usually be fixed by selectively enabling the unlimited version of the rule, if you add it.  Or you could do nothing and wait until someone complains about the rule slowing down their proofs.  :-)

ragerdl:
FWIW, I think leaving the default backchain limit (which I think is unlimited) is a nice and general approach for most users.  Having a limit of 0 can be useful, but if proofs are going too slowly, users will be able to figure out what's going on with accumulated-persistence.

OTOH, I'd like to think that there's a way of interacting with ACL2 that is more proof-step oriented -- having a backchain limit of 0 is more along those lines.  Still, I think the default is the more general and more useful approach.

ericwhitmansmith:
It may depend on what other rules are around about prefixp, since the unlimited rule can introduce prefixp reasoning out of nowhere (in a proof that doesn't have anything to do with prefixp, which most proofs won't).  I've seen similar problems with other rules from std.  But, like I said, I don't feel strongly.

solswords:
I would also prefer that rules that introduce reasoning about something more "specific" (like prefixp) from something more "general" (like equality of len) be at least backchain-limited and preferably disabled by default.  I'm sure there are many rules already in std that don't satisfy this vague criterion.  I'd prefer not to introduce new ones, though, since then if we decide later that they're bothersome and want to limit/disable them, we cause lots of things to break.

ragerdl:
Good insight guys, thanks.

airbornemihir:
I guess my concern would be that a strict backchain-limit would make it 
harder to "debug" a failing proof by means of brr. Most likely, one 
would add lemmas that establish prefixp for whatever subgoals are 
failing, as directed by brr - but the main proof would still fail 
because of the backchain-limit making the new lemmas ineffective.

The above argument applies to all cases where a backchaining is limited, 
I guess, so it might be overly broad. The other thing, though, is that 
I'm having a hard time thinking of when it might be needed to compare 
two lengths outside of a prefixp-related context. Reasoning about strict 
subsets and strict supersets, maybe - are there books which do that?

airbornemihir:
There's also the fact that this book, books/std/lists/prefixp.lisp, does 
not define prefixp and so it won't be included everywhere prefixp is 
used. This book is specifically for reasoning about prefixp.

ericwhitmansmith:
I think brr will tell you when a backchain limit is reached.  What I do is name the main rule something like foo-when-blah, keep it disabled, and also add foo-when-blah-cheap (with a backchain limit of 0 and enabled).  If foo-when-blah-cheap fails to do the trick, the user can enable foo-when-blah.

ericwhitmansmith:
You make a good point about the prefixp book not always being included.  The best practice would probably be to only include it locally, when needed.  But I do see it being included non-locally by at least some books about strings, by std/lists/sublistp, and of course by std/lists/top, which is included many places (including some fundamental acl2(s) books!).

ragerdl:
@ericwhitmansmith Do you have a macro for defining both the cheap and expensive version of the rule, do you duplicate the code, or do you do something else?  In this case, I'm kind of trying to avoid both the first and the second and wondering if there's some ACL2 syntax for making both (maybe something to do with :rule-classes?).

@airbornemihir I think the fastest and still conservative way to proceed w.r.t. the PR is to make it a disabled rule with whatever backchain limit you want, and then enable it whatever filesystem book includes this std library.  That would still allow the rule to be added (for future discussion, etc), without having to settle this issue right now.

Personally, I'm of the view that ideally std/top (which hopefully includes std/lists/top) would be includable by everyone, enabling significantly more proofs to automatically go through, without being terribly expensive in execution/useless proof paths.  Those two goals are in tension, but that's what we're trying to achieve.  Note, of course, experts have their own preferences and we will never all include std/top.

acoglio:
I vote for leaving rules of that kind disabled by default.

acoglio:
I mean the ones that introduce "new concepts" by backward chaining.

airbornemihir:
Thanks everyone for weighing in. I guess I do understand Eric's point 
about dual cheap/expensive rules, and I have seen that pattern in other 
books.

I guess my reluctance to create two rewrite rules is partly about the 
brr issue I mentioned above (which Eric addressed) and partly because 
I'm not particularly fond of having two rewrite rules which rewrite the 
same target under the same hypotheses. This is because of a problem 
described in :doc USING-ENABLED-RULES, where one might need to 
instantiate a lemma and not get the intended simplification because the 
lemma was left enabled and caused its own lemma-instance to get 
rewritten into vacuity. This problem can also arise when there are two 
enabled rules that do the same thing - and this time there won't be a 
warning of the kind described in that :doc topic.

I recognise that this problem can be avoided by being careful about only 
enabling one rule if the other is disabled, and disabling both together 
if instantiation is needed, which is probably what most people do. 
However, I guess it becomes a stylistic choice for me not to want to pay 
that much attention to theories - so I'm opting to keep the new rewrite 
rule in my own book, and revert len-when-prefixp to the way it was 
before this change. I'm currently testing this latest change and I'll 
merge the pull request afterwards unless someone objects.

MattKaufmann:
Regarding:

> I recognise that this problem can be avoided by being careful about only
> enabling one rule if the other is disabled....

This can be enforced; see :doc theory-invariant.

PR #1099, opened 2020-04-08
Workshop 2020 supporting materials -- Quadratic Field Extensions

rag42:
Supporting materials for the workshop. Note that number-field.lisp takes forever to certify, so it maybe shouldn't be in "make all".

### DIFF
--- a/books/GNUmakefile
+++ b/books/GNUmakefile
@@ -548,6 +548,7 @@ ADDED_BOOKS := \
   workshops/2011/krug-et-al/support/MinVisor/va-to-pa-thm.cert \
   workshops/2011/krug-et-al/support/MinVisor/setup-nested-page-tables.cert \
   workshops/2013/hardin-hardin/support/APSP.cert \
+  workshops/2020/gamboa-cowles-gamboa/number-field.cert \
   doc/top-slow.cert
 
 # The following could well be considered as slow books, too.
--- a/books/workshops/2020/README
+++ b/books/workshops/2020/README
@@ -20,6 +20,7 @@ coglio-westfold/
 
 Ruben Gamboa, John Cowles and Woodrow Gamboa
 Quadratic Extensions in ACL2
+gamboa-cowles-gamboa/
 
 ------------------------------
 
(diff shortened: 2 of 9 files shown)


MattKaufmann:
I'd like us to have a bit more confidence that the testing run on
leeroy will succeed, before accepting this pull request.  My main
concern is that if any books require ACL2(r), then there will be a
need to add the following to at least some books.  There are lots of
examples in the community books, e.g., in
nonstd/workshops/1999/analysis/exercise1.lisp:

; cert_param: (uses-acl2r)

Also, regarding:

> Note that number-field.lisp takes forever to certify, so it maybe
> shouldn't be in "make all".

So please add that book's .cert file to ADDED_BOOKS in
books/GNUmakefile -- just search for ADDED_BOOKS and I think it will
be clear how to do that.

Even "make everything" shouldn't take forever though.  How long as
number-field.lisp taken to certify?

airbornemihir:
Running the regression on this pull request branch causes make to produce a number of warning messages about circular dependencies. They are kinda lengthy but they start like this:

```
`Scanning for dependencies...
Dependency loop on workshops/2020/gamboa-cowles-gamboa/floor1-non-R.cert!
Current stack:
   workshops/2020/gamboa-cowles-gamboa/floor1-non-R.cert
Circular dependency found in src_deps of workshops/2020/gamboa-cowles-gamboa/floor1-non-R.lisp
Warning: LD event in book context in workshops/2020/gamboa-cowles-gamboa/floor1-non-R.lisp:
ld floor1-non-R.lisp
Oops, trying to go into /u/mihir/git/acl2/books/workshops/2020/gamboa-cowles-gamboa/arithmetic-5/
bad path in (include-book "arithmetic-5/top")
Bad path in (include-book "arithmetic-5/top") in workshops/2020/gamboa-cowles-gamboa/floor1-non-R.lisp
````

ericwhitmansmith:
Mihir, good catch!  It looks like there are some block comments that should be changed to line comments (using semicolons) to avoid fooling the dependency scanner.  (Actually, perhaps they could just be deleted.  They seem to just explain how to do standard things like include books and use gag mode.  Also, they mention an old ACL2 version, 8.1.)

PR #1102, opened 2020-04-09
Change regexes for ETAGS invocation so that it gets a few kinds of de…

solswords:
…finitions it was previously missing.

The previous regular expressions used for the etags invocation included:
```
def[^tc]\|defttag\|deftheory
```
which seemed like it was intended to omit defthm and defconst events from producing tags, but also had the side effect of omitting several other things that start with "defc" or "deft", such as deftypes, deftagsum, etc.  I have changed this so that it gets everything beginning with "def" except if it begins with "defthm".  Unless I'm missing something it seems harmless to include defconsts.

I also changed from `--ignore-case-regex=//` to `--regex=//i` since the former doesn't seem to be documented whereas the latter is.  Additionally, I split out the regex for `memoize` invocations since they're a bit different than definitions.  I'm not sure if we should really want memoize to generate tags, though.

If this doesn't bother anybody I'll merge it in a day or four.

### DIFF
--- a/books/GNUmakefile
+++ b/books/GNUmakefile
@@ -1623,10 +1623,11 @@ all: centaur/getopt/demo2-test.ok
 # These regular expressions may include things more appropriate for
 # raw Lisp.  May want to remove some of these or extend them to
 # support new types of definitions.
-ETAGS = etags --language=none \
- --ignore-case-regex='/ *set[ \t]?\([^ =]*\)[ \n]*=/\1/' \
- --ignore-case-regex='/(set[^ \t]*[ \t]?\([^ ]*\)[ \t]*/\2/' \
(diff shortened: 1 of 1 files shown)


MattKaufmann:
@solswords I'm hoping that sometime this weekend I'll freeze my local ACL2 distribution for the 8.3 release.  If you want me to wait till this books/GNUmakefile change is merged, let me know; or I could make a note to merge it myself just before I freeze (if it isn't already merged), if you ask me to do that.

solswords:
@MattKaufmann I think I'll just merge it.  I don't anticipate anyone having too much trouble and it won't break regressions or anything.  Not sure if we want memoizations tagged or not but it's easy to change later.

PR #1104, opened 2020-04-11
Hardin 2020 Workshop Materials

david-s-hardin:
A basic example for my talk at the 2020 ACL2 Workshop.

ACL2 books (stk and stk-thms) should certify with standard cert.pl.

The .c and .h files are only "for show" -- the stk.lisp code is generated from them 
using Russinoff's RAC tools, but those tools aren't normally built, and there's no 
reason for the average user to concern themselves with recapitulating the 
C -> ACL2 translation.

Please let me know if there are any issues with these files.

-- David Hardin


david-s-hardin:
Didn't change "Master" to "Testing" branch -- sorry about that.

PR #1105, opened 2020-04-11
Hardin 2020 ACL2 Workshop Materials (Try #2)

david-s-hardin:
A basic example for my talk at the 2020 ACL2 Workshop.

ACL2 books (stk and stk-thms) should certify with standard cert.pl.

The .c and .h files are only "for show" -- the stk.lisp code is generated from them 
using Russinoff's RAC tools, but those tools aren't normally built, and there's no 
reason for the average user to concern themselves with recapitulating the 
C -> ACL2 translation.

Please let me know if there are any issues with these files.

-- David Hardin

### DIFF
--- a/books/workshops/2020/hardin/README.md
+++ b/books/workshops/2020/hardin/README.md
@@ -1,0 +1,5 @@
+Simple stack data type implemented using Restricted Algorithmic (RAC), and automatically 
+translated to ACL2, at which point functional correctness proofs can be performed.
+
+Creation of the ACL2 code from the RAC sources requires that RAC tools are built and installed.
+See projects/rac in the ACL2 books directory for more information on RAC.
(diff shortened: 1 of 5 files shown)


MattKaufmann:
Oops; after accepting #1105 I realized that it was directly into master, too (like #1104).  However, I certified the .lisp books successfully, so no big deal.

However, it now occurs to me that the stk.h will likely get cleaned by books/build/clean.pl.  Ah, but that didn't happen when I just now tried it.  So I think all is well with that, but if someone knows of a looming problem with having that .h file, please reply.

MattKaufmann:
Oh, please disregard my preceding comment -- I realize now that .h files are probably not a problem because clean.pl leaves them alone unless there is a (GCL-generated) gazonk file....

MattKaufmann:
Sorry about the noise; I was confused.  I see now that yes, stk.h presents a problem: it is deleted by cleaning (e.g., by books/build/clean.pl or various "clean" targets of "make").  I am creating a sources/ subdirectory with a cert_pl_exclude file, and moved stk.h there (also stk.cpp), which solves that problem.

PR #1107, opened 2020-04-24
Add congruence rules and update a documentation topic

airbornemihir:
assoc, strip-cars, remove-assoc and put-assoc are generally associated with an alist context, so the congruence rules on list-equiv may seem surprising. However, remove-assoc always returns a true list, even when it doesn't remove anything, and so true-list-fix sometimes appears unexpectedly in an argument to the above four functions during alist-related proofs. Let me know if you'd like this explanation to be a comment in the file.


MattKaufmann:
The changes include a change to books/system/doc/acl2-doc.lisp saying
that :pr does not print induction rules.  The following seems to be a
counterexample.  The change also says that :pr does not print
refinement, which I believe is correct though I didn't test it
(perhaps Mihir would be willing to do so).  I don't mind making the
fix for induction rules after the change gets into master, if that's
easier.

    ACL2 !>(defthm foo
             t
             :rule-classes ((:induction :pattern (floor i 2)
                                        :condition (and (integerp i) (>= i 0))
                                        :scheme (integer-length i))))

    Q.E.D.

    Summary
    Form:  ( DEFTHM FOO ...)
    Rules: NIL
    Time:  0.00 seconds (prove: 0.00, print: 0.00, other: 0.00)
     FOO
    ACL2 !>:pr foo

    Rune:         (:INDUCTION FOO)
    Enabled:      T
    Pattern:      (FLOOR I 2)
    Condition:    (AND (INTEGERP I) (<= 0 I))
    Scheme:       (INTEGER-LENGTH I)
    ACL2 !>

airbornemihir:
@MattKaufmann thanks for the counterexample; I've fixed it in testing so we don't have master pointing to an inaccurate statement.

ericwhitmansmith:
FWIW, I think the comment about why true-list-fix appears, and thus why these rules are needed, is just the kind of thing that would be helpful to see in a comment in the file  Thanks for the explanation!

MattKaufmann:
Perhaps someone familiar with books/std/lists/equiv.lisp will accept this pull request.  (I'm not comfortable accepting pull requests about books/std/, because I don't know enough about that.)

PR #1108, opened 2020-04-29
Add an example to a documentation topic

airbornemihir:


MattKaufmann:
Thanks, @airbornemihir; this looks good.  In fact I think it's a better example than the one already there.  I think it would make sense to switch the two, not only because your example is more straightforward but also because you mention "without the verbosity of a :use hint" and that's important -- maybe it should even be said well before either example.  I'd be happy to accept the pull request as is and make that change later myself, or you can commit the change first, whichever you prefer; just let me know.

By the way, in your text above the example code would best be surrounded by @('..'): thus, @(':restrict'), @('y'), and @(':use').

PR #1109, opened 2020-05-03
Small xdoc updates

airbornemihir:


airbornemihir:
Since books/std/lists/index-of.lisp is a Centaur book, perhaps @shigoel or @solswords could merge this.

PR #1111, opened 2020-05-08
Add stuff to xdoc and std books

airbornemihir:


airbornemihir:
The lemma I added, no-duplicatesp-of-set-difference-equal, could plausibly go in books/std/sets.lisp, since that's where some lemmas relating to set-difference live. Let me know if I should change that (which will require creating a new section for no-duplicatesp-related lemmas.)

PR #1118, opened 2020-05-18
WIP: Attempting to undo Myall's commits

ragerdl:
This is not rewriting history.  All of the code is still there -- you just have to use the write commit hash to get at it.


ragerdl:
Going to hold off on this for now, as it seems Cuong's fixed the other books.

PR #1122, opened 2020-05-19
Generalise and rename an std/lists/append lemma

airbornemihir:
Thanks are owed to @ericwhitmansmith for a Slack discussion in March
about the new name. I had to make a change to @myall86's books to make
the regression pass, so he might have something to say about this pull
request.

### DIFF
--- a/books/projects/async/index.lisp
+++ b/books/projects/async/index.lisp
@@ -13,6 +13,13 @@
 (include-book "utils")
 
 (include-book "std/strings/decimal" :dir :system)
+(defthm acl2::equal-of-appends-when-true-listps
+    (implies (and (true-listp x1)
+                  (true-listp x2))
+             (equal (equal (append x1 y)
+                           (append x2 y))
+                    (equal x1 x2))))
+(in-theory (disable acl2::equal-of-append-and-append-same-arg2))
 (include-book "std/strings/istrprefixp" :dir :system)
 
 ;; ======================================================================
@@ -373,5 +380,3 @@
                               cdr-nthcdr))))
 
 (in-theory (disable sis))
-
-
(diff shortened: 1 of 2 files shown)


myall86:
@airbornemihir I'm fine with your change. Thanks.

ericwhitmansmith:
The new rule looks good to me.  It might be interesting to see if the failure of the async book using the new rule could reveal the need for any additional good rules, but that could always be done later.

airbornemihir:
On 5/19/20 5:54 PM, Eric W. Smith wrote:
> It might be interesting to see if the failure of the async book using 
> the new rule could reveal the need for any additional good rules, but 
> that could always be done later.
I agree, that's generally a useful exercise. However, IIRC there were 12 
async books which were failing on account of the new rule, so it might 
turn out to be non-trivial.

ericwhitmansmith:
> I agree, that's generally a useful exercise. However, IIRC there were 12 async books which were failing on account of the new rule, so it might turn out to be non-trivial.

Ouch.  Sounds like we should skip it for now then.

airbornemihir:
@ericwhitmansmith I actually tried once again examining the failures in the async books after this pull request, and they seem kinda like an almost-doable brainteaser (maddening, haha.) If you have the time and the bandwidth for something like this, it might be nice if you could try looking at those failures and figuring out what additional rules might get the async books to certify without changes. To be precise, you would have to reverse two things, my addition of the rule acl2::equal-of-appends-when-true-listps and my disabling of the rule acl2::equal-of-append-and-append-same-arg2 - you can see the diff for this pull request for details. It's perfectly fine if you don't have the time, but please let me know if you get somewhere.

PR #1124, opened 2020-05-20
Fixing the soundness bug that unary-/ can be interpreted as integer division in Smtlink

pennyan:
Andrew Walter and Pete Manolios have brought to our attention of a soundness bug in Smtlink. It is recommended that theorems depending on Smtlink (especially ones using the macro /)should be recertified using this new version of Smtlink. Sorry for the trouble.

The bug is from how I handled the translation of unary-/. For example, when encountering (unary-/ x) where x is of integerp, Smtlink will translate it into 1.0/x. Z3 does not follow the Python convention but uses integer division in this expression.

In file examples.lisp, the diff is an example showing the soundness problem of this translation. Below theorem gets proved because of this bug.
```
(defthm smt-not-integer-division
   (implies (and (integerp x)
                 (integerp y)
                 (integerp z)
                 (>= x 10)
                 (= y (* (unary-/ x) z)))
            (= y 0))
   :hints (("goal" :smtlink nil))
   :rule-classes nil)
```

In this example, unary-/ is treated as integer division, and because x >= 10, Z3 concludes (unary-/ x) is 0, which makes y 0. Thanks to Andrew and Pete for providing this example.

The fix is to cast the variable x to a real in the expression 1.0/x. The fix can be found in the diff in file ACL2_to_Z3.py.

### DIFF
--- a/books/projects/smtlink/examples/examples.lisp
+++ b/books/projects/smtlink/examples/examples.lisp
@@ -751,3 +751,21 @@ Possible counter-example found: ((M2 (SOME 0)) (M1 (SOME 0)))
   :hints(("Goal"
           :smtlink (:abstract (abstract-p))))
   :rule-classes nil)
+
+;; Thanks to Andrew Walter and Pete Manolios for providing this bug example.
+;; In this example, it used to be that unary-/ is translated to be integer
+;; division in Z3. Since unary-/ is interpreted as integer division and x >=
+;; 10, 1/x is 0, which makes y 0. We fixed this problem by casting the input of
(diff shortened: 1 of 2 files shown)


pennyan:
Maintainers, could you please let me know if I should repost this pull request to any mailing list? I suspect this fix deserves an announcement?

MattKaufmann:
Thanks, Yan; I've merged.  I think it's fine if you want to announce the fix to the acl2-help or acl2 list (maybe others have a stronger opinion one way or the other on whether you *should* do that).  If you do, then I suggest that you wait until the fix is in master; and you could point out that they will need to get the github version in order to get the fix (since ACL2 8.3 won't change).

PR #1125, opened 2020-05-23
community-books-acl2-2020.tex extended and some other changes

temelmertcan:
I added descriptions for my new books to "community-books-acl2-2020.tex"
I also made some changes in projects/rp-rewriter that can speed up the certification of some books.

So I ran a regression ~30 hours ago, and it was working fine. I realized with the new changes that happened in the meantime, ACL2 has to be rebuilt. I have started running another regression now but I don't want to wait for it in case somebody else has to change community-books-acl2-2020.tex. Please deny this request or wait for a few hours if you think it is better.


MattKaufmann:
I think it would suffice to certify your modified books with the
up-to-date ACL2.  You could for example do something like the
following.

    cd books/projects/rp-rewriter/
    ../../build/cert.pl --acl2 <path_to_acl2> -j 4 *.lisp

temelmertcan:
Some of my modified files use Centaur's SV and VL books, and they take a while. I will update here when it finishes.

temelmertcan:
All my books certify.

PR #1126, opened 2020-05-24
[Std/util] Extend DEFLIST with :PREPWORK.

acoglio:
I'm making a PR since this affects STD::DEFLIST. But it shouldn't cause any issues for existing uses.

I'll merge this in a few days if there are no objections.

Now DEFLIST supports the :PREPWORK option, like other macros in Std/util (and in
FTY).


solswords:
Looks good to me

PR #1128, opened 2020-05-30
defrule: suppress nil keyword arguments to defthm

kini:
I found this commit lying around on an old branch and figured I'd submit a PR. I ran a regression successfully except that I ran out of "immobile space" in SBCL when building the manual (I'm pretty sure that's unrelated to this change, though).

----

Before this commit, `defrule` would always add keyword arguments `:hints`, `:otf-flg`, and `:instructions` to the generated `defthm` form, even when their values were `nil` (i.e. equal to their default values).  Now it doesn't add those arguments unless they're necessary, which makes `:pe` output a little less verbose.

Before:

```
ACL2 !>:trans1 (defrule foo bar)
 (DEFTHM FOO BAR
         :HINTS NIL
         :OTF-FLG NIL
         :INSTRUCTIONS NIL)
```

After:

```
ACL2 !>:trans1 (defrule foo bar)
 (DEFTHM FOO BAR)
```

### DIFF
--- a/books/build/features.sh
+++ b/books/build/features.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # ACL2 Community Books build system
 # Copyright (C) 2019 Centaur Technology
--- a/books/std/util/defrule.lisp
+++ b/books/std/util/defrule.lisp
@@ -211,12 +211,12 @@ generated using @(see defthmd) instead of @(see defthm).</p>")
         (t
          (find-goal-entry-in-user-hints (cdr user-hints)))))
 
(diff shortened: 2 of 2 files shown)


ragerdl:
This seems like a small improvement to me.  What about @acoglio or @solswords ?

acoglio:
It looks like `defthm` treats `:hints nil` the same as no `:hints`(*), so the change looks good to me. Thanks to @kini for doing this.

(*) I had a slight doubt about that with respect to the treatment of default hints.

kini:
Thanks for the feedback. I'll merge now, then.

PR #1129, opened 2020-06-01
[svl] some no longer existing rules are removed from svl-run-to-svex-alist.lisp

temelmertcan:
Thanks to Anna Slobodova for pointing out this error. I only removed some old rule names from a constant  in books/centaur/svl/svl-run-to-svex-alist.lisp


temelmertcan:
I requested the merge to testing-centaur so that Anna can access the change quickly.

PR #1130, opened 2020-06-03
Reduce deps of leftist trees

ericwhitmansmith:
This pull request deals with an upcoming name conflict on 'perm'.  The
leftist-trees project brings in a definition of perm from books/sorting/.
However, leftist-trees doesn't need to export anything involving perm.
Furthermore, this version of perm comes into the manual only via leftist-trees.

The changes are in two parts:

1. In books/sorting/, split two functions, how-many and orderedp, into separate
small books containing only these definitions.

2. Change leftist-trees to include these new, smaller books (non-locally) and
to include a book about perm only locally.

The result is that the version of perm from books/sorting/ is not in the manual
and so doesn't conflict with the one in books/kestrel/lists-light/perm (which
we use fairly extensively, including in material we hope to add to the manual
soon).

These changes could be considered beneficial in their own right, since they
make leftist-trees more usable (less likely to conflict with a user's books).

### DIFF
--- a/books/projects/leftist-trees/leftist-tree-sort.lisp
+++ b/books/projects/leftist-trees/leftist-tree-sort.lisp
@@ -25,9 +25,11 @@
 
 (in-package "ACL2")
 
-(include-book "sorting/perm" :dir :system)
-(include-book "sorting/ordered-perms" :dir :system)
-(include-book "sorting/convert-perm-to-how-many" :dir :system)
+;;(include-book "sorting/perm" :dir :system)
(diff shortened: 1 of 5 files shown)


benjaminselfridge:
I'm good with this!

airbornemihir:
Good to hear from you, Ben!

kini:
LGTM as well :) Merging.

ericwhitmansmith:
For the record, J also told me by email that he's fine with this change.

PR #1131, opened 2020-06-03
Fix a typo and add a defpointer form

airbornemihir:


airbornemihir:
Thanks, Matt, for noticing. I have now run a regression, and the exit status is 0.

MattKaufmann:
@airbornemihir Thanks for making the change.  Actually, I think I was mistaken (sorry!): such an addition to *acl2-exports-exclusions* needs to be made for defined functions, macros, and constants that are documented by not in *acl2-exports* -- but too-many-ifs isn't actually defined.  I'll go ahead and accept the pull request and probably delete that addition to acl2-exports-exclusions* later.

PR #1136, opened 2020-06-07
Remove a redundant lemma

airbornemihir:
The removed lemma, set-equiv-implies-iff, is completely subsumed by set-equiv-implies-iff-member-2 from the same file, as demonstrated here:
```
(make-event `(thm ,(formula 'set-equiv-implies-iff t (w state)) :hints
                   (("goal" :in-theory (union-theories '(set-equiv-implies-iff-member-2)
                                                        (theory 'minimal-theory))))))
```

### DIFF
--- a/books/std/lists/list-defuns.lisp
+++ b/books/std/lists/list-defuns.lisp
@@ -309,38 +309,16 @@
 
 (encapsulate
   ()
-  (local (defun simpler-take-induction (n xs)
-           (if (zp n)
-               nil
-             (cons (car xs)
-                   (simpler-take-induction (1- n) (cdr xs))))))
-
-  (local (defthm equivalence-lemma
-           (implies (true-listp acc)
-                    (equal (first-n-ac n xs acc)
-                           (revappend acc (simpler-take-induction n xs))))))
-
-  (local (defthm take-redefinition
-           (equal (take n x)
-                  (if (zp n)
-                      nil
-                    (cons (car x)
-                          (take (1- n) (cdr x)))))
-           :rule-classes ((:definition :controller-alist ((TAKE T NIL))))))
-
-  (local (in-theory (disable take)))
 
   (local (defthm take-when-atom
            (implies (atom x)
                     (equal (take n x)
                            (replicate n nil)))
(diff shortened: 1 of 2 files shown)


airbornemihir:
Adding some more stuff to the PR, about cleaning up the last remnants of take-redefinition which Alessandro Coglio removed in #995 .

solswords:
The fact that the body of set-equiv-implies-iff can be proved without induction doesn't mean that the rule won't work during backchaining when other rules don't.  In particular, I think if (set-equiv x y) can be rewritten to T but not because it is known in the type-alist and not because x rewrites to y under set-equiv (or vice versa), then this rule might work in some cases where (member a x) and (member a y) wouldn't normalize to the same term under iff.  That said, I think it's fairly rare for iff terms to appear during simplification, so if everything in the community books still certifies without that rule then I'm OK with getting rid of it.

airbornemihir:
Hi Sol,

Thanks for looking at this. I guess I should clarify, 
set-equiv-implies-iff and set-equiv-implies-iff-member-2 have basically 
the same formula with different variable names. The latter is more 
powerful because it is a congruence rule, and the regression passes when 
the former is removed. My motivation to remove this is that sometimes 
I'm using find-lemmas to look for a lemma I need and the more output 
there is, the longer it takes...

Thanks,
Mihir.

solswords:
My point was that the congruence rule and rewrite rule do very different things.  But if the regression passes I'm OK with getting rid of it.

airbornemihir:
Hi Sol,

OK, perhaps I didn't get your point before. When you mention the 
situation where "(set-equiv x y) can be rewritten to T but not because 
it is known in the type-alist and not because x rewrites to y under 
set-equiv (or vice versa)" I see a possible example where there's a 
carefully constructed chain of set-equiv rules, but also set-equiv is 
enabled and iff is disabled. There would be a mess of subsetp terms and 
iff terms would stick around, making the application of this rule 
possible. I'm not sure I can quickly make a concrete example; but is 
that what you're thinking about?

Cheers,
Mihir.

PR #1137, opened 2020-06-07
Update satlink_lib.pl to use drat-trim

ragerdl:
The `centaur/satlink/solvers/test-glucose-cert.lisp` books builds with this change.

It seems like the `drup-trim` reference is no longer best practices.  My guess is
that this change is already intended, but just hasn't been done yet.


ragerdl:
Gonna go ahead and merge.  It sounds like most folks aren't even triggering the book that calls drup-trim, because they're unlikely to have ipasir configured.  And those that do have ipasir configured likely are aware of this change.

PR #1138, opened 2020-06-09
Remove `std/util/defval-tests` book

mister-walter:
Based on a discussion with @ragerdl and @acoglio on the ACL2 Slack, remove `std/util/defval-tests`.

The XDOC documentation in `std/util/defval` had a dependency on `std/util/defval-tests` and several books (perhaps most notably `std/strings/url-encode`, which is part of `std/strings/top`) include `std/util/defval` without including `std/util/defval-tests`. This meant that `xdoc::save` would fail if any book had been included which transitively included  `std/strings/top`, or any of the other books which transitively included `std/util/defval` but not `std/util/defval-tests`.

I removed the `std/util/defval-tests` book and updated the `std/util/defval` documentation to look similar to how it did previously while no longer referencing `std/util/defval-tests`.

There are two places where `std/util/defval-tests` are referenced that I did not modify: `centaur/fgl/top` (where the `include-book` is commented out), and `kestrel/auto-termination/td-cands` (where the `fib` function defined in `defval-tests` is included in an automatically generated database of termination schemes). I can modify these if desired, but didn't want to touch other books without discussion.


mister-walter:
To reproduce the original issue, simply certify a book containing the following code (be sure to allow all ttags during certification)
```
(in-package "ACL2")
(include-book "std/util/defval" :dir :system) ;; or std/strings/top, or any other book 
                                              ;; that transitively includes defval
                                              ;; without defval-tests
;;(include-book "std/util/defval-tests" :dir :system) 
                                              ;; certification succeeds when this line
                                              ;; is uncommented
(include-book "xdoc/save" :dir :system)
(xdoc::save "./doc" :error t :redef-okp t)
```

acoglio:
@mister-walter Thanks for making this PR.

I see that the reference to `defeval-tests` in `kestrel/auto-termination/td-cands.lisp` is not an explicit book inclusion, but is referenced in a constant. Perhaps @MattKaufmann, who wrote that file, can have a look to see if this file needs adjustment? I'm assuming that the regression all works with this PR.

It looks like the Centaur file only references `defeval-test` in a comment, but @solswords may want to remove it.

MattKaufmann:
@acoglio @mister-walter The file `kestrel/auto-termination/td-cands.lisp` is automatically generated, and I think the issue will resolve itself when it is next generated.  If there is a problem with the regression, then I think it would be fine for now to comment out manually, or delete, the form, `(STD::FIB 1 :SYSTEM . "std/util/defval-tests")`.

mister-walter:
Regression test is still running on my machine, but `kestrel/auto-termination/top` certifies with these changes.

ragerdl:
Not that I've thought carefully about this, but I'd have thought that we'd want to keep the test around, but not make a reference to it in `defval`.  *If that doesn't work*, then I'm guessing that once @mister-walter confirms that the regression completed, we should merge this.

mister-walter:
Apologies for the delay! I'll update the PR to keep the test but remove the reference to it in the `defval` docs. One quick question: Should `std/util/defval-tests` be moved to `std/util/tests/` or should it remain inside the `std/util` directory? And should it be included in `std/util/top` if it is left inside the `std/util` directory?

I recall the regression test failing on my machine due to some problem interfacing with one of the external tools; I'll rebase the PR and rerun the regression test to remind myself what failed.

acoglio:
@mister-walter Thank you for working on this.

I would probably move those tests under the `tests` subdirectory just or consistency within `std/util/`. 

I would not include tests in `top.lisp`.

acoglio:
@mister-walter @ragerdl @MattKaufmann I wonder where we stand on this. Let me know if/how I can help.

MattKaufmann:
@acoglio I haven't been following this issue other than to post the comment I made about kestrel/auto-termination/td-cands.lisp.  I'm willing to dive in if asked.

mister-walter:
Apologies for the delay - I ended up having problems running a regression test and this fell off my stack. I'll rebase and try to run a regression test later today.

One comment - there already exists a file `std/util/tests/defval.lisp` which contains some tests for defval. Should I move the test from `std/util/defval-tests.lisp` into that file, or move `std/util/defval-tests.lisp` to `std/util/tests/defval-2.lisp` or something like that?

mister-walter:
Oops I screwed something up when rebasing, should be good now. I moved the test from `std/util/defval-tests` into `std/util/tests/defval` alongside the other defval tests, and removed the old `std/util/defval-tests.lisp` file.

acoglio:
Thanks @mister-walter. Adding the test to the existing file, as you've done, looks good to me.

I see that the PR is made against master. Would you mind making it against testing instead? Sorry for the trouble, but this is the typical protocol that we follow.

mister-walter:
Done, sorry about that!

PR #1139, opened 2020-06-16
Add STRINGS to the parents of :DOC CHAR.

acoglio:
I'm making a PR since this affects the ACL2 documentation. @MattKaufmann can merge if he approves the change.

Since CHAR takes a string as argument and returns a character, it seems
appropriate for its documentation topic to appear under both :DOC STRINGS and
:DOC CHARACTERS, similarly to COERCE.


MattKaufmann:
@acoglio That change looks good.  Perhaps I'd leave CHARACTERS as the first parent of CHAR and make STRINGS the second parent, but maybe not; it probably doesn't matter.  I'll merge.

acoglio:
@MattKaufmann Thanks. Fine with me to switch the parents order. But somehow I think of `char` as being more of a "string operation" than a "character operation"; that's why I put the `strings` parent first.

PR #1150, opened 2020-07-15
Add type-prescription corollaries to an existing rule

airbornemihir:
This pull request fixes a failure in the "everything" regression, but also changes an STD book. Ideally these two things would be separate from each other, I know, but the remote machines I usually use for rebuilding books appear to be heavily contended by other users and redoing these changes separately will take a while.


MattKaufmann:
I'm accepting the pull request so that master (currently broken) is fixed.  But I don't understand why those new type-prescription rules in books/std/lists/intersection.lisp are useful.  I'd expect that (member a (intersection$ x y) would normally be rewritten to (and (member-equal a x) (member-equal a y)), so that the intersection$ call doesn't even appear.  (I'd also use member-equal instead of equal, but I suppose that removal of guard-holders makes that change unnecessary.)  I think type-prescription rules are normally written to match terms that are in normal form (i.e., fully rewritten).

airbornemihir:
Hi Matt,

On 7/15/20 11:30 AM, MattKaufmann wrote:
> But I don't understand why those new type-prescription rules in 
> books/std/lists/intersection.lisp are useful. I'd expect that (member 
> a (intersection$ x y) would normally be rewritten to (and 
> (member-equal a x) (member-equal a y)), so that the intersection$ call 
> doesn't even appear. (I'd also use member-equal instead of equal, but 
> I suppose that removal of guard-holders makes that change 
> unnecessary.) I think type-prescription rules are normally written to 
> match terms that are in normal form (i.e., fully rewritten).
Thanks. At the moment it turns out these rules aren't used anywhere, 
which I checked by grepping through cert.out files, so that may be a 
reason to take them out. If I'm not mistaken though, the 
type-prescription rules are a little more general in the sense that they 
may replace a term matching (member a (intersection$ x y)) with nil even 
outside of an iff context, whereas the rewrite rule is tied to the iff 
context.

MattKaufmann:
@airbornemihir Yes, I should have thought of the non-iff context case (though it's probably rare, as you basically suggest).

PR #1157, opened 2020-07-30
Fix ACL2(p) bug in SBCL

ragerdl:
In SBCL, change the way we create the
*waterfall-parallelism-timings-ht* by including an argument intended
to make read/writes threadsafe.

Without this change, we encounter errors like:

***********************************************
************ ABORTING from raw Lisp ***********
********** (see :DOC raw-lisp-error) **********
Error:  Unsafe concurrent operations on #<HASH-TABLE :TEST EQUAL :COUNT 8191 {1011E10023}> detected.
***********************************************

My hope is that this is easy for the acl2 maintainers to review and incorporate.  Please let me know if you have any questions -- perhaps with a phone call to minimize the back and forth.  Also, if you want to just copy+paste this into your own copy of `master` and push it yourself, that's okay with me.  I've made this request against `master`, as it's my guess that the maintainers would prefer to merge this change directly into `master` and then run the purity script, rather than wait for it to percolate through `testing`.  I'm also happy to create it against `testing`, as desired.

### DIFF
--- a/prove.lisp
+++ b/prove.lisp
@@ -6891,14 +6891,18 @@
            (let ((new-ht (make-hash-table :test 'equal :size (expt 2 13)
 
 ; Parallelism blemish: CCL locks these hashtable operations automatically
-; because of the argument :shared t below.  However in SBCL and LispWorks, we
-; should really lock these hashtable operations ourselves.  Note that the SBCL
-; documentation at http://www.sbcl.org/manual/Hash-Table-Extensions.html
-; describes a keyword, :synchronized, that is like CCL's :shared but is labeled
-; as "experimental".  At any rate, we are willing to take our chances for now
-; with SBCL and Lispworks.
+; because of the argument :shared t below.  SBCL documentation at
+; http://www.sbcl.org/manual/#Hash-Table-Extensions introduces the keyword,
+; :synchronized, that provides such a locking mechanism.  SBCL's :synchronized
+; is claimed to be experimental but has been around since at least 2011.  Using
+; this experimental feature is easier (and probably more performant) than
+; implenting the locking ourselves.  In LispWorks, we should probably lock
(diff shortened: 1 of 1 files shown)


ragerdl:
BTW, while I did check that ld'ing `projects/tau/bounders/elementary-bounders.lisp` successfully ld's after this change, I haven't run full tests against it.  I'm not saying it's impossible that I haven't correctly proofread this change, but if Matt's going to run a regression anyway, I'd prefer to let him do it.  Otherwise, I'll rig something up in leeroy2 that handles whatever cross-product of Lisps and ACL2 features that we want to test.

MattKaufmann:
Thanks, @ragerdl!  I've incorporated your change into my copy (with minor edits to the comments) and am about to run a regression.  If things look good, I'll push to github (and credit you of course).

ragerdl:
Thanks for the quick turn-around!

MattKaufmann:
I incorporated this change in commit 90e46f3ee13e3e69e25a136b98a5ada25acaabe1; so I will close this pull request momentarily.  Now all books are certifiable with an ACL2(p) regression with ACL2_CUSTOMIZATION set to acl2-customization-files/parallel-resource-based.lisp.  Note that waterfall-parallelism is turned off in some books because of features it doesn't support, but one book, books/projects/fm9001/reg.lisp, has waterfall-parallelism turned off (by subsequent commit 283d2c2db635f0b613052b4a64c99336c7a13ff5) because of a mysterious hang that I was able to reproduce.

ragerdl:
The SBCL `regression-everything` build is confirmed to be working with SBCL (see http://leeroy2.defthm.com/job/acl2-multiple-flavors-and-lisps-monthly-regression-everything/ACL2_PAR=p,ACL2_REAL=no_r,LISP=sbcl/).  This is neat -- thanks for the help Matt!

PR #1135, opened 2020-06-06
Make no-sqrt.lisp certifiable in all ACL2 versions

kini:
Another old branch I found lying around. I think I was planning to edit `iter-sqrt.lisp` in a similar fashion, to add an actual `sqrt` function and prove `(equal (* (sqrt x) (sqrt x)) x)` for appropriate `x` (in ACL2(r) only, of course). But I never got around to figuring out how to do that, haha...

Link to diff view of this PR with whitespace changes ignored: https://github.com/acl2/acl2/pull/1135/files?w=1

#### Make `no-sqrt.lisp` certifiable in all ACL2 versions

By replacing `realp` with `real/rationalp` and `complexp` with `complex/complex-rationalp`, this book becomes certifiable with standard ACL2.  As far as I've seen, usually this is done to make ACL2-certifiable books become ACL2(r)-certifiable rather than vice versa, but I think it's nice to try to make ACL2(r)-related books ACL2-certifiable if possible, too.

The final theorem says that if x^2 = 2, then x is `real/rationalp` and not `rationalp`.  In ACL2(r) this means that x is an irrational real number, but in standard ACL2, this means that x is both rational and not rational, which is impossible (i.e. such an x cannot exist).  So for the standard ACL2 case (`#-non-standard-analysis`), I make the theorem a corollary which simply states that there is no x such that x^2 = 2.

The comments in this file need to be fixed as well -- they were written in a way that implied that ACL2(r) had become the mainstream ACL2 distribution, whereas in fact it is (currently) maintained in parallel with a standard version of ACL2 that does not include non-standard analysis support.  I'll do that in a separate commit.

#### Clean up comments

See parent commit's message for details.

#### Move iter-sqrt to nsa/sqrt-iter

(I figure it's better for the name to start with "sqrt" so it's visually close to sqrt.lisp in a directory listing.)

#### Deduplicate sqrt-iter.lisp and sqrt.lisp

Since after the previous few commits sqrt-iter.lisp was very similar to the first part of sqrt.lisp, this commit now removes the first part of sqrt.lisp and replaces it with (include-book "sqrt-iter") to avoid having two copies of those events lying around.

PR #1159, opened 2020-08-02
Make include-book local; change a few references

ragerdl:
This is at least a partial fixing of the ACL2(r) build.

The key was to make the inclusion of arithmetic/top local.  I think it's unreasonable to roast @kini for this oversight, because he had clearly asked for a review.  It was interesting to me how many books broke as a result.

I'm currently running both a vanilla and acl2(r) build overnight on leeroy2 (see http://leeroy2.defthm.com/job/fixing-acl2r/) and will merge accordingly tomorrow.  I'll need to merge in the latest from the testing branch before doing that, but I don't plan to worry about it too much.

Please nobody "undo" Keshav's PR #1135.  At this point, let's just push through with it.  Hopefully Keshav can address whatever feedback Ruben has in a follow-up commit.

### DIFF
--- a/books/nonstd/nsa/sqrt-iter.lisp
+++ b/books/nonstd/nsa/sqrt-iter.lisp
@@ -48,7 +48,7 @@ To load this book, it is sufficient to do something like this:
 
 (in-package "ACL2")
 
-(include-book "arithmetic/top" :dir :system)
+(local (include-book "arithmetic/top" :dir :system))
 
 ;;
 ;; We start out by defining the bisection approximation to square root.  At one
--- a/books/workshops/2018/gamboa-cowles/complex-continuity.lisp
+++ b/books/workshops/2018/gamboa-cowles/complex-continuity.lisp
@@ -10,7 +10,6 @@
 (include-book "complex-lemmas")
 (include-book "norm2")
 
-
 (defun region-p (region)
   (and (consp region)
        (interval-p (car region))


kini:
Thanks for the fix, David! Now I see how the problem arose.

In commit 4b20676b85be79cc197615d829f66a0b3bdc581d, I tried to make `iter-sqrt.lisp` (later renamed to `sqrt-iter.lisp`) the same as the top part of `nsa/sqrt.lisp`. Then, in commit 203cd3d5fe3a4b4428a9a696da95015029a6053c, I replaced the top part of `nsa/sqrt.lisp` with `(include-book "sqrt-iter")`. This was supposed to be a no-op because I had already made `sqrt-iter.lisp` equal to the top part of `nsa/sqrt.lisp`.

But I had made an oversight -- I neglected to sync up the `include-book` forms at the top of each file.

`iter-sqrt.lisp`:
```lisp
(include-book "arithmetic/top" :dir :system)
```

`nsa/sqrt.lisp`:
```lisp
(local (include-book "arithmetic/idiv" :dir :system))
(local (include-book "arithmetic/realp" :dir :system))
(local (include-book "arithmetic/top" :dir :system))
```

So if this had been caught in review before my PR was merged, I'd have probably amended commit 4b20676b85be79cc197615d829f66a0b3bdc581d to add that change (i.e. changing the first block above into the second block above in `iter-sqrt.lisp`).

ragerdl:
Per http://leeroy2.defthm.com/job/fixing-acl2r/3/ the vanilla and acl2(r) regression-everythings both passed, so I'm going to rebase and merge.

W.r.t. @kini's comment -- indeed, as you know but I will highlight anyway, since it's an opportunity to make a well-known but often forgotten pair of points: _unless one is thinking carefully about the interface of one's library, it's often best to have local theory events_.  This is illustrated once folks start including one's library and then they end up needing to maintain their books whenever the underlying library changes.  Also, in the case of the arithmetic books, _folks often prefer to dynamically choose their arithmetic library_.  So, it's kind of poor form to non-locally include any arithmetic book.

In case anyone is curious how this type of problem can be diagnosed, after manually disabling a couple rules (found by `(brr t)` and `(cw-gstack)`), I asked the question, "where did this problematic rule come from?" via `:pe <rulename>`.  Looking through the `:pe` answer enabled me to figure out that arithmetic/top used to effectively be a local include but wasn't anymore.  So, there was a good chance that just making the include-book local would work, and it did.


In @kini's case, his methodology of preserving local include books would have worked.  Just trying to make an abstraction available for others to learn too.

ragerdl:
BTW, I don't think I'd have done this work without a 32+ core machine available to me, so +1 for leeroy2.  This regression work is a pain on small machines.

PR #1160, opened 2020-08-02
Be more flexible in loading libipasirglucose4.so

kini:
The doc topic IPASIR::BUILDING-AN-IPASIR-SOLVER-LIBRARY claims that to
get the ipasir shared library working, it is sufficient to set
$IPASIR_SHARED_LIBRARY to the library's filename alone, and put its
directory in the $LD_LIBRARY_PATH list.

Unfortunately, while the loading code in load-ipasir-sharedlib-raw.lsp
itself is able to handle this, features.sh is stricter and requires
$IPASIR_SHARED_LIBRARY to contain a full path to the shared library,
so running `make regression` will skip the ipasir books if the library
is pointed to by a directory path in $LD_LIBRARY_PATH and a filename
in $IPASIR_SHARED_LIBRARY.

This commit generally expands the different ways in which the user is
able to provide an ipasir shared library for the ipasir-related books
to build from.  See the changes made to the aforementioned doc topic
for more details.

### DIFF
--- a/books/build/features.sh
+++ b/books/build/features.sh
@@ -73,6 +73,45 @@ rm -f Makefile-features;
 ACL2_CUSTOMIZATION=NONE $STARTJOB -c "$ACL2 < cert_features.lsp &> Makefile-features.out" || echo "*** Failed to run ACL2! ***" 1>&2
 
 
+search_ld_library_path () {
+    local target=$1
+    local dirs=(); IFS=:; for x in $LD_LIBRARY_PATH; do dirs+=("$x"); done; unset IFS
+    for x in "${dirs[@]}"; do
+        [[ -e $x/$target ]] && return 0
+    done
+    return 1
+}
+
+check_ipasir () {
+    if [[ $IPASIR_SHARED_LIBRARY == */* && -e $IPASIR_SHARED_LIBRARY ]]; then
+        # found at user-supplied absolute or relative path (ld.so
+        # treats it as such if it contains a slash)
+        return 0
+    else
+        # look in $LD_LIBRARY_PATH for either the user-supplied library name or
+        # libipasirglucose4.so if the user didn't supply a library name
+        if search_ld_library_path ${IPASIR_SHARED_LIBRARY:-libipasirglucose4.so}; then
+            return 0
+        fi
+    fi
+    # Ideally there should be another if-branch here checking the actual
+    # dynamic load path using `ldconfig -p` or `ld.so` or something but there
+    # doesn't seem to be an easy portable way to do that, so I'm skipping it.
+    if [[ -n $IPASIR_SHARED_LIBRARY ]]; then
+        # user supplied path but this simple script couldn't find it;
+        # maybe ld.so still can, so we press on
+        cat <<EOF 1>&2
+!!! WARNING: \$IPASIR_SHARED_LIBRARY was set to "$IPASIR_SHARED_LIBRARY", which
+             could be found neither as an absolute or relative path, nor in
+             \$LD_LIBRARY_PATH.  Books requiring an ipasir shared library will
+             be tried anyway, but may fail.  If you want to skip them, unset
+             \$IPASIR_SHARED_LIBRARY.
+EOF
+        return 0
+    fi
+    return 1
+}
+
 echo "Determining whether Glucose is installed" 1>&2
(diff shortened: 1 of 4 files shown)


kini:
Thanks for the review!

> This looks pretty good to me. The behavior is a bit different in that when the user has set an IPASIR_SHARED_LIBRARY that features.sh couldn't find, we'll try to forge ahead with a warning anyway, in case it falls into some behavior of the dynamic library loader that we're not predicting. I think I would slightly prefer to fail early instead, but if you've found actual cases where we can't find the library but the loader can, then I'm OK with it. 

Before this PR, the early failure was what was causing me problems -- following the documentation, I had set `LD_LIBRARY_PATH` and then set `IPASIR_SHARED_LIBRARY=libipasirglucose4.so`, but it wasn't working since `features.sh` expected an absolute path. So I figured that going forward, I'd prefer for features.sh to forge ahead if it wasn't sure.

My eventual goal is to bring this more in line with a standard interaction between an installed application and an installed library in a system. In particular, if I create an OS-level package for libipasirglucose4.so and installed it on my system in a standard way, I'd like ACL2+books to automatically recognize the fact that an IPASIR shared library is now available, without the need to change the way I'm running ACL2 -- much like what currently happens for other tools like glucose, z3, and abc -- i.e., without the need for setting environment variables.

I think the easiest and most accurate way to detect whether an IPASIR shared library is available is to just call `dlopen()` and see if we're able to load one or not. If I can do that, that simplifies the logic quite a bit, I think.

But I couldn't think of an easy way to do that from a bash script before certifying any books. I suppose I could compile a small C program on the fly (`#include<dlfcn.h>\nint main(int x,char**y){return!dlopen(y[1],RTLD_LAZY);}`), but do we already require a working C toolchain to build the books, or would that be adding a (pretty large) dependency? 

Or if you can think of another easy way to call `dlopen()` from `features.sh`, let me know. I know very little about Perl, but maybe there's some way to use Perl here? Clearly Perl is already a dependency of ACL2+books.

By the way, this could also be used to check for the presence of `libssl`, which is a currently invisible dependency of `USE_QUICKLISP=1` builds. Multiple people have written to the acl2-help mailing list after being confused by an error about not being able to load `libssl` when trying to certify `quicklisp/base`. It would be nice if we could check for that in `features.sh` and set `USE_QUICKLISP=0` in such cases.

>I'm not totally decided about whether it's appropriate to default to libipasirglucose4.so when the user hasn't set the environment var -- it should also be possible to use e.g. Cadical or (I believe) MiniSAT as well. It also should be possible to build one on MacOS, in which case it would be a .dylib rather than a .so. But I suppose it's practical given that probably most people using ipasir with ACL2 are running linux and followed the instructions in the documentation.

Good point about `.dylib` vs `.so`; probably it would be best to autodetect the platform and use the appropriate file extension, both in `features.sh` and in `load-ipasir-sharedlib-raw.lsp`. But for now, Mac users can still rely on the old behavior by setting `$IPASIR_SHARED_LIBRARY` to an absolute path, since my PR should be backwards-compatible.

In terms of selecting alternate implementations, once `features.sh` is smart enough to only need the library name and not a full path (`load-ipasir-sharedlib-raw.lsp` already is smart enough), then `$IPASIR_SHARED_LIBRARY` can be used as a selector for which library you want to use.

There's already some precedent for this since the documentation before my PR already indicates that you can use `$IPASIR_SHARED_LIBRARY` to indicate just a library filename, as long as you set `$LD_LIBRARY_PATH` properly (even though that wasn't actually true). So a generalization of that would be that you can also use `$IPASIR_SHARED_LIBRARY` to indicate just a library filename even if the library is installed into system dirs and cached with `ldconfig`, with no need for setting `$LD_LIBRARY_PATH`.

That said, I think it's fine for there to be a default choice among several. libipasirglucose4 seems as good a choice as any, I suppose (?).

> I haven't run this or anything but I trust you've tested the obvious scenarios. If you have good reasons for the choices I mentioned above then I'm good with it.

I haven't run full regressions on every scenario, but I tested what `features.sh` did on each scenario, and then I also ran a full regression with `$LD_LIBRARY_PATH` set and `$IPASIR_SHARED_LIBRARY` unset, to make sure that the new code path in `load-ipasir-sharedlib-raw.lsp` worked properly.

solswords:
OK, seems fine.

PR #1161, opened 2020-08-03
Update quicklisp bundle

kini:
Mainly because I'd like to incorporate the changes from jaredcdavis/shellpool#14, which are needed for using `USE_QUICKLISP=1` on NixOS when using SBCL as the underlying lisp. (`USE_QUICKLISP=1` should already work fine with Nix and SBCL on a non-NixOS distribution, though -- the problem is that there's no `/bin/bash`, `/usr/bin/bash`, or `/usr/local/bin/bash` on NixOS.)

Note: still running the regression currently.


kini:
Regression completed successfully except for these two books:

- `nonstd/workshops/1999/calculus/solutions/riemann-rcfn-helpers`
- `projects/taspi/tree-score/opt-pairwise`

In these two cases, ccl apparently segfaulted, but only *after* successfully writing the `.cert` files for each book. So rerunning `make regression` afterwards did nothing. I also removed the two `.cert` files and recertified those two books with no issues. So whatever was causing the segfaults must not be related to this PR, I guess, especially since neither of those books seem to use anything quicklisp-related. Maybe this is related to the fact that my ccl version is 1.11.5 while it's recommended to use 1.12, but I've successfully run several full regressions with ccl v1.11.5 in the last few days.

ragerdl:
This is a big enough change that I'd like to run it on leeroy before reviewing it.  Here's the build, which I just kicked off.  Feel free to nudge me later if I forget to follow up on this.

http://leeroy2.defthm.com/job/build-to-test-large-changes-1/

kini:
Hmm, looks like there were some failures due to running out of disk space on leeroy?

ragerdl:
That passed.  Gonna let anyone else like @solswords or @acoglio chime in and then @kini can [rebase and] merge

PR #1164, opened 2020-08-05
Print certification failure tails atomically

kini:
When running a regression, books print only two lines when they are certified
successfully (the "Making ~" and the "Built ~" lines), but when they fail, they
print the last 300 lines of their `.cert.out` file to aid in debugging.  Before
and after this 300-line tail are header and footer lines identifying which book
the failure log is coming from.

When running regressions with a large `-j` number (i.e. many builds running in
parallel), sometimes output starts interleaving a bit, and the headers and
footers can get separated from the 300-line tail, making it harder to tell at a
glance which file a 300-line tail is coming from.

While this is sort of inevitable (short of taking full control over
stdout/stderr of the build processes and doing some smart buffering), one
low-hanging fruit is the fact that the header and footer are printed in
`make_cert_help.pl` while the 300-line tail is printed in two new chained
processes spawned between those two `print` statements, which has to do disk I/O
to seek to the last 300 lines of the appropriate `.cert.out` file, adding some
latency.

This commit makes `make_cert_help.pl` create and format the 300-line tail in
memory before dumping the header, the tail, and the footer all in one `print`
statement.  Hopefully this will mitigate the interleaving to some extent...


books/build/make_cert_help.pl
```diff
@@ -831,15 +831,16 @@ sub extract_hostname_from_outfile
...
-    system("tail -300 $outfile | sed 's/^/   | /'") if $outfile;
-    print "\n\n";
+    print "\n**$taskname FAILED** for $dir$file.lisp\n\n" .
+        ($outfile ? `tail -300 $outfile | sed 's/^/   | /'` : "") .
+        "\n**$taskname FAILED** for $dir$file.lisp\n\n";
 
     if ($ON_FAILURE_CMD) {
-	system($ON_FAILURE_CMD);
+        print "\n-- Executing ON_FAILURE_CMD='$ON_FAILURE_CMD' for $dir$file.lisp\n\n" .
+            `$ON_FAILURE_CMD` .
```
kini, line 840: What do people put in `$ON_FAILURE_CMD`? If it prints stuff to `stderr`, then this won't be atomic anymore (and in fact the `stderr` output may show up before it prints "Executing ON_FAILURE_CMD=..."). It seems like writing `` `{ $ON_FAILURE_CMD ; } 2>&1` `` ought to work, but maybe I'm missing some weird corner case.

solswords: Personally I forgot that ON_FAILURE_CMD existed. I don't know that anyone uses it.  Looks like Jared added it in 2011.

solswords:
This seems good to me.

kini:
Thanks!

PR #1165, opened 2020-08-05
Add to the lemmas automatically generated by defsort

airbornemihir:
I thought it would be nice to have the obvious set-equiv theorem automatically generated by defsort. I mostly stuck to local events, but I did add an event to the nthcdr book. Additionally, I made local copies of lemmas that appear in other books, such as subsetp-trans, subsetp-member and subsetp-append1. Let me know if you would prefer me to include those books instead.

### DIFF
--- a/books/defsort/defsort.lisp
+++ b/books/defsort/defsort.lisp
@@ -797,9 +797,8 @@ it has a special hack for that particular case.</p>
                        consp-of-comparable-mergesort
                        ,subst1)))
 
-           (defthm ,(mksym prefix "-IS-IDENTITY-UNDER-SET-EQUIV")
-             (equal (consp (,sort x . ,extra-args))
-                    (consp x))
+           (defthm ,(mksym prefix "-SORT-IS-IDENTITY-UNDER-SET-EQUIV")
+             (set-equiv (,sort x . ,extra-args) x)
              :hints ((defsort-functional-inst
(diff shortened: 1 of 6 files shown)


books/defsort/defsort.lisp
```diff
@@ -795,6 +795,13 @@ it has a special hack for that particular case.</p>
                     (consp x))
              :hints ((defsort-functional-inst
                        consp-of-comparable-mergesort
+                       ,subst1)))
+
+           (defthm ,(mksym prefix "-IS-IDENTITY-UNDER-SET-EQUIV")
+             (equal (consp (,sort x . ,extra-args))
+                    (consp x))
+             :hints ((defsort-functional-inst
+                       comparable-mergesort-under-set-equiv
```
solswords, line 804: This doesn't seem like the right thorem body?

airbornemihir: @solswords You're correct. I've fixed this and I'm waiting for the regression to go through.

BTW, there are a number of lemmas directly preceding this which appear to be misnamed. As a concrete example, see examples.lisp which generates lemmas `string-no-duplicatesp-equal`, `string-true-listp`, `string-len` and `string-consp` which should all have `string-sort` in their names somewhere. I would submit a patch but it looks like there's code further down the line which depends on these awkward names.

PR #1166, opened 2020-08-06
Automatically generate some rules

airbornemihir:
There's a fairly common kind of rule which can be automatically generated by `fty::defalist`. I removed some existing rules which either had a name conflict with the newly generated rules, or else did the same thing under a different name.


acoglio:
The new rule looks good to me.

PR #1168, opened 2020-08-09
Add the notion of total order to defsort and prove something useful about sorting based on a total order

airbornemihir:
It turns out that it is possible to prove
`(implies (set-equiv x y) (equal (comparable-mergesort (remove-duplicates x)) (comparable-mergesort (remove-duplicates y))))`
under the assumption that `compare<` is a total order. This is a useful property in some work I'm doing, so I decided to prove it generically and incorporate it into the defsort books. However, I haven't attempted to change the actual `defsort` macro to automatically prove this theorem - I'm happy to help out with that task later, but for now I'm going to instantiate it manually and get moving.
I also fixed up some misnomers generated by the `defsort` macro - specifically, theorem names like `string-no-duplicatesp-equal`, `string-true-listp`, `string-len` and `string-consp` which can be seen in the defsort examples. I changed these names to refer to "string-sort" etc. instead of "string" etc.


airbornemihir:
Quick note: I mentioned above that I haven't changed `defsort` but I did manage to find an example of the kind of comparator to which this would be applicable (`<<`) and prove the property (`common-<<-sort-for-perms`) with a computed hint of the same kind as used in `defsort`. It's in the examples file.

airbornemihir:
A commit recently added to this PR adds a new kind of rule to the rules generated by `defalist`.

PR #1178, opened 2020-09-04
Add a lemma about str::charlisteqv

airbornemihir:
This lemma follows from the definition of charlisteqv, but doesn't seem to exist although I grepped for similar lemmas.


airbornemihir:
Thanks @acoglio .

PR #1179, opened 2020-09-06
Add a (disabled) lemma to std/lists/take

airbornemihir:
I'd previously added this lemma to the defsort books; I figured I'd move it to a place where it could be used more generally. It's helpful for `dec-induct` inductions.


airbornemihir:
Interested in any suggestions people might have for the lemma name which is kinda non-standard, and if it makes sense to warn people about likely problems from enabling this at the same time as `take`, perhaps in the form of a theory-invariant.

acoglio:
Since it is essentially a redefinition of `take`, an option could be `take-alt-def-as-append-and-nth` or just `take-as-append-and-nth`. I would add a theory invariant too, as you suggest.

PR #1180, opened 2020-09-07
small doc changes to run-script and xdoc

bendyarm:
I changed the wording a little in run-script that cleared up a misunderstanding I had when I first read it, and added some links that explain the keyword arguments.
The only change in xdoc was a cut-and-paste error where it referred to csee when it should have referred to tsee.


MattKaufmann:
The changes look good to me; thanks, Eric.

PR #1182, opened 2020-09-09
add depends-on for script file

bendyarm:
add depends-on form referencing script file to run-script doc and to mini-proveall demo


MattKaufmann:
Looks good; thanks!

PR #1183, opened 2020-09-12
[FTY] Add the :EXECUTABLEP option to DEFFIXTYPE.

acoglio:
[Making a PR on purpose -- please don't merge until Sol reviews it.]

This was discussed with Sol Swords.

The documentation of DEFFIXTYPE mentions :EXECUTABLEP, but the implementation
uses :EXECP. The former seems preferable to the latter, so we want to replace
the latter with the former in the implementation.

As suggested by Sol, we do that in three steps:

1. We add :EXECUTABLEP without removing :EXECP; the default T, its conjoining
with EXECP, and the disjoining of its negation with (NOT EXECP), should ensure
no change of behavior in existing calls. The full regression passes.

2. We switch existing calls to use :EXECUTABLEP, whose treatment in the
implementation is symmetric to :EXECP, so again there should be no change in
behavior.

3. We remove :EXECP.

This commit does step #1.

### DIFF
--- a/books/centaur/fty/fixtype.lisp
+++ b/books/centaur/fty/fixtype.lisp
@@ -55,7 +55,7 @@
 (defun get-fixtypes-alist (world)
   (cdr (assoc 'fixtype-alist (table-alist 'fixtypes world))))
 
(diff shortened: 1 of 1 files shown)


solswords (approved):
Looks good to me. Thanks!

PR #1184, opened 2020-09-17
Remove reference to if*, unrelated to non-case-splitting-logic

airbornemihir:
Unless I'm mistaken, if* has its own documentation topic and is separate from star logic.


solswords:
I'm two ways about this. That book doesn't define if* (it's built in) or any rules about it but it is basically similar to the functions defined in that book.  Maybe it could use some more explanation, especially since the if* topic is all about using it with the BDD package, but it's not unrelated -- the starlogic functions are named according to the convention established by if*, for example.

PR #1185, opened 2020-09-17
Add two lemmas to std/lists/prefixp

airbornemihir:
I've left subsetp-when-prefixp disabled because it can potentially cause a lot of backchaining.


books/std/lists/prefixp.lisp
```diff
@@ -195,4 +196,14 @@ the list @('y')."
...
-             :in-theory (enable prefixp)) )))
+             :in-theory (enable prefixp)) ))
+
+  (defthmd subsetp-when-prefixp
+    (implies (prefixp x y)
+             (subsetp-equal x y))
+    :hints (("goal" :in-theory (enable prefixp)
+             :induct (prefixp x y))))
+
+  (defthm when-prefixp-append-same
```
acoglio, line 207: It's difficult to come with a good name for this theorem. Perhaps I would go with `prefixp-of-append-x-y-and-x`.

airbornemihir: Thanks for taking the time to review; this one is hard to name as you said. @ericwhitmansmith perhaps you could weigh in, if you have a moment.

airbornemihir: There's also a loose convention of referring to to arguments of a function in terms of numbers - this is true for every lemma name automatically generated by `defcong`. Going by this, we could rename the existing `prefixp-of-append` to `prefixp-of-append-2` and this lemma, `when-prefixp-append-same`, to `prefixp-of-append-1`. Renaming the existing lemma is likely to require very little change to the community books since only two books seem to use that lemma at present.

ericwhitmansmith: FWIW:

I agree with using arg numbers rather than arg names in rule names. I usually use "arg1", "arg2", etc. to be clear.  I use plain numbers to indicate different variants of a rule (foo-1, foo-2, etc.).

I would use the name prefixp-of-append-arg1-same for the rule whose body is (iff (prefixp (append x y) x) (atom y)).  (Actually, I would also add -iff to the end of the name, but in this case you can just replace iff with equal to make the rule a bit better.)

I try to use "same" in a rule name whenever a variable appears more than once in the LHS, since that makes the rule less general.  

Sometimes, when the names are awkward, it indicates that rules can be generalized.  For example, instead of the rule above, I would prove this more general one:

(defthm prefixp-of-append-arg1
  (equal (prefixp (append x y) z)
         (and (<= (+ (len x) (len y)) (len z))
              (equal (true-list-fix x)
                     (take (len x) z))
              (prefixp (true-list-fix y)
                       (nthcdr (len x) z)))))

I think your rule follows fairly simply from this (using a few facts about take and nthcdr).

PR #1188, opened 2020-09-23
[cert.pl] Add option to suppress stack traces.

ericwhitmansmith:
Disclaimer: I don't know perl.  Someone who does should look this over.

### DIFF
--- a/books/build/make_cert_help.pl
+++ b/books/build/make_cert_help.pl
@@ -581,7 +581,7 @@ $instrs .= "#+(and hons (not acl2-par)) (profile-fn 'prove)\n";
 $instrs .= "#+(and hons (not acl2-par)) (profile-fn 'certify-book-fn)\n";
 $instrs .= "(acl2::lp)\n";
 # We used to comment this out, but maybe it's better to leave this enabled by default?
-$instrs .= "(set-debugger-enable :bt)\n" if ($STACK_TRACES);;
+$instrs .= "(set-debugger-enable :bt)\n" if ($STACK_TRACES);
 $instrs .= "(acl2::in-package \"ACL2\")\n\n";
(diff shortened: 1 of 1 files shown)


books/build/make_cert_help.pl
```diff
@@ -580,7 +581,7 @@ sub parse_certify_flags
 $instrs .= "#+(and hons (not acl2-par)) (profile-fn 'certify-book-fn)\n";
 $instrs .= "(acl2::lp)\n";
 # We used to comment this out, but maybe it's better to leave this enabled by default?
-$instrs .= "(set-debugger-enable :bt)\n";
+$instrs .= "(set-debugger-enable :bt)\n" if ($STACK_TRACES);;
```
ragerdl, line 584: Is the double semi-colon here intentional?

solswords:
Seems fine, other than the double-semi (which probably doesn't do anything?)

ericwhitmansmith:
Double semi was not intentional, so I fixed it.   Thanks!

MattKaufmann:
Eric, would you document the new NO_STACK_TRACES environment variable, perhaps somewhere under the :doc for cert.pl?

ericwhitmansmith:
I'd be happy to do that, but I could use some guidance on where to put it.  I looked for a precedent, but I actually don't see any xdoc for most of the other environment vars in that section (e.g., ON_FAILURE_COMMAND, or ACL2_BOOKS_DEBUG, whatever those are).  Maybe I missed it.  Or maybe we can take documenting these vars off the critical path for accepting this change.

MattKaufmann:
Good point, Eric; I didn't know (or remember) about the others.  Yes, I agree that documentation shouldn't be a requirement for accepting the pull request.

PR #1189, opened 2020-09-23
Add a new utility, find-funs

airbornemihir:
I hacked together this nearly-verbatim copy of find-lemmas, modified to instead search for functions. My use case for it was to look for existing functions that might already do something I was considering doing, so I might benefit from someone else already having proved the obvious lemmas about that function. The way I went about it was to include the book `doc/top` and then call `find-funs` - I figure this use case might apply to others, too.

Readers who might not want to merge this pull request themselves can still help by answering: Should this utility have a different name (perhaps `wander-the-world-looking-for-fun`)? Should this utility be in a different file? Should this utility be subsumed into `find-lemmas` with some kind of parameter telling the new `find-lemmas` to include function definitions?


MattKaufmann:
That looks like a good idea.  I don't plan to read the code, but I'm happy for you to make such a change -- just please clearly label what you added.  For example, you could could say, just below the existing header, that the design and implementation of find-funs are due to Mihir Mehta.  I wonder if a better name is find-defs (since the analogue of a lemma is a definition, rather than a symbol or a function).

As for subsuming, that might be a good idea too.  Perhaps all that's needed is to replace `(member-eq type '(defthm defaxiom defchoose))` by `(member-eq type '(defthm defaxiom defchoose defun))`.  Maybe find-lemmas could be changed to use keyword arguments instead of the current optional argument, and one keyword could be :types, which defaults to `'(defthm defaxiom defchoose)` to give the current behavior.

But those are just ideas that you might want to consider.  I'll be happy as long as the current functionality is preserved (modulo change to keyword arguments, if you like) and the credits are updated.

kini:
I haven't looked into the code, but if the definitions are substantially similar to those for `find-lemmas`, I'd try to create a general purpose utility that could find both lemmas and functions, and maybe alias `find-lemmas` to call it with the appropriate parameter. May I suggest `apropos` as a name for such a utility? (By analogy with the Emacs command of the same name, which itself derives from the Unix utility.)

MattKaufmann:
I like Keshav's idea.  The only problem is with the name:
acl2::apropos is imported into the "ACL2" package from the
"COMMON-LISP" package, so it can't be used as the name of an
ACL2 function.  (In fact it's a very useful Common Lisp
function.)  Maybe find-events?  So: `(find-lemmas x)` is
really

`(find-events x :types '(defthm defaxiom defchoose))`

and `(find-defs x)` is really

`(find-events x :types '(defun defuns))`.

MattKaufmann:
Thanks, Mihir, for adding those tests.  I'll wait until you've completed revisions before accepting the pull request.

MattKaufmann:
Looks good; thanks for doing this.  Some comments on find-lemmas.lisp (I didn't look at find-events-tests.lisp):

1. Find-defs should probably deal with mutual-recursion.  For example, if you submit the mutual-recursion near the top of :doc mutual-recursion and then run (find-defs 'consp), you won't see evenlp or oddlp.  As I suggested in a previous comment, "(find-defs x) is really (find-events x :types '(defun defuns))."

2. Probably find-events should include a guard of (symbol-listp types), perhaps strengthened to include that types is a subset of '(defthm defaxiom defchoose defun defuns).  For example, we get a raw Lisp error from (find-events '< :types 17).

3. If you want to polish the documentation, you could replace (defxdoc find-lemmas ...) by (defxdoc find-events ...) with minimal editing, and then have very short topics for find-lemmas and find-defs saying how they expand to find-events forms (e.g., for find-defs as in my comment in item 1 above).

MattKaufmann:
One more thing I just noticed: The new file ind-events-tests.lisp needs a copyright/author/license header.

MattKaufmann:
Replying to your comment in a separate email at https://github.com/airbornemihir/acl2/commit/90ec71443e5e0d086cf32b260ea8ecadedac7984:

Right, the namex for mutual-recursion is a list of the names; e.g., in the evenlp/oddlp example, it's (EVENLP ODDLP).  So instead of (formula namex t wrld) you can loop through the members of namex, collect their formulas, and conjoin them (or use all-fnnames-lst instead of all-fnnames in that case).

airbornemihir:
Thanks @MattKaufmann . I'll get back to this once I'm done with a couple of deadlines at the end of the week.

airbornemihir:
Hi @MattKaufmann . It looks like I never really got the time to understand the namex and make your requested changes. What do you think about accepting this request for now and discussing further what might need to be done to address mutual recursion?

MattKaufmann:
I'm OK with accepting the pull request, if you agree to the following.

1. Let me know whether mutual-recursion is supported.  If not, then
you'll add a comment like the following.

```
; Wart: mutual-recursion is not yet supported.  It may be reasonably
; straightforward to modify find-events-fn to add such support, as follows.
; The key is to modify the use of the let-bound variable, namex, which is a
; list of names in the mutual-recursion case rather than a single name.  So
; instead of (formula namex t wrld) you can loop through the members of namex,
; collect their formulas, and conjoin them before calling all-fnnames; or,
; leave them as a list and call all-fnnames-lst instead of all-fnnames.
```

2. Extend your leading comment to mention documentation, for example as
follows.

```
; This code was revised and extended with the addition of find-defs and
; find-events in September 2020 by Mihir Mehta.  It remains to document
; find-defs and perhaps find-events.
```

PR #1198, opened 2020-10-14
Enable a disabled lemma in std/lists/sets and add corollaries

airbornemihir:
The corollaries added are pretty much along the lines of subsetp-member in the same file. The only changes needed to actual books as a result of this change are in my filesystem books - and these changes show how these new corollaries make proofs simpler. Everywhere else, books that failed to certify were able to certify again once their useless-runes.lsp files were removed.


airbornemihir:
The timing is an interesting question. I decided to let Leeroy do some of the work of measuring timing, so I first pushed a whitespace-only change to sets.lisp:
http://leeroy2.defthm.com/job/acl2-testing/356/console
Then I pushed a version of sets.lisp with four corollaries,i.e. five rewrite rules, as displayed in this pull request:
http://leeroy2.defthm.com/job/acl2-testing/357/console
At a later point I'll push a version with two of the corollaries removed, so there are only three rewrite rules. I have such a version already, and I've checked that the regression passes - I'm just waiting for Leeroy to finish the first job so I don't end up pushing the second job out of the queue.

airbornemihir:
So, Leeroy job 356 took 41 minutes. Job 357 is running now, and the version with only three rewrite rules will pop up soon:
http://leeroy2.defthm.com/job/acl2-testing/358/console

ericwhitmansmith:
FWIW, I agree with Sol about corollaries.  I never use them, for the reason he cites: I prefer for each rule to have it's own simple name (a symbol) that indicates what it does.

airbornemihir:
Yep, I was going to write about that! I will throw up my hands and admit that I like corollaries, but even setting that aside, I think it's more justified to have corollaries now that there are only two, as opposed to four in the initial version of the PR. Also, I am hard-pressed to think of a better name for this lemma than intersectp-member - and I wouldn't want to change it in a rush anyway because I think some other books use this lemma, or a local redefinition of this lemma.

airbornemihir:
So, job 357 took 47 minutes, and job 358 took 45 minutes. These measurements are a little noisy because of the way Leeroy does things, but possibly some things are getting slowed down by the new rules. I inadvertently closed this PR while doing all this Leeroy stuff, but I'm still open to making the rule disabled again, and enabling it in the books which use it. Is there agreement on this point?

MattKaufmann:
I'm running an SBCL-based "everything" regression that I expect to complete around 11am Central time.  I plan to report how its time compares with one I did on Oct. 10.  It's on an unloaded machine that tends to produce non-noisy timings.

airbornemihir:
That would be great - thanks!

MattKaufmann:
It looks like the "everything" regression time is fine; see below.

Oct. 15 (today):
80705.357u 3128.755s 3:25:52.08 678.7%  0+0k 1689072+9249488io 6pf+0w

Oct. 10:
80680.849u 3150.000s 3:25:35.09 679.6%  0+0k 1700208+9239656io 250pf+0w

PR #1199, opened 2020-10-19
Extend fty::deflist as discussed on Slack

acoglio:
(Some changes are just removal of trailing whitespace, which my Emacs does automatically.)


solswords (approved):
Looks good, thanks!

PR #1205, opened 2020-11-06
Fix "make TAGS" and "make git-status" to work in Mac OS X zsh.

westfold:


MattKaufmann:
@westfold Thanks.  Have you tested thesechanges on Linux and on an older Mac OS that doesn't use zsh?

westfold:
Hi Matt,

I have tested on linux. I’ll try booting up an old Mac that I haven’t used in months and test there and get back to you. I got this fix from googling and it does seem pretty standard.

Stephen

westfold:
I tried running it on Mac OS 10.13.6 (High Sierra) under bash (and zsh) and it works.

Stephen

PR #1207, opened 2020-11-26
Move some lemmas to community books

airbornemihir:
This pull request moves two lemmas of general applicability to the STD and FTY libraries. It also updates the filesystem books and a number of useless-runes.lsp files - because of which it may be easier to examine it commit-by-commit.


airbornemihir (commented):
These lemmas might actually be a result of list-equiv refining true-list-list-equiv, which in turn is an instance of a general property that holds for all equivalence relations produced by fty::deflist as long as the final cdr is required to be nil.

airbornemihir:
> These lemmas might actually be a result of list-equiv refining true-list-list-equiv, which in turn is an instance of a general property that holds for all equivalence relations produced by fty::deflist as long as the final cdr is required to be nil.

I substantiated this observation by actually admitting a defrefinement form that made those two lemmas trivial. I figured that would probably be good enough for now, although there's room for modifying fty::deflist to prove these kinds of refinements automatically, and merged the PR.

PR #1208, opened 2020-12-14
Make a minor addition to :doc apt::simplify-defun-examples

airbornemihir:
I was trying out simplify-defun earlier, and I had to spend a while figuring out the correct book to include. Perhaps adding this line will help the next reader.

### DIFF
--- a/books/kestrel/apt/simplify-defun-doc.lisp
+++ b/books/kestrel/apt/simplify-defun-doc.lisp
@@ -704,6 +704,7 @@
  (defun bar (x)
    (declare (xargs :measure (nfix x)))
    (if (zp x) 0 (+ 1 1 (bar (+ -1 x)))))
+ (include-book \"kestrel/apt/simplify-defun-tests\" :dir :system)
  (simplify-defun bar
                  :new-name bar-simp
                  :thm-name bar-simplified
--- a/books/kestrel/apt/simplify-defun-doc.lisp
+++ b/books/kestrel/apt/simplify-defun-doc.lisp
@@ -704,7 +704,7 @@
  (defun bar (x)
    (declare (xargs :measure (nfix x)))
    (if (zp x) 0 (+ 1 1 (bar (+ -1 x)))))
- (include-book \"kestrel/apt/simplify-defun-tests\" :dir :system)
+ (include-book \"kestrel/apt/simplify\" :dir :system)
  (simplify-defun bar
                  :new-name bar-simp
                  :thm-name bar-simplified


MattKaufmann:
Good idea, adding the include-book.  But I wouldn't include the tests there, since that isn't what people typically should include.  Instead, this (but escaping the "" as you did before):

`(include-book "kestrel/apt/simplify" :dir :system)`

airbornemihir:
@MattKaufmann thanks, I can do that. How about, instead, having this include-book line in the file `books/kestrel/apt/simplify-defun-doc.lisp`? I'm not sure if it's a convention, but I've grown used to expecting that when a documentation topic starts like this:
Topic: APT::SIMPLIFY-DEFUN-EXAMPLES
Parent list: (APT::SIMPLIFY-DEFUN)
Source: [books]/kestrel/apt/simplify-defun-doc.lisp
I should be able to get the functions and theorems mentioned therein by including the book named as the source, that is, `kestrel/apt/simplify-defun-doc.lisp`.

MattKaufmann:
@airbornemihir I'm guessing that many readers of the topic may view that include-book as the way to start when using the tool, which is why I suggested what I did.  I think your point though is that (at least) some will view the include-book as a way to get the examples described in the topic.

We can have both, by saying something explicit about where to find this topic.  But I confess that I don't know why one would prefer to read the marked-up xdoc (note that simplify-defun-doc.lisp consists entirely of xdoc) rather than the documentation page.  Moreover, the book containing the documentation is at the top of every page (in both the web and acl2-doc versions).

Maybe I'm missing your point, though.

MattKaufmann:
Maybe I don't understand what you mean by "having this include-book line in the file".

airbornemihir:
On reflection, I agree with you. The three-line snippet I posted was from my Emacs buffer `acl2-doc`, but it's possible to read the documentation without it, for instance on the web or by typing `:doc` in the read-eval-print loop. So perhaps the "Source" file doesn't matter and it's best to just have this `include-book` explicitly in a code snippet, as I originally suggested and as you amended. If it looks fine to you now, perhaps you can merge this pull request.

MattKaufmann:
Thanks; I've merged.  Perhaps that acl2-doc line saying "Source" is a bit misleading; I think I'm willing to change it to ":DOC Source" or "Documentation source", for example (unless that breaks something).

airbornemihir:
That seems like a reasonable change. While many books include their own documentation, some don't; and your change would be a good way to make this clear.

PR #1218, opened 2021-01-05
[std] Add xdoc for defrulel and defruledl.

ericwhitmansmith:
This pull request adds documentation for defrulel and defruledl.


acoglio:
Looks good to me.

PR #1229, opened 2021-02-26
Add rewrite-quoted-constants to the rule-classes of some FTY events 

airbornemihir:
I've already found it useful, in my filesystem books, to rewrite some constants under the relation true-equiv which is automatically generated by FTY. I'm guessing there are circumstances where others will find this useful too, for their own types or for the base types.

Filesystem books are also updated as part of this pull request.


solswords:
Interesting.  I think this might make the normalize-const theorems generated by deffixequiv somewhat unnecessary, but will have to see.

airbornemihir:
Yep, it's interesting. A quick grep shows that we now have 1348 places across the regression suite where :rewrite-quoted-constant rules are used, the vast majority being derived from this change. I imagine Dr Moore might be amused to hear about this (he wrote the code for this new rule class), in addition to the fact that it caused zero proof failures anywhere.

PR #1231, opened 2021-02-27
Remove some normalize-const lemmas

airbornemihir:
There are a few different reasons for getting rid of these rules:

- Following the changes in #1229, I suspect they are redundant.
- They were disabled in a LOT of places, and at least one author of books disabling them has asked (https://utlists.utexas.edu/sympa/arc/acl2-help/2021-02/msg00046.html) for a mechanism to keep from generating them.
- I suspect they will lead to weird behaviour when they are enabled but the corresponding executable-counterpart rules are disabled. I can't offhand provide a thm showing this, but without the :e rule there is a possibility of rewriting constant c to (some-fix c) to (some-fix (some-fix c)) ... The :rewrite-quoted-constant mechanism, which is our replacement for these rules from #1229 , handles this possibility by refusing to rewrite if it can't establish that a different constant is being brought in by the rewrite.

I've had to make a lot of changes to books, so perhaps @TheBeaNerd @solswords and @shigoel will want to take a look.


books/centaur/aignet/observability.lisp
```diff
@@ -475,10 +475,20 @@
...
 (local
- (defthm nth-of-lit-eval-list
-   (bit-equiv (nth n (lit-eval-list lits invals regvals aignet))
-              (lit-eval (nth n lits) invals regvals aignet))
-   :hints(("Goal" :in-theory (enable nth lit-eval-list)))))
+ (encapsulate
+   ()
+
+   (local
+    (defthm lemma
```
solswords, line 482: Do you know why this lemma is needed? I would have thought the rewrite-quoted-constant rule for lit-fix under lit-equiv would have caused the NIL to normalize to 0, and I think there's probably also a rewrite rule for lit-eval of 0.

airbornemihir: Hi @solswords - as I think back to what I did here 9 days ago, I recall that without this lemma there was a subgoal with the hypothesis (equal (lit-eval nil invals regvals aignet) 1). The existence of that subgoal suggests that there may not have been a rewrite rule for lit-eval of 0.

PR #1236, opened 2021-03-12
Add a congruence to FTY

airbornemihir:


solswords:
I think this is fine, but I'd like to keep an eye on e.g. the VL/SV books in case something like this starts slowing down proofs. After this is merged, then when there are a large number of FTY lists in the theory, there will be one congruence for TAKE, NTH, etc for each, and so on every encounter of one of these functions the rewriter will need to check this list against the current geneqv to see what congruences it needs to propagate. I haven't seen this become a problem yet, but it's something to be aware of.  AFAIK we don't usually check to see whether something noticeably slowed down certification of certain books so I want to check on that when we know of something that might have that effect.

airbornemihir:
Hi Sol - that makes sense, although it seems that my most recent 
additions pertaining to take and nthcdr will only affect the geneqv when 
there already exists a foo-list-equiv context within which a take or 
nthcdr expression can be rewritten. But to address the broader concern 
about all these FTY lemmas taken together, I started working on them 
following commit 8ad1e8e 
<https://github.com/airbornemihir/acl2/commit/8ad1e8ee5c129b6cd4e31faabd9599813ac6e18a> 
so if there's a timing test needed, that would be the commit to compare 
against. I've also been somewhat sparing in what I add, in the sense 
that I've only added lemmas that I can see other books generating on 
their own - I have several lemmas still that I've kept within my own 
development (most or all of these are in hifat.lisp.)

I guess this could be more of an issue if there are many list 
equivalences in some kind of refinement hierarchy, i.e. something like 
p-list-equiv refining q-list-equiv refining r-list-equiv. That would be 
one place where the geneqv could start blowing up, linearly in the depth 
of the refinement hierarchy. Is that a concern in SV and VL?

Thanks for merging all the same.

airbornemihir:
There is actually an oranges-to-oranges timing comparison to be found on 
Leeroy - consider these two builds 
(http://leeroy2.defthm.com/job/acl2-master/720/ , 
http://leeroy2.defthm.com/job/acl2-master/889/), a month and a half 
apart, which both take 50 min to certify very similar sets of books.

MattKaufmann:
For what it's worth, I just did a CCL-based "everything" regression on
linux, using what was in git master around 11:35 am today Austin time
(commit hash df5a3f79df4fc5657574cda9a5a1d8b59d0a2b34), on the same UT
CS machine as on Feb. 20 (commit hash
1c9e1a432bb866f620839a499129e45e14110127), and there was no slowdown.

; today
120828.241u 2512.840s 4:57:12.70 691.6%	0+0k 2904264+9290024io 209pf+0w

; Feb. 20
121632.393u 2632.748s 4:59:01.06 692.6%	0+0k 1748152+9442512io 54pf+0w

-- Matt

airbornemihir:
Thanks, Matt, for keeping the kinds of detailed records that allow you to
ride up and answer these timing questions!

PR #1237, opened 2021-03-21
Add an example to xdoc

airbornemihir:


airbornemihir:
I guess I mixed up list-equiv and set-equiv for a moment. Never mind.

PR #1240, opened 2021-04-01
Slightly generalise two lemmas

airbornemihir:
This generalisation, applied to both these lemmas, is somewhat limited
because ACL2 is often able to draw the obvious conclusion when two
number-valued terms satisfy <= and >= at the same time. Sometimes ACL2
isn't, though, and that makes this generalisation worthwhile. It
brings in a case-split arising from the definitions of <= and >=
respectively, but that has not proven to be a problem anywhere in the
regression suite.

### DIFF
--- a/books/std/lists/prefixp.lisp
+++ b/books/std/lists/prefixp.lisp
@@ -135,7 +135,7 @@ the list @('y')."
                   (atom x))))
 
   (defthm prefixp-of-append-when-same-length
-    (implies (equal (len x) (len y))
+    (implies (<= (len x) (len y))
              (equal (prefixp x (append y z))
                     (prefixp x y)))
     :hints(("Goal"
@@ -143,7 +143,7 @@ the list @('y')."
             :in-theory (enable prefixp list-equiv))))
 
   (defthm prefixp-when-equal-lengths
-    (implies (equal (len x) (len y))
+    (implies (>= (len x) (len y))
              (equal (prefixp x y)
                     (list-equiv x y)))
     :hints(("Goal" :in-theory (enable prefixp list-equiv))))


airbornemihir:
If people have thoughts about new names for these lemmas to better reflect these changes, I'd be interested.

MattKaufmann:
I'd probably call the first one prefixp-append and the second one prefixp-is-list-equiv, but I'm not familiar with naming conventions in std/, so I'm far from sure that those names are appropriate.

I'm not seeing any case split, by the way.  But I haven't thought about this book, so maybe I'm missing something.

airbornemihir:
Ah, I thought that might be a little obscure! The term `(<= (len x) (len y))` translates to `(not (< (len y) (len x)))`, that is, `(if (< (len y) (len x)) nil t)`. Similarly for the other term, the one with a `>=`.

MattKaufmann:
Right, but those inequalities are in a hypothesis position, so rewriting alone will be applied -- no case-splitting, just rewriting of an IF call.  (Maybe that's what you meant be "case-split", but normally that term is used in ACL2 to denote splitting into subgoals.)

airbornemihir:
That's what I meant, and I guess I used the wrong term. To be just a little more precise, the problem I was thinking about is that `:brr` will report a failed rewrite sometimes just because one hypothesis has an `if` expression in it and ACL2 is unwilling to do the little bit of extra work (exploring both branches of the if-expression) that is needed to relieve that hypothesis. Thankfully, that hasn't been an issue anywhere in the regression suite as I mentioned above.

airbornemihir:
At any rate, your renaming suggestion reminded me that one of these lemmas is actually subsumed by another one - thanks!

MattKaufmann:
As far as I know, the rewriter isn't hesitant about exploring an IF in the manner you mentioned.  I wonder if the :brr report you mentioned is misleading.

airbornemihir:
Well, I don't have an example of ready at hand, sadly. I will, however, keep this on my list of things to email you about as and when they recur.

PR #1245, opened 2021-04-15
Update useless-runes.lsp files

airbornemihir:
I needed to update useless-runes files for the filesystem books, and it didn't take much longer to do it for other books as well.


airbornemihir:
Note, this commit is brought to you by our sponsors, bin/new-useless-runes-files.sh.

PR #1247, opened 2021-04-22
Defaggregate suppress-xdoc option

bendyarm:
The automatically-generated xdoc for defaggregate is not always what we want.  This PR adds the option :suppress-xdoc t to turn off xdoc for defaggregate, and documents the option.  The second commit makes use of the option for documenting r1cs.

### DIFF
--- a/books/kestrel/crypto/r1cs/sparse/doc.lisp
+++ b/books/kestrel/crypto/r1cs/sparse/doc.lisp
@@ -14,8 +14,10 @@
 
 (include-book "kestrel/utilities/gen-xdoc-for-file" :dir :system)
 
-;; Introduced by defaggregate
+;; No longer introduced by defaggregate.
+;; Now attaching to primary r1cs topic generated below with gen-xdoc-for-file.
 (acl2::defxdoc r1cs-constraintp
+  :parents (r1cs)
   :short "Recognize An R1CS constraint"
   :long "An R1CS constraint is a defaggregate with 3 fields, A, B, and C, each of which is a sparse vector.")
 
--- a/books/std/util/defaggregate.lisp
+++ b/books/std/util/defaggregate.lisp
@@ -204,7 +204,8 @@ appropriate, e.g., for @('name') above, you may wish to use a different
 @(':rule-classes') option.</p>
 
 <p>The embedded @(see xdoc) documentation gets incorporated into the
-documentation for the aggregate in a sensible way.</p>
+automatically-created documentation for the aggregate in a sensible way.
+This is overridden by the @(':suppress-xdoc t') option.</p>
 
 <p>The @(':default') value only affects the Make macro (see below).</p>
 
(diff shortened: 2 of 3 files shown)


acoglio (approved):
Looks good to me -- thanks!

solswords (approved):
Looks fine, thanks!

books/kestrel/crypto/r1cs/sparse/doc.lisp
```diff
@@ -14,8 +14,10 @@
 
 (include-book "kestrel/utilities/gen-xdoc-for-file" :dir :system)
 
-;; Introduced by defaggregate
+;; No longer introduced by defaggregate.
+;; Now attaching to primary r1cs topic generated below with gen-xdoc-for-file.
```
ericwhitmansmith, line 18: This seems incorrect.  The r1cs topic comes from books/kestrel/crypto/r1cs/doc.

books/std/util/defaggregate.lisp
```diff
@@ -1117,6 +1125,7 @@ would have had to call @('(student->fullname x)').  For instance:</p>
     :parents
     :short
     :long
+    :suppress-xdoc
```
ericwhitmansmith, line 1128: Suggest the polarity of the option be reversed, to avoid double negation.  Something like :generate-xdoc with default t.

books/kestrel/crypto/r1cs/sparse/doc.lisp
```diff
@@ -14,8 +14,10 @@
 
 (include-book "kestrel/utilities/gen-xdoc-for-file" :dir :system)
 
-;; Introduced by defaggregate
+;; No longer introduced by defaggregate.
```
ericwhitmansmith, line 17: This seems incorrect or at least unclear.  r1cs-constraintp is still introduced by defaggregate.

ericwhitmansmith (changes requested):
Thanks!  See 3 comments in this review.  It might be easiest if you change only defaggregate, not the r1cs files.  Then I can use the new option and deal with the r1cs doc changes.  In that case, the only relevant comment I have is to suggest switching the polarity of the new option (see comment attached to that file).

books/kestrel/crypto/r1cs/sparse/doc.lisp
bendyarm: If there were no call to gen-xdoc-for-file, there would be no primary r1cs topic.  The call to gen-xdoc-for-file is below my comment in doc.lisp. I believe my comment is correct.

books/std/util/defaggregate.lisp
bendyarm: The problem with reversing is that it generates automatically.  I think suppressing is the right concept.

books/kestrel/crypto/r1cs/sparse/doc.lisp
bendyarm: No, the xdoc for r1cs-constraintp is generated in doc.lisp by the defxdoc immediately below the comment.  It looks to me like :suppress-xdoc is suppressing all forms of xdoc on the defaggregate.

bendyarm:
> Thanks! See 3 comments in this review. It might be easiest if you change only defaggregate, not the r1cs files. Then I can use the new option and deal with the r1cs doc changes. In that case, the only relevant comment I have is to suggest switching the polarity of the new option (see comment attached to that file).

The change to defaggregate is in a separate commit from the r1cs files.  Is that not sufficient separation?

books/kestrel/crypto/r1cs/sparse/doc.lisp
bendyarm: Oh.  Now I get what you are saying here.  I misunderstood your comment so I made the wrong change.
OK, then let's merge the defaggregate commit and ignore the other one.

bendyarm:
I will open a new PR with just the defaggregate commit.

PR #1249, opened 2021-04-24
Make file locations shown at top of xdoc topic pages be links

ericwhitmansmith:
This patch improves the ACL2 manual by making the file paths shown at the top of each topic be clickable links to GitHub.

When the path is something like "[books]/path/to/file.lisp" the link goes to that particular file on GitHub.

When the path is just "ACL2 sources", the link goes to the GitHub page that lists the files in the main ACL2 directory (since we can't tell from the text "ACL2 Sources" which source file should be shown).

Warning: This required editing Javascript, but I don't know Javascript!

### DIFF
--- a/books/xdoc/fancy/xdoc.js
+++ b/books/xdoc/fancy/xdoc.js
@@ -781,9 +781,23 @@ function datLongTopic(key)
     }
 
     var from = xdata[key][XD_FROM];
-    var fromp = (from == "Unknown")
-                   ? "<p class='from'>Unknown Origin</p>"
-                   : "<p class='from'>" + from + "</p>";
+    var fromp;
+    if (from == "Unknown") {
+        fromp = "<p class='from'>Unknown Origin</p>";
+    }
+    else if (from == "ACL2 Sources") {
+        // link to main ACL2 sources dir:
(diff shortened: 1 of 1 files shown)


ericwhitmansmith:
I suppose a concern with this change could be that old versions of the manual will link to the bleeding edge versions of the files on GitHub, which may have been changed or even moved in the meantime.  But I think most serious users use the bleeding edge version of the manual, which should not diverge much from the bleeding edge versions of the files.  And I think that when someone clicks a link and ends up at GitHub, they will not be surprised to see the latest versions of the files (as is standard when browsing GitHub).

If desired, we could look into suppressing the links to GitHub when building the manual for a release.

MattKaufmann:
As far as I know, nobody really maintains the web-based manual generation since @jaredcdavis has been busy with other things.  I don't mind accepting this pull request (without understanding the code), but @solswords, do you want to take a look first?

solswords:
I think this is a nice idea. I suppose the links could be fixed to the version from which the manual was derived but I think always pointing to the current master branch is probably a better choice.

MattKaufmann:
Since @solswords commented without objecting, I plan to accept this pull request tomorrow if nobody does so or objects before then.

PR #1251, opened 2021-04-27
Restructure and extend fty::defomap to allow it to be used in deftypes

westfold:
Allows recursive map types.
Adds useful theorems to map book.


solswords (approved):
There are lots of tests so it seems clear that it at least generally works, and I don't see any problems on a cursory read of the changes.

books/kestrel/utilities/omaps/core.lisp
```diff
@@ -153,7 +156,12 @@
   (defrule alistp-when-mapp
     (implies (mapp x)
              (alistp x))
-    :rule-classes (:rewrite :forward-chaining)))
+    :rule-classes (:rewrite :forward-chaining))
+
+  (defrule consp-when-non-empty-mapp
```
ericwhitmansmith, line 161: This looks like it should perhaps be disabled or rephrased (maybe as a :forward-chaining rule or with a backchain limit), since it may be tried a lot and introduces map reasoning out of nowhere.  Also, the name should probably mention 'car'.

books/kestrel/utilities/omaps/core.lisp
```diff
@@ -13,6 +13,9 @@
 (include-book "std/osets/top" :dir :system)
 (include-book "std/util/define" :dir :system)
 (include-book "std/util/defrule" :dir :system)
+(include-book "misc/total-order" :dir :system)
```
ericwhitmansmith, line 16: Can any of these new include-books be made local?

ericwhitmansmith (commented):
Added a couple of questions, in case they are helpful.

PR #1252, opened 2021-04-29
Fix package problem in building pretty-program. Cleanup core.lisp adds

westfold:


airbornemihir:
I've merged this on the command line and will push when properly regressed. Others, of course, are free to pre-empt me.

PR #1256, opened 2021-05-04
Tonelli-Shanks interface functions and proof of correctness

jagadishb09:


books/kestrel/number-theory/tonelli-shanks-proof.lisp
```diff
@@ -0,0 +1,1220 @@
...
+    (defthm lrs-aux-lemma-2
+      (implies (and (natp tt^2^i)
+                    (posp i)
+                    (natp m)
+                    (natp p)
+                    (< 2 p)
+                    (< i m)
+                    (equal (least-repeated-square-aux i tt^2^i m p) 0))
+               (equal (least-repeated-square-aux (- m 1) tt^2^i m p) 0))
+      :hints (("goal"
```
bendyarm, line 131: Does lowercase "goal" work?  There are a few instances of this.

jagadishb09: Yes, it works. I was to do make regression and certify all the files, and I have confirmed that it works with Stephen and Alessandro in one of Friday's meetings.

bendyarm: It seems to work with "goal" or "Goal", but not if I take out the hint.  So I guess lowercase "goal" does work!

jagadishb09: That's good to know. Thanks.

MattKaufmann: Yes, case-insensitivity of the goal-spec is documented behavior.  See
this in :doc hints 

    where [goal-spec] is as described elsewhere (see [goal-spec]) and the

and then see this in :doc goal-spec:

    The matching algorithm is case-insensitive.

PR #1257, opened 2021-05-07
Minor xdoc fix for defflexsum

ericwhitmansmith:
This pull request addresses a minor xdoc issue that produces warnings like this:

WARNING: in topic AIGNET::AXI-LIT, subtopic order mentions topics that
are not children: AIGNET::AXI-LIT.

Note that the topic is being listed among its own children, which I think never makes sense.  Disclaimer: I don't know much about defflexsum, but the call (member x.name type-names) a few lines above my change indicates that this situation can happen.

With this change, the related warnings go away.

### DIFF
--- a/books/centaur/fty/docgen.lisp
+++ b/books/centaur/fty/docgen.lisp
@@ -542,8 +542,10 @@ binder.</p>")
     (mv (append (and (not sum-name-shared-with-prod-name) main-doc)
                 prods-doc
                 `((xdoc::order-subtopics ,x.name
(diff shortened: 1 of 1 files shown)


solswords (approved):
I think this is a good idea, but I'd apply the remove to the whole list, because occasionally someone wants a type whose name is the same as the pred.

ericwhitmansmith:
Thanks, Sol.  I took your suggestion.  If you're happy with this now, please merge the PR.

PR #1259, opened 2021-05-12
[xdoc] Fix defpointer to allow pointing to topics with unusual names.

ericwhitmansmith:
This pull request fixes a problem that occurs when defpointer is used to point to an xdoc topic whose name is an unusual symbol, such as a symbol that includes spaces.  Without this patch, the following gives an error:

(include-book "xdoc/topics" :dir :system)
(defxdoc |foo bar| :short "example")
(defpointer my-pointer |foo bar|)
:doc my-pointer

The error message is:
; xdoc error in MY-POINTER: 
"In SEE directive, expected ) after FOO. Near See @(see acl2::foo bar)..".
ACL2::MY-POINTER -- Current Interactive Session

Note the space between foo and bar in the generated @(see ...) form; this is not suitable for reading back in the symbol |foo bar|.

### DIFF
--- a/books/xdoc/top.lisp
+++ b/books/xdoc/top.lisp
@@ -504,17 +504,30 @@
       (concatenate-symbol-names (list ,@args))
       mksym-package-symbol)))
 
+;; Convert SYM to a string in such a way that it could be read
+;; back in as SYM.
+(defund symbol-as-readable-string (sym)
+  (declare (xargs :guard (symbolp sym)
+                  :mode :program))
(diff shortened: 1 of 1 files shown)


MattKaufmann:
Have you considered whether it's really OK not to print the package prefix explicitly?  It looks like you assume that the environment where fmt1 prints the symbol deals properly with a possible package prefix.  And maybe that's the case, but it seems worth checking whether you've thought this through (I haven't).

PR #1269, opened 2021-06-07
[JSON] parse \uXXXX escapes

bendyarm:
In JSON, Unicode U+0000 through U+001F in strings must be escaped.  For the points without simple escapes, they will be serialized as \uXXXX.  This PR makes the ACL2 JSON file reader understand those and return a corresponding string in UTF-8.

### DIFF
--- a/books/kestrel/json-parser/parse-json.lisp
+++ b/books/kestrel/json-parser/parse-json.lisp
@@ -15,8 +15,11 @@
 
 ;; Written from http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf.
 
-;; TODO: Add support for Unicode.  Currently, this uses ACL2 characters, of
-;; which there are only 256.
+;; TODO: Add more complete support for Unicode.  Currently, this uses ACL2 characters, of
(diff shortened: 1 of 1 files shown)


books/kestrel/json-parser/parse-json.lisp
```diff
@@ -160,10 +163,70 @@
...
+(defund hex-digit-to-nat (d)
+  (declare (xargs :guard (and (characterp d) (hex-digitp d))))
+  (let ((code (char-code d)))
+    (cond ((and (<= 48 code) (<= code 57))
+           (- code 48))
+          ((and (<= 65 code) (<= code 70))
+           (- code 55))
+          ((and (<= 97 code) (<= code 102))
+           (- code 87))
+          (t #x110000)))) ; an error value that should never happen
```
acoglio, line 175: Since this should be unreachable under the guard, you could consider returning `(prog2$ (impossible) 0)`, so you get a guard obligation from `(impossible)` and you still return an (irrelevant) natural number so this function unconditionally returns a natural number.

acoglio (approved):
Looks fine to me.

bendyarm:
I will make the change and resubmit a new PR.

PR #1271, opened 2021-06-09
[JSON] parse \uXXXX escapes

bendyarm:
In JSON, Unicode U+0000 through U+001F in strings must be escaped. For the points without simple escapes, they will be serialized as \uXXXX. This PR makes the ACL2 JSON file reader understand those and return a corresponding string in UTF-8.


acoglio (approved):
Looks fine to me.

bendyarm:
Closing.  The problem was fixed by commit https://github.com/acl2/acl2/commit/df035fa08bfcd30b74558f1f5e8e5f37efad426f

PR #1278, opened 2021-07-09
coi xdoc updates

TheBeaNerd:
Repaired orphan xdoc topics with links to miscellaneous (for now).


MattKaufmann:
Thanks!  It will be good to eliminate those warnings..

Might it make sense for COI-DEBUG::ASSERT and COI-DEBUG::FAIL to have ERRORS as a parent?

Might it be good for ADVISER::ADVISER to have COMPUTED-HINTS as a parent?

If you like these ideas, feel free to ask me to make those changes in master later.  But for now I'll simply accept the pull request.

PR #1280, opened 2021-07-12
[centaur] Reduce dependencies of tshell.

ericwhitmansmith:
Tshell currently depends (transitively, including via local includes) on 200 books.  Many of these seem unnecessary, such as IHS and osets and much more.  This change cuts the dependencies down to 24 books.

The comment "used in the raw code" on an include-book that I commented out gave me pause, but it doesn't seem to be true.  Everything still seems to build (after I added a few include-books in other places).

We have a use case where it would be very nice to have tshell depend on less stuff, because it supports an application we want to package up with Docker, and we want to certify and ship less stuff.

### DIFF
--- a/books/centaur/aignet/abc.lisp
+++ b/books/centaur/aignet/abc.lisp
@@ -32,6 +32,7 @@
 
 (include-book "aiger")
 (include-book "centaur/misc/tshell" :dir :system)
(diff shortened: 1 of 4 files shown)


solswords:
It looks like tshell-raw.lsp used to use some functions from the `str` package before Jared changed it to be based on the Quicklisp shellpool package, which explains that comment.

PR #1281, opened 2021-07-12
Add stp support

ericwhitmansmith:
This pull request adds support for STP as an external tool.  Books that depend on it should include:

; cert_param: (uses-stp)

to be excluded from the build.

I mostly just copied what was done for existing tools, without really understanding the mechanisms, but things seem to work.

I did not change books/build/shell.nix, because I don't know what that is.  Perhaps @kini could take a look.


MattKaufmann:
I took a quick look at the diffs and didn't see anything amiss, so I'm accept this pull request.

PR #1282, opened 2021-07-13
Remove manual building from books/doc/top-slow.

ericwhitmansmith:
This repurposes books/doc/top-slow as solely a top-level book to check for name conflicts among community books.  Manual building is now taken care of by books/doc/top.

The material removed in this commit was almost identical to the analogous material in books/doc/top.

In a future commit, we can move/rename this file to be books/top.lisp.


MattKaufmann:
Has the job of copying stuff from top-slow.lisp to top.lisp been truly completed?  For example, I see the form `(defpointer assocs patbind-assocs)` in top-slow.lisp, but there is no :doc topic for `patbind-assocs`.  There may be other such deletions to consider, too.

ericwhitmansmith:
> Has the job of copying stuff from top-slow.lisp to top.lisp been truly completed? For example, I see the form `(defpointer assocs patbind-assocs)` in top-slow.lisp, but there is no :doc topic for `patbind-assocs`. There may be other such deletions to consider, too.

Hi Matt, I'm not really sure, but I would assume that doc/top.lisp contains whatever people want to be in the manual, since for some time it's been the manual that is built by the regression.  And I believe you use doc/top for uploading to the web, yes?

There are some differences between doc/top and doc/top-slow, especially regarding Centaur books that are mostly unfamiliar to me.  Some of these differences may be intentional, due to the different roles of the two books.  For example, doc/top.lisp includes centaur/aignet/top-doc whereas doc/top-slow.lisp includes centaur/aignet/top.  Presumably the former is what is intended for inclusion in the manual, whereas the latter is more appropriate for checking for conflicts.

Another data point, in the certification output, I see:
For top-slow: ; Saving JSON files for 41635 topics.
For top: ; Saving JSON files for 41562 topics
So there are about 70 topics in top-slow that are not in top.  Maybe we can figure out which topics they are.

Regarding patbind-assocs, I'm confused because I see the same defpointer in both books, and :doc patbind-assocs works after including doc/top.

MattKaufmann:
I think this is fine, so I'll accept the pull request.  (Sorry, I was wrong about patbind-assoc being missing in the manual.)  Eric and I are exchanging email about possible further steps to take.

PR #1283, opened 2021-07-21
Added directory books/projects/arm/second

russinoff:
Added directory "books/projects/arm/second/".


MattKaufmann:
This is to master so probably should be closed, not accepted.   I've emailed Russinoff about this.

russinoff:
I changed the label from "master" to "testing" when I began the process.
When I first tried to submit the request, I was told that I was
unauthorized.  It seemed that I needed to make that chamge to "testing" one
more time.  When I did that and resubmitted the request, I thought it was
happy.  Did that second request not go through?

Issue #1293, opened 2021-08-08
with-output :off :all permits time$ output


bendyarm:
Consider:
```
ACL2 !>(with-output :off :all (defconst *foo1* 1))
 *FOO1*
ACL2 !>(with-output :off :all (time$ (defconst *foo2* 1)))
; (EV-REC *RETURN-LAST-ARG3* ...) took 
; 0.00 seconds realtime, 0.00 seconds runtime
; (5,616 bytes allocated).
 *FOO2*
```
According to
https://www.cs.utexas.edu/users/moore/acl2/manuals/latest/index.html?topic=ACL2____WITH-OUTPUT
`with-output :off :all` should

> Turn off all output during evaluation

but it fails to turn off the output from `time$`.

The place where this comes up for me is not with my use of `time$`, but I am writing `defconsts` forms into a book and then calling `(with-output :off :all (certify-book ...))` which calls `time$`, resulting in unwanted output.

MattKaufmann:
Here's a two-part proposal for dealing with this Issue.  Comments are welcome.

(1) Fix :DOC with-output to replace mention of "all output" by something like "all output that can be controlled by set-inhibit-output-lst".

(2) Add a new keyword to with-output, :all!, that causes all output to be turned off (not just output controllable by set-inhibit-output-lst).  Note that this includes goal names, so :gag-mode :goals is effectively coerced to :gag-mode nil.

Implementation Notes. `(With-output :off :all! <form>)` will macroexpand in raw Lisp to `(let ((*standard-output* (make-broadcast-stream))) <form>)`.  Probably it will be implemented with return-last, in part so that top-level evaluation can do the right thing (as implemented by ev-rec and especially ev-rec-return-last).
Oh, while that works for printing, it doesn't quite work for `time$` (really, `time`) in particular!  Consider, in CCL:

    ? (let ((*standard-output* (make-broadcast-stream)))
        (time (* 3 4)))
    (* 3 4)
    took 4 microseconds (0.000004 seconds) to run.
    During that period, and with 16 available CPU cores,
         5 microseconds (0.000005 seconds) were spent in user mode
         3 microseconds (0.000003 seconds) were spent in system mode
    12
    ? 

So probably I'll bind another special variable and change time$ to be sensitive
to it.  End of Implementation Notes.

[...]

PR #1295, opened 2021-08-14
Fix acl2s output after with-output change

mister-walter:
A recent change to with-output (709d5a5526c38f47a3bba245578fb06cf7b01392) results in `cw` output now being hidden
by `(with-output :off :all ...`. I have enabled the `comment` output
type in several `with-output` calls in ACL2s to restore the previous
behavior.

### DIFF
--- a/books/acl2s/cgen/cgen-rules.lisp
+++ b/books/acl2s/cgen/cgen-rules.lisp
@@ -38,7 +38,7 @@ date created: [2016-03-22 Tue]
                    (and lst (cadr lst))))
        (keys (defdata::remove-keywords-from-args '(:verbose) keys)))
     `(with-output 
-      ,@(and (not verbosep) '(:off :all :on error)) :stack :push
+      ,@(and (not verbosep) '(:off :all :on (error comment))) :stack :push
       (make-event
        (cgen-rule-events ',name ',keys ',verbosep (w state) state)))))
 
--- a/books/acl2s/cgen/top.lisp
+++ b/books/acl2s/cgen/top.lisp
@@ -156,6 +156,7 @@
     `(with-output
       :stack :push
       ,(if debug :on :off) :all
+      ,@(if debug nil (list :on 'comment))
       :gag-mode ,(not debug)
      (make-event
       (test?-fn ',form ',hints ',kwd-val-lst state)))))
--- a/books/acl2s/check-equal.lisp
+++ b/books/acl2s/check-equal.lisp
@@ -135,12 +135,12 @@
 
 (defmacro check (form)
   `(with-output
-    :off :all
+    :off :all :on comment
     (make-event (fcheck ',form state) :check-expansion t)))
 
 (defmacro check= (form1 form2)
   `(with-output
-    :off :all
+    :off :all :on comment
     (make-event (fcheck= ',form1 ',form2 state) :check-expansion t)))
 
 
--- a/books/acl2s/defdata/defdata-attach.lisp
+++ b/books/acl2s/defdata/defdata-attach.lisp
@@ -106,7 +106,7 @@ data last modified: [2015-06-09 Tue]
        (keys (remove-keywords-from-args '(:verbose :override-ok) keys)))
 
   `(with-output
-    ,@(and (not verbosep) '(:off :all)) :stack :push
+    ,@(and (not verbosep) '(:off :all :on comment)) :stack :push
     (make-event
         (cons 'progn
               (defdata-attach-fn ',name ',keys ',verbosep ',override-ok-p (w state)))))))
--- a/books/acl2s/defdata/defdata-core.lisp
+++ b/books/acl2s/defdata/defdata-core.lisp
@@ -977,12 +977,12 @@ Example use
                    (and lst (cadr lst)))))
     `(with-output
       ,@(and (not verbosep)
-             '(:off :all :on (summary error) :summary-off (:other-than acl2::form acl2::time)))
+             '(:off :all :on (summary error comment) :summary-off (:other-than acl2::form acl2::time)))
       :gag-mode t :stack :push
       (encapsulate
        nil
        (with-output
-        ,@(and (not verbosep) '(:off :all))
+        ,@(and (not verbosep) '(:off :all :on comment))
         :gag-mode t
         :stack :push
         (make-event
--- a/books/acl2s/defdata/defdata-core.lisp
+++ b/books/acl2s/defdata/defdata-core.lisp
@@ -371,7 +371,7 @@ Does not seem to be used.
 (defun defdata-core-events (a1 wrld)
   (b* (((list D kwd-alist) a1)) ;a1 is the result of parse-defdata
     `(WITH-OUTPUT
-      :ON (SUMMARY ERROR) :OFF (PROVE EVENT OBSERVATION)
+      :ON (SUMMARY ERROR COMMENT) :OFF (PROVE EVENT OBSERVATION)
       :SUMMARY-OFF (:OTHER-THAN ACL2::FORM ACL2::TIME)
       (PROGN
        ,@(collect-events :pre-events D kwd-alist)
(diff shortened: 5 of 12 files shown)


MattKaufmann:
Most likely, tomorrow I'll be adding utilities cw+ and fmt-to-comment-window+ that are like cw and fmt-to-comment-window except that they do not inhibit COMMENT output.  Would changing some cw calls to cw+ be a simpler solution for you?

bendyarm:
@MattKaufmann  Does someone need `cw+`?  It seems to me that is just pushing the problem to another place.  Then someone who wanted to inhibit `cw+` output would need another indicator for that, etc.

Another possibility that is fine with me would be to change back `with-output .. :all` so that the `:all` means "all except COMMENT, and to add `:all+` to mean "all including COMMENT".  That would be more backward-compatible (to before commit 709d5a5), and I could easily change my instances of `:all` (which are not yet in the community books, anyway) to `:all+`.  This change could be in addition to or instead of this PR, depending on @mister-walter 's preferences.

mister-walter:
I'm inclined to agree with @bendyarm here, it didn't really make a whole lot of sense that `cw` output was not silenced by `with-output :off :all` previously, and adding something like `cw+` is likely to result in someone asking for a way to disable all output including `cw+` at some point.

I'm ambivalent as to whether the current behavior is kept, or the behavior of `with-output :off :all` is reverted and a `:all+` argument is added. I would wait and see what @pmanolios says, since he knows the ACL2s output code better than I. (in fact, he should probably be a reviewer for this PR)

If the current behavior is kept, it may be somewhat labor-intensive to determine which community books are affected - one must audit each `with-output :off :all` call and determine whether anything called inside of it ever calls `cw` or `fmt-to-comment-window`, keeping in mind the `with-output` stack, nested `with-output`s, etc...

As a side note, adding some "output testing" to ACL2's test suite (e.g. having a way to assert that ACL2's output after evaluating an event is of a particular form) might be useful, allowing one to catch breakage like that which occurred with ACL2s in an automated fashion.

MattKaufmann:
The change I mentioned earlier in this thread is now in GitHub (commit 766509edfba40c8e32a361a6d2016e539a6ceb1e).

Regarding posting from @bendyarm.  "Does someone need cw+": I used it to fix an issue with er-soft+ in that commit.

Regarding posting from @mister-walter.  You might be right that "adding something like cw+ is likely to result in someone asking for a way to disable all output including cw+ at some point."  However, my response would be: "Do you want to disablle output from fmt too?  How about output from print-object$?"  There will likely always be printing utilities that can't be silenced; the idea of output types was to provide a way to have silenceable output.  Now, cw is silenceable too.  I agree that in principle, "it may be somewhat labor-intensive to determine which community books are affected."  However, I think that this can be dealt with easily on a case-by-case basis as necessary, and in fact I think many will be happy to see cw output silenced.

But the bottom line is that while I agree that a different design is possible, including the notion from @bendyarm about adding :all+ and excluding COMMENT from :all, nevertheless I think the current design is reasonable (and I actually like it, for what it's worth).  Importantly -- and this is key -- the design has already been discussed publicly in Issue #1293, where a lengthy discussion converged.  That convergence resulted in this morning's commit, which including reworking :doc.  So I'm not inclined to revisit this.  However, I'll follow up with the funders of this work.

Oh, and regarding "adding some "output testing" to ACL2's test suite (...) might be useful: yes, there are several books like that; if you run the command

    fgrep --include='*.acl2' -ri run-script .

in the books/ directory you'll see about 20 such uses of run-script.  I'd be happy for people to add more such books.  In particular, @mister-walter, please feel free to add such books for ACL2s (see those .acl2 files and/or :DOC run-script for help).

mister-walter:
Thanks Matt! I'll follow up with @pmanolios and see if he prefers using `cw+`/`fmt-to-comment-window` or adding `comment` to `with-output` where needed. And thanks for the pointer to `run-script`, I'll look into that! I'll mark this as draft in the meantime.

acoglio:
My vote is to:
* Not have `:all` suppress everything but comments and `:all!` suppress comments, because to me 'all' means all, so to speak. (It's reminiscent to the different between the older targets `make everything` and `make all`.) While it may be reasonable to assign more specific technical meanings to English words if the situation warrants it, my tendency is to avoid that as much as possible.
* Not introduce `cw+`, for similar reasons. It's a 'comment-plus' vs. just a 'comment', and comment-pluses are not part of 'all'. As @bendyarm points out, it's somewhat "moving" the problem elsewhere.

I realize that this may involve changing certain community books. For instance, my utility [make-event-terse](https://www.cs.utexas.edu/users/moore/acl2/manuals/latest/index.html?topic=ACL2____MAKE-EVENT-TERSE) now implicitly turns off comment output, so all my event macros that use `cw` to print results no longer print them. So I plan to change `make-event-terse` to leave comments on.

Regarding the issue of [er-soft+](https://www.cs.utexas.edu/users/moore/acl2/manuals/latest/index.html?topic=ACL2____ER-SOFT_B2), I'm surprised that the printed message is under comment input vs. error input. (If this was discussed at our recent meeting and I missed/forgot it, apologies.) Maybe there are technical reasons why those error messages have to be comments so that `er-soft+` can be in logic mode, but again I would think that it should be error output.

MattKaufmann:
I think that user-interface issues like this tend to be more complicated than may at first appear; hence this unfortunately long reply.  In short, it looks to me like this discussion could be reaching closure, without further action to be taken.  I'll respond below to comments not yet addressed.

I agree with @acoglio's preference for :all to mean "all suppressible output is suppressed".  If there becomes a clear need for :all+ I might change my mind, but my sense is that :all+ could open up issues that aren't obvious at first (such as, should :all+ suppress output from raw Lisp, and is that feasible while allowing other output -- there's related discussion in Issue #1293 that I'd prefer not to revisit).

> Regarding the issue of er-soft+, I'm surprised that the printed message is under comment input vs. error input.

Er-soft+ is now controlled by error output, not comment output.

Regarding the use of some flavor of cw in support of er-soft+, I think that the explanation from @acoglio is correct: keeping things in :logic mode.  Formatted printing going to state (as opposed to going to the comment window) involves fmt and such functions, which I believe would take considerable work to put into :logic mode, especially since built-in :logic-mode functions must be guard-verified.  The utility error-fms-soft-logic (defined in books/tools/er-soft-logic.lisp) gets around that problem by using comment-window printing, carefully crafted to be sensitive to whether error output is suppressed.  After the addition of the COMMENT output type, I changed the implementation of error-fms-soft-logic so that it uses fmt-to-comment-window+ instead of fmt-to-comment-window, so that printing is guarded by suppression only of ERROR output, not COMMENT output.

As for cw+ and other "+" utilities, I see them as analogous to print-object$, fmt (and fms etc.), and princ$, none of which can be suppressed using with-output etc.  I don't see a good reason to delete cw+, fmt-to-comment-window+, and the "!+" versions, especially given that fmt-to-comment-window+ supports er-soft+ as described above.  And by the way, in general we can't simulate (cw+ ..) by (with-output :off comment (cw ..)), because with-output may only be used with error-triples.

That said, I have no objection to replacing the use of the "+" suffix by a better naming convention.  If the community gets to agreement on alternate names then I'd be happy to make the changes.  (But I'm part of the community and I'll say up front that I object to replacing "+" by "!", since that would present a new level of potential confusion, as cw! has long been defined to print so that output can be read back in.)

acoglio:
@MattKaufmann That all makes sense, thanks.

Since you mention names, it sounds to me like, in a way, `fmt-to-comment-windows+` could be `fmt-to-error-window` since it's controlled by `error` output, while `fmt-to-comment-window` is controlled by `comment` output. I suppose that then `cw+` could be `ew`. But maybe the situation is less "symmetric" than I think.

MattKaufmann:
@acoglio Actually, fmt-to-comment-window+ is controlled by nothing; it prints unconditionally.  Error output control is handled explicitly in the utility error-fms-soft-logic (defined in books/tools/er-soft-logic.lisp) that supports er-soft+, as follows:

    (defun error-fms-soft-logic (ctx str alist state)
      (declare ...)
      (and ...
	   (not (member-eq 'error (f-get-global 'inhibit-output-lst state)))
	   (fmt-to-comment-window+
	    ....

pmanolios:
I'm coming in at the tail end of this discussion, so I'm not going to make any suggestions and will just update ACL2s once this is settled. Thanks to @MattKaufmann , @bendyarm , @mister-walter and @acoglio for working to improve ACL2 output generation.

PR #1296, opened 2021-08-15
Remove "/usr/bin/env" in smtlink config

kini:
/usr/bin/env is not available in Nix build environments, and its use
in the smtlink default configs seems to be redundant.


pennyan:
Hi @kini , sorry for the late reply. It looks good to me.

pennyan (approved):
LGTM

kini:
Thanks!

PR #1297, opened 2021-08-18
Small fix to CCL Mac installation instructions.

acoglio:
Mention the Mac executable, not the Linux one.


MattKaufmann (approved):
Good catches; thanks for the fix!

PR #1299, opened 2021-08-20
Fix typo in SET-EVISC-TUPLE doc.

acoglio:


MattKaufmann (approved):
Thanks for catching that typo.

PR #1303, opened 2021-09-02
Remove duplicate entry in x86isa publications list

kazarmy:
I'm not sure whether the repetition was for emphasis, but normally afaik such lists don't have duplicates.


kazarmy:
Sorry didn't read that closely enough, see #1304.

PR #1305, opened 2021-09-02
[FTY] Tweak a call of translate-cmp.

ericwhitmansmith:
This calls translate-cmp with logic-modep=nil, to prevent an error when the form involved contains :program mode functions.

This is related to deffixequiv and seems needed to allow a define with :hooks (:fix) to work when submitted in :program mode.  My head's not in this machinery, and perhaps there is a legitimate reason to ban this.  But I suspect this use case was simply not considered.  Ideally, tools like define would still work even in :program mode (but proofs would be skipped, of course).

### DIFF
--- a/books/centaur/fty/fixtype.lisp
+++ b/books/centaur/fty/fixtype.lisp
@@ -293,7 +293,7 @@
                       (concatenate
                        'string (symbol-name arg) "-EQUIV")
                       pkg)))
-       ((mv err tr-form) (acl2::translate-cmp form t t nil 'deffixequiv-basic-parse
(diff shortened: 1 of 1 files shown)


solswords:
I was confused at first but I think this is a good idea.  It's going to be an error if we actually submit a defthm (in logic mode) with program mode names, but there's no need to catch that error early -- ACL2 will provide a better error message for that than I would have the energy to write.  OTOH, if we're in program mode (which I think is how define works -- it puts a `(program)` event at the beginning of its encapsulate rather than adding a `:mode :program` declaration) then the defthm will just be skipped without error.  So LGTM.

ericwhitmansmith:
Ok for me to I merge this now?

PR #1306, opened 2021-09-07
Some changes to improve multiplier verification library (projects/rp-rewriter/lib/mult3)

temelmertcan:
I only recertified the books that I think would be affected by these commits, and I did not run a whole regression. If that is a problem, please do not accept this request and I can run a complete regression.


MattKaufmann:
I don't know either whether it's OK to accept, but I think so -- you're a careful guy and you clearly did a bunch of testing, and you're merging into testing-centaur, not master.  So I think it's OK and I'll accept this pull request.  If not, then someone should scold me and I'll try not to mess up again!

PR #1308, opened 2021-09-09
[FTY] Reduce consing in recognizers for defprod, etc.

ericwhitmansmith:
This avoids executing strip-cars every time a structure created by defprod is checked.  Profiling showed that the call to strip-cars was allocating the majority of the memory in a large proof that checked well-formedness of many (tens of millions) of small structures created with defprod.  This change also reduces the number of passes through the alist.  This change uses MBE and so effects execution only, not the logical definitions generated.

Feel free to use all, some or none of this implementation, to change the style as desired, etc.


solswords:
LGTM. I think I was lax in thinking about efficiency of recognizers in general in FTY, but they do get called a lot in some proofs.

ericwhitmansmith:
Ok if i merge this now?

Issue #1309, opened 2021-09-12
switch from logic mode to program mode causes certification failure


bendyarm:
In most cases, if I take a logic mode file and add `(program)` at the top,
it certifies just fine, since the theorems are generally just ignored.

However, if I have a `defines` where the `define` components have a `:returns` type specifier,
and it certifies in logic mode, adding `(program)` causes certification to fail.
It seems to me like this is a bug, since it appears the intent is that adding `(program)`
should ignore proof obligations as it does in other contexts.

Another reason why it seems like a bug is if I leave the file in `(program)` mode
and I get rid of the `:returns` type declarations, the file certifies and works fine.
It seems like `:returns` type declarations should be ignored in program mode.

I have a workaround (a suggestion from Alessandro Coglio) where I put the return type
in string quotes.  This makes it easier to switch back to logic mode when I am ready
to do so.

Here's a simple example, simplified from the first example in the xdoc for Defines:
https://www.cs.utexas.edu/users/moore/acl2/manuals/latest/index.html?topic=ACL2____DEFINES
```
(in-package "STD")

(include-book "std/util/defines" :dir :system)
(include-book "std/util/deflist-base" :dir :system)

(program)

(defines terms

    (define my-termp (x)
      :returns (yes/no booleanp)
      (if (atom x)
          (natp x)
        (and (symbolp (car x))
             (my-term-listp (cdr x)))))

    (define my-term-listp (x)
      :returns (yes/no booleanp)
      (if (atom x)
          t
        (and (my-termp (car x))
             (my-term-listp (cdr x)))))

    )
```

ericwhitmansmith:
I pushed a candidate fix in:
https://github.com/acl2/acl2/commit/1dbb1a74137a7b5c1eb283f70de85d90f3f5e7a0
I'm still testing it locally, after which I'll make a pull request.

bendyarm:
Fixed by PR #1310.  Thanks!  Closing.

PR #1310, opened 2021-09-13
[DEFINES] Suppress return-value theorems in :program mode.

ericwhitmansmith:
Fixes issue #1309.  An existing code comment mentioned this issue but seems wrongheaded, as exemplified by the issue filed by Eric McCarthy.  It seems perfectly reasonable to supply return value types in :program mode, either simply as documentation or because some code is in :program mode only temporarily.  In that case, this commit ensures that defines will not fail.

Normally, generated theorems would be simply skipped in :program mode.  The failure here is caused by the use of defthm-flag to prove them, since the corresponding make-flag is skipped in :program mode.

Standard disclaimers apply: My head is not really in this code.

### DIFF
--- a/books/std/util/defines.lisp
+++ b/books/std/util/defines.lisp
@@ -862,13 +862,13 @@ encapsulate), and is mainly meant as a tool for macro developers.</dd>
 
        (defsection rest-events-1
          ,@(and want-xdoc-p `(:extension ,name))
-         ;; It seems OK to do this even if we're in program mode, because in that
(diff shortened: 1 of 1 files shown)


solswords:
This seems good to me.  Ready to merge?

bendyarm:
Looks good to me.

ericwhitmansmith:
Sol, please merge if you are happy with it.  Or I can, if you want.  Thanks!

PR #1312, opened 2021-09-13
add function that parses and returns json and at the same time runs s…

bendyarm:
…ha-256 on the bytes and returns that as well


ericwhitmansmith:
I would prefer not to have the json stuff depend on sha-256.  But I've refactored parse-file-as-json so that parse-file-as-json-aux now returns the parsed json and the characters read from the file (with which you can then do whatever you like).  Is that enough to give you what you want?

bendyarm:
Thanks, Eric!  Closing.

PR #1313, opened 2021-09-15
Delete questionable files ending in *.bak and *.orig

lucasmitrak:
Currently, after `make basic` is run and the books directory is ready
for installation, five files ending in *.bak or *.orig remain which
should not be included in the final installation. All five files reside
in the directory books/workshops/1999/ivy/ivy-v2/ivy-sources/ and
should instead be cleaned. This has already been done in Debian [1]
and Gentoo [2].

[1] https://debian.pkgs.org/9/debian-main-amd64/acl2-emacs_7.2dfsg-3_all.deb.html
[2] https://github.com/gentoo/sci/blob/a9a6dc87759ff69f93105d2b85b5cd90755380e5/sci-mathematics/acl2/acl2-8.4.ebuild#L31


MattKaufmann:
You say "after `make basic` is run", but actually those five files exist even before "`make basic`" is run, both in the released ACL2 8.4 and in the current github master.  I'm guessing that those files were contributed by the original authors for the 1999 ACL2 Workshop; unfortunately, one author is deceased, and the other has been away from the ACL2 community for some time as far as I know.  I think it might be best to leave these harmless files in place, then, for archival purposes or in case there's some subtle purpose for them.  Anyone who disagrees is welcome to take responsibility for accepting this pull request, as far as I'm concerned.

lucasmitrak:
Thank you for your response. I understand these files are not created during `make basic` and they were created a long time ago, my wording was just poor. I meant to imply they should not exist when the books directory is ready to be installed. However, I understand your reasoning for wanting to leave them. My intention was to submit the changes I made upstream on principle, so feel free to do what you wish with this PR.

PR #1316, opened 2021-09-24
[centaur] Fix various xdoc issues.

ericwhitmansmith:
I think these are all straightforward.


solswords (approved):
Wow, I really have a problem closing my `@('` delimiters, don't I.

PR #1318, opened 2021-10-01
Use $(MAKE) variable, not the explicit command name ‘make’

lucasmitrak:
Currently, when creating the target `doc` the dependency acl2-manual
is invoked. acl2-manual does not use the MAKE variable but instead
uses the make command directly. The target
books/system/doc/rendered-doc.lsp does the same. According to the
make manual, "recursive make commands should always use the variable
MAKE, not the explicit command name ‘make’" [1]. Since both of these
are recursive calls, the variable should be used instead.
This has already been patched and tested within the Gentoo science
repository [2]. Gentoo's package manager portage will throw a Q/A issue:
"make[1]: warning: jobserver unavailable: using -j1. Add '+' to parent
make rule" when compiling from source code if the patch is not applied.
This further suggests that this change is necessary.

[1] https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html
[2]
https://github.com/gentoo/sci/blob/master/sci-mathematics/acl2/files/acl2-use_make_variable.patch

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -606,7 +606,7 @@ doc/home-page.html: doc/home-page.lisp
 # xdoc::save that populates doc/manual/ (not under books/).
 acl2-manual: check-books
 	rm -rf doc/manual books/system/doc/acl2-manual.cert
-	cd books ; make USE_QUICKLISP=1 system/doc/acl2-manual.cert
+	cd books ; $(MAKE) USE_QUICKLISP=1 system/doc/acl2-manual.cert
 	rm -rf doc/manual/download/*
 
 # WARNING: The dependency list just below isn't complete, since it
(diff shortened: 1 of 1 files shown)


MattKaufmann:
Good catch!  Thanks; those inappropriate uses of "make" were an oversight, and I'll accept the pull request.  FYI, normally only changes under the books/ directory (often called the "community books") are made by other than me and J Moore, and this is checked by a script that I use, bin/purity.sh (invoked by bin/pull.sh).  If accepting this pull request causes a problem with that, though, I can fix it (by updating the value of SINCE in bin/purity.sh).  Anyhow, thanks again.

MattKaufmann:
Ah -- I see that since I'm the one accepting the pull request, the purity script doesn't complain.  Excellent.  That makes sense.  @lucasmitrak , thanks again.

MattKaufmann:
Sorry, I spoke too soon -- the purity script did catch this.  But I've updated it so all's well now.

PR #1319, opened 2021-10-06
Irrelevant formals fix

ericwhitmansmith:
Here is an example for which the old version gave what seems to be the wrong answer:

(irrelevant-formals-info
 '(mutual-recursion
   (defun even-natp (x irrelevant1 irrelevant2)
     (if (zp x)
         t
       (not (odd-natp (+ -1 x) irrelevant1 irrelevant2))))
   (defun odd-natp (x irrelevant3 irrelevant4)
     (if (zp x)
         nil
       (not (even-natp (+ -1 x) irrelevant3 irrelevant4))))))

It gave:
((EVEN-NATP . IRRELEVANT2)
 (ODD-NATP . IRRELEVANT4))

which is missing irrelevant1 and irrelevant3.  After this change, it gives:

((EVEN-NATP IRRELEVANT1 IRRELEVANT2)
 (ODD-NATP IRRELEVANT3 IRRELEVANT4))

### DIFF
--- a/books/std/system/irrelevant-formals-input.lsp
+++ b/books/std/system/irrelevant-formals-input.lsp
@@ -144,8 +144,8 @@
    (defun even-natp (x irrelevant1 irrelevant2)
      (if (zp x)
          t
-       (not (odd-natp (+ -1 x) irrelevant irrelevant2))))
-   (defun odd-natp (x irrelevant irrelevant2)
+       (not (odd-natp (+ -1 x) irrelevant1 irrelevant2))))
+   (defun odd-natp (x irrelevant3 irrelevant4)
      (if (zp x)
          nil
-       (not (even-natp (+ -1 x) irrelevant irrelevant2))))))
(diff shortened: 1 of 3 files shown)


MattKaufmann:
That looks fine to me (after a rather quick, but probably adequate, look).  I don't know why on lines 63 and 64 of  books/std/system/irrelevant-formals.lisp we find

                            (append (cdr (assoc-eq (car slot) alist))
                                    (list (cddr slot)))

rather than

                            (cons (cddr slot)
                                  (cdr (assoc-eq (car slot) alist)))

but I guess it's to preserve order.  I imagine the consing is minimal anyhow so this is probably fine.  (The alternative would be, I think, to call reverse at the end.)  Thanks for adding a test, which adds a lot of confidence.

ericwhitmansmith:
Yep, it's to preserve order.  (I originally wrote it the other way, but that listed the irrelevant params in the opposite order, so I switched it.)  Thanks for merging!

PR #1327, opened 2021-11-04
Move books/doc/top-slow.lisp to books/top.lisp.

ericwhitmansmith:
This was discussed on the acl2-books mailing list.  See also the comments at the top of books/top.lisp.

I believe, but am not positive, that I've updated the auto-termination stuff appropriately after this change.


MattKaufmann (approved):
This looks fine to me, though I note that for the auto-termination stuff, although the instructions and updating code were revised, it doesn't look like they were employed to make any actual updates to the database.  I imagine that will work, and if not, and if I'm the one to run into trouble, I might ask for help.

PR #1337, opened 2022-01-06
Make cert.pl use the useless runes files the same way as 'make' does.

bendyarm:
Prior to this if you use `cert.pl` to certify a book,
and if the environment variable `ACL2_USELESS_RUNES`
is not defined and exported, then `cert.pl` ignores
the useless runes files.

After this PR, `cert.pl` looks at `ACL2_USELESS_RUNES`
and if it is not defined, it sets it to -25, which is the default
for `make` in the community books directory.

One common use case that this speeds up a lot is when
you want to work on a particular book (whether that is in
the community books or not), and you want to certify all
of its dependencies without doing a full regression.
It is convenient to just do `cert.pl oracle` or some such command, 
but if there are any slow community books that need to be certified,
this could take longer than doing the full regression due to
`cert.pl` ignoring the useless runes files.

### DIFF
--- a/books/build/cert.pl
+++ b/books/build/cert.pl
@@ -123,6 +123,10 @@ my $make = $ENV{"MAKE"} || "make";
 my @make_args = ();
 my $acl2 = $ENV{"ACL2"};
 my $acl2_books = $ENV{"ACL2_SYSTEM_BOOKS"};
+# add default useless runes setting if undefined
+if (! defined($ENV{"ACL2_USELESS_RUNES"}) ) {
+    $ENV{"ACL2_USELESS_RUNES"} = "-25";
+}
 my $startjob = $ENV{"STARTJOB"};
 if (! $startjob ) { $startjob = "bash"; }
 my $keep_going = 0;


MattKaufmann (changes requested):
Suppose ACL2_USELESS_RUNES is set to the empty string as noted for example in :DOC verify-guards-for-system-functions: "ACL2_USELESS_RUNES= ".  Will this do the right thing, or does the new !defined check need to be disjoined with a check for the empty string?

Also, please update :doc build::cert.pl or an appropriate subtopic accordingly.

acoglio (approved):
This looks like an improvement to me. Thanks!

ericwhitmansmith:
Having `cert.pl` do the same thing as `make` wrt useless runes seems like a good idea to me.

solswords (approved):
I think this is fine. Arguably it could be done in only one place (make_cert or make_cert_help.pl or in ACL2 itself) rather than both here and in Makefile-features (via cert_features.lsp), but probably not worth untangling it all if this works.

bendyarm:
> Suppose ACL2_USELESS_RUNES is set to the empty string as noted for example in :DOC verify-guards-for-system-functions: "ACL2_USELESS_RUNES= ". Will this do the right thing, or does the new !defined check need to be disjoined with a check for the empty string?
> 
> Also, please update :doc build::cert.pl or an appropriate subtopic accordingly.

I believe it will "do the right thing", but the phrase
"because of how proofs differ" gives me pause.

Question: is the point of doing `ACL2_USELESS_RUNES= `
in the `make .. devel-check.cert` to turn off useless runes file processing?
By preventing the `export ACL2_USELESS_RUNES ?= -25` in `Makefile-features`
from setting it to -25?

If so, then I think this logic in cert.pl is the same.  Here is some evidence.
Take the following perl file, and put it in a temporary testing directory:
```
#!/usr/bin/env perl                                                                                                                                                                                                                                                                                                                                                                             

print "before:\n";
print "\$ENV{\"ACL2_USELESS_RUNES\"} exists: " . (exists $ENV{"ACL2_USELESS_RUNES"} ? "yes" : "no") . "\n";
print "\$ENV{\"ACL2_USELESS_RUNES\"} is defined: " . (defined $ENV{"ACL2_USELESS_RUNES"} ? "yes" : "no") . "\n";
if ( defined $ENV{"ACL2_USELESS_RUNES"} ) {
    print "Value is " . $ENV{"ACL2_USELESS_RUNES"} . "\n";
}

print $ENV{'ACL2_USELESS_RUNES'};
print "\n";

if (! defined($ENV{"ACL2_USELESS_RUNES"}) ) {
    $ENV{"ACL2_USELESS_RUNES"} = "-25";
}

print "after:\n";
print "\$ENV{\"ACL2_USELESS_RUNES\"} exists: " . (exists $ENV{"ACL2_USELESS_RUNES"} ? "yes" : "no") . "\n";
print "\$ENV{\"ACL2_USELESS_RUNES\"} is defined: " . (defined $ENV{"ACL2_USELESS_RUNES"} ? "yes" : "no") . "\n";
if ( defined $ENV{"ACL2_USELESS_RUNES"} ) {
    print "Value is " . $ENV{"ACL2_USELESS_RUNES"} . "\n\n";
}
```
Then run it with various values of `ACL2_USELESS_RUNES` to see what it does.  I get this:
```
$ unset ACL2_USELESS_RUNES                                                                                                                                                                                                                                                                                                     
$ perl tpl.pl                                                                                                                                                                                                                                                                                                                  
before:                                                                                                                                                                                                                                                                                                                        
$ENV{"ACL2_USELESS_RUNES"} exists: no                                                                                                                                                                                                                                                                                          
$ENV{"ACL2_USELESS_RUNES"} is defined: no                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                               
after:                                                                                                                                                                                                                                                                                                                         
$ENV{"ACL2_USELESS_RUNES"} exists: yes                                                                                                                                                                                                                                                                                         
$ENV{"ACL2_USELESS_RUNES"} is defined: yes                                                                                                                                                                                                                                                                                     
Value is -25          
                                                                                                                                                                                                                                                                                                         
$ ACL2_USELESS_RUNES= perl tpl.pl                                                                                                                                                                                                                                                                                              
before:                                                                                                                                                                                                                                                                                                                        
$ENV{"ACL2_USELESS_RUNES"} exists: yes                                                                                                                                                                                                                                                                                         
$ENV{"ACL2_USELESS_RUNES"} is defined: yes                                                                                                                                                                                                                                                                                     
Value is                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                               
after:                                                                                                                                                                                                                                                                                                                         
$ENV{"ACL2_USELESS_RUNES"} exists: yes                                                                                                                                                                                                                                                                                         
$ENV{"ACL2_USELESS_RUNES"} is defined: yes                                                                                                                                                                                                                                                                                     
Value is                                                                                                                                                                                                                                                                                                                       

$ ACL2_USELESS_RUNES=24 perl tpl.pl                                                                                                                                                                                                                                                                                            
before:                                                                                                                                                                                                                                                                                                                        
$ENV{"ACL2_USELESS_RUNES"} exists: yes                                                                                                                                                                                                                                                                                         
$ENV{"ACL2_USELESS_RUNES"} is defined: yes                                                                                                                                                                                                                                                                                     
Value is 24                                                                                                                                                                                                                                                                                                                    
24                                                                                                                                                                                                                                                                                                                             
after:                                                                                                                                                                                                                                                                                                                         
$ENV{"ACL2_USELESS_RUNES"} exists: yes                                                                                                                                                                                                                                                                                         
$ENV{"ACL2_USELESS_RUNES"} is defined: yes                                                                                                                                                                                                                                                                                     
Value is 24                                                                                                                                                                                                                                                                                                                    
$
```

MattKaufmann:
@bendyarm Sorry about any confusion.  Yes, I agree that your new code looks right (and, the answer to your question is "yes").  So all that's left is to update :doc build::cert.pl or an appropriate subtopic accordingly.

bendyarm:
There is still some discussion on Slack about how to handle this for ACL2(r) and maybe ACL2(p) so we should not merge this PR until we decide that and update the doc accordingly.

bendyarm:
Superseded by cert-heed-useless-runes-2; closing.

PR #1341, opened 2022-01-13
Cert heed useless runes 2

bendyarm:
This contains all the changes discussed on slack.
I did a regression and interactively tested ACL2(r) as well and it turns off useless runes there.
There is another branch containing the doc.

### DIFF
--- a/bin/new-useless-runes-files.sh
+++ b/bin/new-useless-runes-files.sh
@@ -65,7 +65,8 @@ echo '(defabbrev ur-fname (x)
     (if p
         (concatenate (quote string)
                      (subseq x 0 p)
-                     "/.sys"
+                     #-non-standard-analysis "/.sys"
+                     #+non-standard-analysis "/.sysr"
                      (subseq x p (length x))
                      "@useless-runes.lsp")
       (er hard? (quote ur-fname)
--- a/books/.gitignore
+++ b/books/.gitignore
@@ -1,3 +1,4 @@
+.sysr/
 workxxx*
 *@expansion.lsp
 *.out
--- a/books/build/cert.pl
+++ b/books/build/cert.pl
@@ -123,6 +123,10 @@ my $make = $ENV{"MAKE"} || "make";
 my @make_args = ();
 my $acl2 = $ENV{"ACL2"};
 my $acl2_books = $ENV{"ACL2_SYSTEM_BOOKS"};
+# add default useless runes setting if undefined
+if (! defined($ENV{"ACL2_USELESS_RUNES"}) ) {
+    $ENV{"ACL2_USELESS_RUNES"} = "-25";
+}
 my $startjob = $ENV{"STARTJOB"};
 if (! $startjob ) { $startjob = "bash"; }
 my $keep_going = 0;
--- a/books/build/cert_features.lsp
+++ b/books/build/cert_features.lsp
@@ -88,9 +88,10 @@
            (state (newline channel state))
            (state (princ$ "EXPORTED_VARS += ACL2_THINKS_BOOK_DIR_IS" channel state))
            (state (newline channel state))
-; Matt K. mod: It seems to be too much trouble to arrange that the
-; @useless-runes.lsp files be used for ACL2(r) and ACL2(p) in addition to ACL2.
-           #-(or acl2-par non-standard-analysis)
+; Eric M: although we default to ACL2_USELESS_RUNES=-25 now,
+; to keep ACL2(r) safe from incompatible useless runes files,
+; we now turn off useless runes for ACL2(r) in both
+; useless-runes-value and useless-runes-filename.
            (state (princ$ "export ACL2_USELESS_RUNES ?= -25" channel state))
            (state (newline channel state))
            (state (princ$ "EXPORTED_VARS += ACL2_USELESS_RUNES" channel state))
@@ -252,4 +253,3 @@
 (write-defrec-certdeps *defrecs* state)
 
 (good-bye 0)
-
--- a/other-events.lisp
+++ b/other-events.lisp
@@ -16121,6 +16121,8 @@
   ;; This is for extra safety in ACL2(r)
   ;; but it is probably not needed due to the
   ;; concurrent change to defun useless-runes-filename.
+  #+non-standard-analysis
+  (declare (ignore useless-runes-r/w useless-runes-r/w-p))
   (let ((useless-runes-r/w
          #+non-standard-analysis nil
          #-non-standard-analysis useless-runes-r/w
--- a/other-events.lisp
+++ b/other-events.lisp
@@ -16115,6 +16115,21 @@
 ; (We could allow 0 but that would mean the same as nil, which could perhaps
 ; lead to confusion somehow.)
 
+  ;; Ignore useless-runes info in ACL2(r),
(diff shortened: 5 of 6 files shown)


MattKaufmann:
Note that this pull request modifies ACL2 source files.  Please don't anyone accept this pull request (or #1342) until I've approved it in my review.  I'm in email communication with @bendyarm about this.

acoglio (approved):

MattKaufmann (commented):
(The following concern is only about ACL2(r), so I'm OK with this if
Ruben is.  But I hope it will be considered.)  It appears that the modified definition of useless-runes-value now causes the :useless-runes argument of certify-book to be ignored in ACL2(r).  Is that what's intended?  It seems to me better to avoid routine use of useless-runes in ACL2(r), not to prohibit it entirely if that's what somebody wants.  I can also imagine installing useless-runes for some specific ACL2(r) community books when that doesn't make them fail to certify and their certification is very slow otherwise (some books under books/projects/filesystems/ might be examples).

I don't understand why .sysr is in .gitignore.  Even if the intent is to ignore entirely the :useless-runes argument of certify-book (see comment above), wouldn't we want to see inappropriate .sysr directories with "git status"?

[I'm out of my depth on this next comment because I'm only somewhat familiar with the build system implementation; maybe Sol will have a comment on this.]  Should environment variable ACL2_USELESS_RUNES be set in both cert_features.lsp and cert.pl?  Might cert_features.lsp suffice?  There seems to be precedent for such a setting in cert_features.lsp but I don't see one in cert.pl.  I think maybe the answer is that cert_features.lsp is used only with "make" but the point of this pull request is to change cert.pl.  In that case, could we remove handling of useless-runes from cert_features.lsp?  Or would that change how custom makefiles handle useless-runes?  It might be good to address these questions in commennts in cert_features.lsp and cert.pl, in case they come up again.

bendyarm:
> It appears that the modified definition of useless-runes-value now causes the :useless-runes argument of certify-book to be ignored in ACL2(r). Is that what's intended? 

Yes, that was added as a measure of safety for you (@MattKaufmann), so that you would not be worried about supporting useless-runes on ACL2(r).

> It seems to me better to avoid routine use of useless-runes in ACL2(r), not to prohibit it entirely if that's what somebody wants. I can also imagine installing useless-runes for some specific ACL2(r) community books when that doesn't make them fail to certify and their certification is very slow otherwise (some books under books/projects/filesystems/ might be examples).

Ruben and I discussed this and we would agree with that.  Our plan was, if this PR is merged and works for people, that we would then do some more experiments with useless-runes in ACL2(r) by removing the restriction in `useless-runes-value`, and if that works under all scenarios we can think of, proposing another PR that carefully extends useless-runes to ACL2(r).  We would separate the issue of allowing (r) users to use useless-runes on their own books from the issue of precomputing useless-runes in the distributed `books` directory.

> I don't understand why .sysr is in .gitignore. Even if the intent is to ignore entirely the :useless-runes argument of certify-book (see comment above), wouldn't we want to see inappropriate .sysr directories with "git status"?

The idea is to prevent inadvertent pushes to the community books of `.sysr/` directories.  The purpose is so that you do not have to worry about maintaining useless-runes on ACL2(r).  It is an additional measure of protection.  If we ever support precomputed useless runes for (r) in the community books, then `.sysr/` would have to be removed from the file `books/.gitignore`.

> [I'm out of my depth on this next comment because I'm only somewhat familiar with the build system implementation; maybe Sol will have a comment on this.] Should environment variable ACL2_USELESS_RUNES be set in both cert_features.lsp and cert.pl? Might cert_features.lsp suffice? There seems to be precedent for such a setting in cert_features.lsp but I don't see one in cert.pl. I think maybe the answer is that cert_features.lsp is used only with "make" but the point of this pull request is to change cert.pl. In that case, could we remove handling of useless-runes from cert_features.lsp? Or would that change how custom makefiles handle useless-runes? It might be good to address these questions in commennts in cert_features.lsp and cert.pl, in case they come up again.

Experiments prove that the `export ACL2_USELESS_RUNES ?= -25` in `cert_features.lsp` does not suffice.  Since `cert.pl` is used for certifying a book and just the books it depends on, a common use case, that is the natural place for setting `ACL2_USELESS_RUNES` to `-25`.  

If we remove the export command from `cert_features.lsp` then `make` in the books directory will no longer use the useless-runes.  That would slow down builds a lot.

I do not know what happens with custom makefiles.  We could add an issue for testing what happens with useless-runes with a custom makefile, in ACL2, ACL2(p), and ACL2(r).  Probably no change with ACL2.   I could see a possible change in the behavior of custom makefiles with ACL2(p) and ACL2(r), since with this PR those variant builds no longer prevent the makefile feature `export ACL2_USELESS_RUNES ?= -25` from happening.  In the case of ACL2(p), useless runes files could be used when they were not expected (although you already did a regression of that for the main makefile), and in the case of ACL2(r) the other changes prevent useless runes files from being used, so no change there with this PR.

MattKaufmann (approved):
That all looks good, where I'm assuming from your answer that putting .sysr into .gitignore prevents adding files under .sysr directories to the repository.  (Otherwise we should discuss the .gitignore plan further.)

You can ignore my previous comment about custom makefiles (and your own last paragraph just above) as far as I'm concerned: your explanation seems fine to me for why you changed both cert_features.lsp and cert.pl, especially if @solswords does the requested review.

By the way, my concern about useless-runes files for ACL2(r) is mainly about who will maintain them.  I don't mind doing that (as I do for ACL2) if it's easy, but if I run into difficulties then I might ask you or Ruben to do so, or even make a release without things being in order for ACL2(r) (normally I test first).

I suggest one additional change, but I'd be happy to make it (should only take me a few minutes): to print a warning when a non-nil value is supplied for the :useless-runes option of certify-book.  If you disagree, let me know; otherwise I'll probably do that (unless you prefer to do it yourself).

bendyarm:
> That all looks good, where I'm assuming from your answer that putting .sysr into .gitignore prevents adding files under .sysr directories to the repository. (Otherwise we should discuss the .gitignore plan further.)

Yes, it tells git not to track them.  So `git add --all` and `git commit --all` don't accidentally get them.  However, that's just an added safety, since we don't expect there will be any files written to `.sysr/` directories.

> I suggest one additional change, but I'd be happy to make it (should only take me a few minutes): to print a warning when a non-nil value is supplied for the :useless-runes option of certify-book. If you disagree, let me know; otherwise I'll probably do that (unless you prefer to do it yourself).

I'm not sure all the cases you want to cover, so please go ahead and add the warnings that you want.  
Thanks!

MattKaufmann:
OK, I'll probably add the certify-book warning myself after the pull request makes it into master (perhaps in function useless-runes-info).

I plan to wait for @rubengamboa and @solswords to approve the pull request before I accept it (along the lines of what you (@bendyarm) suggested to me in an email).

By the way, last month I started a patch to ACL2 to permit LD to use useless-runes.  I'll probably wait to complete it till after this pull request is merged.  There's no urgency as far as I know, though I'd like to complete that within 3 or 4 weeks from now.

solswords (approved):
LGTM

rubengamboa (approved):
This stuff looks good for now. It won't break ACL2(r), at any rate. A future version should allow useless-runes for ACL2(r), and I'll try to get to that in the next couple of weeks.

PR #1342, opened 2022-01-13
doc corresponding to change making cert.pl use useless runes

bendyarm:
This is the documentation for the cert-heed-useless-runes-2 branch PR#1341.
This does not include the three derived files `doc.lisp doc/acl2-code-size.txt doc/home-page.html`
because I think it is possible that there will be merges that would require they be regenerated.
I can add them to this PR if needed.

### DIFF
--- a/books/doc/relnotes.lisp
+++ b/books/doc/relnotes.lisp
@@ -135,31 +135,6 @@
 
    ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-   (xdoc::h4 (xdoc::seetopic "build::cert.pl" "Build Library"))
-
-   (xdoc::p
-    "The <see topic=\"BUILD____CERT.PL\">cert.pl</see> shell command
-     by default now makes use of @(see useless-runes) files when they
-     are available.  It does that by binding the environment variable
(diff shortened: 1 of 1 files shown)


MattKaufmann (changes requested):
Thanks for updating the :doc.

File acl2-doc.lisp is about changes to the ACL2 system only (maintained by J and me, as opposed to the community books and its supporting infrastructure).  So please move the item you added on cert.pl to :doc note-8-5-books -- actually you already have an item there, so maybe just delete the one in acl2-doc.lisp or else merge it into the one in note-8-5-books.

Also, please move the other two items in acl2-doc.lisp into the "Experimental Versions" section near the end of :doc note-8-5, since that's where items go that have nothing to do with ACL2 proper, but only ACL2(r) and/or ACL2(p).  I'd be happy to change the name of that section, for example from "Experimental Versions" to "Variants of ACL2" (how about "variants of concern"? ;).

bendyarm:
@MattKaufmann I made those changes.  Also improved the <see ...> tags to use @(url ..).  Also removed the new "Build Library" section of note-8-5-books that I had added and moved the contents into the preexisting "Build System Updates" section.

MattKaufmann (approved):
Looks great; thanks, @bendyarm.

PR #1346, opened 2022-01-19
One more application of the U-substitution rule in ACL2(r).

jagadishb09:


MattKaufmann:
Directories under workshops/ are traditionally for accepted papers.  Since 2022 workshop papers haven't yet been reviewed, it's premature to create books/nonstd/workshops/2022/.  Please move u-substiitution-example.lisp elsewhere; it can always be moved again later.

PR #1353, opened 2022-02-03
[xdoc] Fix defpointer to tolerate symbols with unusual names.

ericwhitmansmith:
This pull request fixes a problem that occurs when defpointer is used to point to an xdoc topic whose name is an unusual symbol, such as a symbol that includes spaces. Without this patch, the following gives an error:

(include-book "xdoc/topics" :dir :system)
(defxdoc |foo bar| :short "example")
(defpointer my-pointer |foo bar|)
:doc my-pointer

The error message is:
; xdoc error in MY-POINTER:
"In SEE directive, expected ) after FOO. Near See @(see acl2::foo bar)..".
ACL2::MY-POINTER -- Current Interactive Session

Note the space between foo and bar in the generated @(see ...) form; this is not suitable for reading back in the symbol |foo bar|.

This fix is more conservative than one I proposed a long time ago in that it always includes the package of the symbol.

### DIFF
--- a/books/xdoc/top.lisp
+++ b/books/xdoc/top.lisp
@@ -504,17 +504,38 @@
       (concatenate-symbol-names (list ,@args))
       mksym-package-symbol)))
 
+
+;; Converts SYM to a string in such a way that it can be read back in as SYM.
+;; Always includes the package prefix and always puts vertical bars around the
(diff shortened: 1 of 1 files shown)


MattKaufmann:
I didn't really look at this, but I figure it will be quickly obvious if it introduces problems, so I'm accepting without looking.

PR #1354, opened 2022-02-05
Connect syntheto and wrap-output documentation fixing xdoc generation…

westfold:
… for some b* binders.

Was causing xdoc errors for binders in different packages or with lower case characters.

### DIFF
--- a/books/kestrel/apt/doc.lisp
+++ b/books/kestrel/apt/doc.lisp
@@ -52,6 +52,8 @@
 (include-book "tailrec")
 (include-book "tailrec-doc")
 
+(include-book "wrap-output")
+
 ; (depends-on "images/apt-logo.png")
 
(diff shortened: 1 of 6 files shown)


acoglio (approved):

ericwhitmansmith:
It would be better not to change bstar to include xdoc/top.  Maybe factor out symbol-as-readable-string.

westfold:
Yes, although I did check and found that xdoc/top doesn’t contain much beyond xdoc/base.

ericwhitmansmith:
It looks like it would bring in 2 new books (xdoc/top itself and xdoc/book-thms).  That seems unfortunate, since bstar is used a lot.

westfold:
Yes, although the books aren’t very big.

The simple fix for this is to move symbol-as-readable-string to xdoc/base, but ideally it should be in a printing utility book.

ericwhitmansmith:
it turn out there's a function, full-escape-symbol, in xdoc/top, that seems like it may be more robust than symbol-as-readable-string.  I'm going to try to factor it out into a separate book.

PR #1369, opened 2022-03-16
[X86ISA] Fix bounds in IDIV specification.

acoglio:
It appears that the lower and upper bounds of the signed quotient were slightly
off, as they should be the normal two's complement limits. This commit fixes
them.

This issue was first noticed by Eric Smith.


shigoel (approved):
Sorry about that, and thank you so very much!

PR #1375, opened 2022-04-19
Don't create xdoc for b*-binder unless parents, short, or long given

westfold:
Reduces size of B*-binders xdoc page from ~2000 entries to <40 entries.


solswords (approved):
OK with me

PR #1376, opened 2022-04-26
[xdoc] Avoid including centaur/fty/portcullis in xdoc/cert.acl2.

ericwhitmansmith:
Prior to this change, books/xdoc/cert.acl2 contained:
(include-book "centaur/fty/portcullis" :dir :system)
which seemed unnecessary.  I don't see any mentions of fty in xdoc/, and nothing broke when I removed this include-book.

The include-book led to unnecessary dependencies, for example b* depending on books/centaur/bitops/portcullis.lisp (since fty depends on centaur/bitops/portcullis).

### DIFF
--- a/books/kestrel/fty/deffixequiv-sk-doc.lisp
+++ b/books/kestrel/fty/deffixequiv-sk-doc.lisp
@@ -10,6 +10,7 @@
 
 (in-package "FTY")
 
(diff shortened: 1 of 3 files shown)


solswords (approved):
LGTM

acoglio (approved):

PR #1378, opened 2022-04-28
[VL] Factor string-fix and fast-memberp/make-lookup-alist into separa…

ericwhitmansmith:
…te files.

This is in support of reducing the dependencies of some things that include VL utilities (including defrstobj, the x86 model, and Axe) which I will work on in future (small) commits.

This change just pulls a few functions into separate files.  The utilities involved actually seem more general than VL, so perhaps one day they should be moved out of VL and put into a different package, but I didn't do any of that, to avoid breakage.


acoglio (approved):
Looks like a good idea to me.

ericwhitmansmith:
Sol told me on Slack that he's fine with this change, so I'm going to merge it.

PR #1387, opened 2022-05-03
[VL] Reduce deps of vl/util/namedb.

ericwhitmansmith:
This change prevents centaur/vl/util/namedb from including centaur/vl/util/defs.  Various include-books have to be added in namedb.lisp, and other files, to compensate, but the net effect is to greatly reduce what is included by namedb and tools that include it, including defrstobj, the x86 model, and Axe.

This change reduces the number of books included by defrstobj2 from 213 to 127.  Among the things no longer included are utilities for encoding strings in HTML!   The number of books included by kestrel/axe/top decreases from 1238 to 1169.  And projects/x86isa/top goes from 398 books to 329 books.

Further improvements could probably be made in vl by reducing what is included in centaur/vl/util/defs.lisp itself (e.g., not including std/strings/top, which is presumably what brings in the aforementioned HTML utilities).

### DIFF
--- a/books/centaur/aig/aig2c.lisp
+++ b/books/centaur/aig/aig2c.lisp
@@ -30,6 +30,7 @@
 
 (in-package "ACL2")
 (include-book "std/util/top" :dir :system)
+(include-book "centaur/vl/util/defs" :dir :system)
 (include-book "centaur/vl/util/namedb" :dir :system)
 (include-book "aig-base")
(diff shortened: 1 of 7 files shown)


solswords (approved):
Looks good. Thanks!

acoglio (approved):

shigoel (approved):
This is great! Thanks, Eric.

PR #1392, opened 2022-05-04
Two calls to debug_print_event got errors when $debugging was set.

bendyarm:
This change makes $debugging work.


solswords (approved):
looks good

PR #1393, opened 2022-05-04
add flag to cert.pl to write book dependency info in S-expression form

bendyarm:
Here is a review of the changes:

- `books/build/cert.pl`: Add "`--smakefile`" option.  If provided, it causes dependency information to be written out in S-expression format.

  From the COMMAND LINE OPTIONS help string:
```
   --smakefile
           Also write out S-expression version of the dependency information.
           Appends ".lsp" to the <makefile-name> provided by "-o",
           or Makefile-tmp.lsp by default.

```

- `books/GNUmakefile`: When doing a scan of the community books using `cert.pl`,
  provide the `--smakefile` flag.  This will cause a file
 ` books/build/Makefile-deps.lsp` to be written out.
  (To write out the deps files without actually doing a build, you can do, e.g.:
     `make --dry-run basic > /dev/null`)

  The `Makefile-deps.lsp` file is a little under 4 MB --- less than the `Makefile-deps`
  but more than the `Makefile-cache`.
  To inhibit this behavior, the environment variable `INHIBIT_DEPS_LSP` can
  be set to any nonempty string.  Note that this variable only affects how
  `GNUmakefile` calls `cert.pl` for scanning.  It does not affect what `cert.pl`
  does with a given argument list.

- `books/build/clean.pl`: Add `Makefile-tmp.lsp` and `Makefile-deps.lsp` to the list of files that get cleaned.

- `books/.gitignore`: Add `Makefile-tmp.lsp` and `Makefile-deps.lsp` to the list of paths that get ignored.


solswords (approved):
I'm fine with doing this, just curious as to why (and why by default)? Is this going to be used in some book that's certified as part of the regression?

bendyarm:
@solswords Thanks!  We are already using the lisp dependency information for doing queries and we have some ideas for tools that will use it.  

About the default: `cert.pl` does not default to writing the `.lsp` dependency information, but `GNUmakefile` does, when it does the dependency scan.  We think it is an artifact that is more `(read)`able by most ACL2 users than is the Makefile-cache and could be used for various things, and it is more likely to be noticed and used if people see it there.  If someone doesn't want it they can turn it off with the environment variable.  If people don't like that we could switch the default the other way.

 We have no plan to make the actual value be part of a certificate. That's an interesting idea but we would have to be careful about it since it is large and depends on all the community book sources.

ericwhitmansmith:
Looks good to me, and will be helpful to some stuff I'm doing.  I suggest having the name of the environment var INHIBIT_DEPS_LSP start with ACL2_.

ericwhitmansmith (approved):

bendyarm:
@ericwhitmansmith Good idea, thanks!  I made that change.

PR #1398, opened 2022-05-15
Arithmetic deps

ericwhitmansmith:
Include cowles/acl2-crg only locally (adjust a few workshop books to compensate).

Remove arithmetic/cert.acl2 since it is unnecessary.  This also prevents the unncessary inclusion of std/portcullis in some books.

Avoid exporting cowles/portcullis when possible by adding no_port to local include-books.

As an example of the reduced inclusions from this change, now including arithmetic/mod-gcd brings in just that book, whereas before we also got books/cowles/portcullis.lisp and books/std/portcullis.lisp.


acoglio (approved):

solswords (approved):
Ok with me

PR #1401, opened 2022-05-18
Allow libssl library from recent Ubuntu

westfold:


solswords:
Does running update-libs.sh in the quicklisp dir fix the problem? Could you install a supported libssl library version as a workaround?  If you commit changes to imported quicklisp sources, those changes will disappear as soon as someone decides it's time to update quicklisp, which isn't great.

ericwhitmansmith:
This ssl problem seems to bite a lot of people who try to install ACL2 and build the books.  Could we make the use of ssl (and thus, I guess, hunchentoot) optional and off by default?

ericwhitmansmith:
Is it the case that most users already have an appropriate SSL library installed, so what's needed is for ACL2 to just recognize it, or does building the books often require installing some additional SSL library?  If the latter, and if we can't suppress the need for SSL, it seems we should clearly explain in the ACL2 installation instructions how to ensure the right SSL libraries are installed, but that sounds hard in general, since it could depend on the OS.

MattKaufmann:
I don't think J or I want to maintain installation instructions pertaining to any of this.  But we could add a pointer to something (e.g., in xdoc) that provides such instructions.

solswords:
It would be nice to make all the quicklisp libraries independent, so that one could e.g. use osicat without having to worry about whether hunchentoot works.  The problem is that if we try to compile the libraries in parallel, they can step on each other's mutual dependencies, so we have to somehow serialize them.  The way we currently do this is by having quicklisp/base actually build everything, and everything else depends on that.  We could instead have some linear sequence of dependencies that compile different things, so we know that only one compile is running at a time in a parallel build. A pragmatic approach might be to take hunchentoot out of base and have it compile in its own book that depends on base.  Another approach might be to have our books reflect the actual dependencies of the libraries (if we can find and reliably update that information), and have each book correspond to exactly one library that needs compiling, which doesn't touch any other library's files.  Another approach might be to have separate quicklisp directories and imported bundles for some separate libraries (but that might have its own problems).

westfold:
No. Installing one of the older versions of the library would presumably work but I have no idea how difficult that is, especially without root access. I could try reporting a quicklisp bug.

Stephen

kini:
Looks fixed upstream: https://github.com/cl-plus-ssl/cl-plus-ssl/commit/c08c60e624050c114e55e7b8c4e3581cf54f2f4b

solswords:
So I tried updating Quicklisp but unfortunately the cl+ssl version that's current there (2022-03-31) doesn't have this change, so I'm not sure if it's worth pushing the update.

I'm basically fine with separating out Hunchentoot, which I think is the only library that needs cl+ssl. For now I'd go with the scheme I suggested before where the rest of the libraries are loaded in base.lisp and hunchentoot is loaded separately but depends on that (just need to be very clear in comments that we need to maintain a linear chain of dependencies to prevent parallel builds).  However, unfortunately, the VL server depends on Hunchentoot and is built into the VL Kit, so we'd need to figure out how to conditionalize that.

I guess if this sort of edit to Quicklisp source files solves your problems then I'm ok with it, but maybe we can add some instructions for reapplying the patches when we update quicklisp.

PR #1402, opened 2022-05-26
Updated defstruct XDOC

karthiknukala:
@acoglio 

Updated the XDOC topic for ATC `defstruct`s
1. Included general form + example of usage, generated C code
2. Clarified input type ("sequence of members" instead of "list of members")
3. Fixed small typos


books/kestrel/c/atc/defstruct.lisp
```diff
@@ -37,18 +37,31 @@
...
+           (field_name_n field_type_n))"))
+   (xdoc::evmac-section-inputs
+    (xdoc::desc
+     "@('struct_name')"
+     (xdoc::p
+      "Name [C:6.7.2.3] of the structure type"))
+    (xdoc::desc
+     "@('(field_name_1 field_type_1), ..., (field_name_n field_type_n)')"
+     (xdoc::p
+      "A sequence of members")
```
acoglio, line 54: members => members.

books/kestrel/c/atc/defstruct.lisp
```diff
@@ -847,8 +876,8 @@
            (define ,struct-tag-write-name ((val ,typep) (struct ,struct-tag-p))
              :returns (new-struct ,struct-tag-p)
              (value-struct-write (ident ,(ident->name name))
-                                 (,type-fix val)
-                                 (,struct-tag-fix struct))
+                                  (,type-fix val)
```
acoglio, line 879: Extra space shouldn't be there. Also in the line below.

acoglio (changes requested):
Thanks! A few suggested changes. Also, the PR should be made against testing or testing-kestrel, not master (even though it's the default).

PR #1403, opened 2022-06-08
Fixed certification issue with projects/rp-rewriter/lib/mult3/demo/demo-3.lisp

temelmertcan:
And some other changes


MattKaufmann:
Does anyone want to look at the oslib change?  I'll probably accept this on Tuesday if nobody replies before then.

MattKaufmann:
I said I'd probably merge today, but I see now that the target is testing-intel, not testing; so I'm not sure it's my place to merge.  Comments welcomed, especially from someone at Intel.

PR #1404, opened 2022-06-23
Lift fresh-namep-msg-weak and fresh-namep-listp-msg-weak into logic mode

gjurgensen:
These functions may be defined in logic mode now that their system function dependencies have been lifted to logic mode.


MattKaufmann:
Since this is a merge into testing-kestrel, I'll let someone at Kestrel merge it.  But here's a question.  Will any files outside books/ be changed when this migrates eventually to master?  I think maybe the answer is "no" since the ACL2 source files modified here might be the ones I pushed yesterday.  Is there a way that one can easily be sure?

MattKaufmann:
Another question: Was the addition of books/kestrel/std/system/.fresh-namep.lisp.swp intentional, and if so, what is its nature?

MattKaufmann:
Sorry about the noise from my first question: I realize now that I just needed to look at the authors of the three commits.  I'm still interested in an answer to my second question, about the .swp file.

gjurgensen:
The inclusion of the .swp file was definitely not intentional, I'll remove it. Thanks for pointing that out.

acoglio:
@MattKaufmann I think that the application of the changes to the files outside `books/` would be a no-op because of the reason you said (i.e. that you already made the same exact changes), but it may still affect the commit history of those files, which may interfere with your purity scripts.

@gjurgensen It may be better to create a new PR that only affects files under `books/`.

ericwhitmansmith:
Good point, Alessandro.  I agree that a clean PR affecting only books/ would be better.

gjurgensen:
Sure, we can close this PR and I can submit another which only affects  `books/`. The changes to `books/` depend on Matt's updates on `master` though, so I'll wait until that makes its way to testing-kestrel before submitting that second PR.

acoglio:
@gjurgensen I think it's fine to wait for Matt's commits to go into `testing-kestrel` (which, by the way, doesn't happen automatically but some of us normally "syncs" them). Or you can make a PR against the `testing` branch, which has them.

PR #1406, opened 2022-07-05
Fix fastnumio on sbcl

westfold:
Also allows it to work on M1 Macs
Also removes bug workaround that was fixed long ago


MattKaufmann:
Thanks, Stephen!  I hope someone accepts this soon (but I probably shouldn't, since I'm not quicklisp-savvy and since the merge is into testing-kestrel).

PR #1409, opened 2022-07-22
[Std/util] Add an :IFERR option to the ER binder.

acoglio:
In
```
 (b* (...
      ((er x) <something>)
      ...)
   ...)
```
if `<something>` returns `(mv erp val state)` with non-`nil` `erp`, the `b*` also returns
`(mv erp val state)`.

Instead, in
```
 (b* (...
      ((er x :iferr eval) <something>)
      ...)
   ...)
```
if `<something>` returns `(mv erp val state)` with non-`nil` `erp`, the `b*` returns `(mv
erp eval state)`, i.e. the chosen `eval` instead of `val`.

This is useful, for example, when writing code that traffics in error triples
whose value components must have different types, when we want those types to be
unconditional. For instance, the `b*` could be in a function `g` whose returned `val`
is unconditionally a string, while `<something>` is a call of a function `f` whose
returned `val` is unconditionally an integer: using just `(er x)` would lead to a
type violation (manifesting as a return theorem not being proved), while using
`(er x :iferr "")` would let `g` unconditionally return a string.

This is more compact then binding the whole triple and checking `erp`, i.e.
```
 (b* (...
      ((mv erp val state) <something>)
      ((when erp) (mv erp eval state))
      ...)
   ...)
```
The guard of `patbind-er`, which was previously just `destructure-guard`, has been
relaxed to allow the option, while inlining the relevant parts of
`destructure-guard`.

### DIFF
--- a/books/std/util/bstar.lisp
+++ b/books/std/util/bstar.lisp
@@ -27,6 +27,7 @@
 ;   DEALINGS IN THE SOFTWARE.
 ;
 ; Original author: Sol Swords <sswords@centtech.com>
+; Contributing author: Alessandro Coglio <coglio@kestrel.edu>
 
 (in-package "ACL2")
 (include-book "xdoc/base" :dir :system)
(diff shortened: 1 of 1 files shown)


solswords (approved):
Lgtm

PR #1416, opened 2022-08-04
Guard-verify some numbered-names utilities

gjurgensen:
The `get-numbered-name-index-*` functions required minor modifications to prove the functions had the suggested return types.

### DIFF
--- a/books/kestrel/utilities/system/numbered-names.lisp
+++ b/books/kestrel/utilities/system/numbered-names.lisp
@@ -13,9 +13,12 @@
 (include-book "kestrel/std/strings/nondigit-chars" :dir :system)
 (include-book "kestrel/std/system/fresh-namep" :dir :system)
 (include-book "kestrel/std/system/pseudo-event-formp" :dir :system)
+(include-book "kestrel/std/system/table-alist-plus" :dir :system)
 (include-book "std/strings/decimal" :dir :system)
 (include-book "std/util/defval" :dir :system)
+(include-book "std/util/defrule" :dir :system)
 (include-book "system/kestrel" :dir :system)
+(local (include-book "kestrel/alists-light/assoc-equal" :dir :system))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
@@ -123,8 +126,7 @@
   "{")
 
 (define get-numbered-name-index-start ((wrld plist-worldp))
-  :returns (start "A @(tsee numbered-name-index-start-p).")
-  :verify-guards nil
+  :returns (start numbered-name-index-start-p "A @(tsee numbered-name-index-start-p).")
   :parents (numbered-name-index-start)
   :short "Retrieve the current starting marker
           of the numeric index of numbered names."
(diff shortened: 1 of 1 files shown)


books/kestrel/utilities/system/numbered-names.lisp
```diff
@@ -123,17 +126,25 @@
...
           of the numeric index of numbered names."
   :long
   "<p>
    If the starting marker is not set yet, the default is returned.
    </p>"
-  (let ((pair (assoc-eq 'start (table-alist 'numbered-name-index-start wrld))))
-    (if pair (cdr pair) *default-numbered-name-index-start*)))
+  (let ((pair (assoc-eq 'start (table-alist+ 'numbered-name-index-start wrld))))
+    (if (and pair
+             (numbered-name-index-start-p (cdr pair)))
```
acoglio, line 139: I think it’s fine as is, but in general I would throw an internal error if this condition fails, since (I think) we expect it to never fail. Something like
```
(raise “Internal error: ~x0 does not satisfy NUMBERED-NAME-INDEX-START-P.” (cdr pair))
```

books/kestrel/utilities/system/numbered-names.lisp
```diff
@@ -123,17 +126,25 @@
...
+        (cdr pair)
+      *default-numbered-name-index-start*)))
+
+(defrule stringp-of-get-numbered-name-index-start
+  (implies (plist-worldp wrld)
```
acoglio, line 144: Do you need this hyp?

books/kestrel/utilities/system/numbered-names.lisp
```diff
@@ -275,14 +302,20 @@
   (defmacro set-numbered-name-index-wildcard (wildcard)
     `(table numbered-name-index-wildcard 'wildcard ,wildcard)))
 
+;; TODO: move?
```
acoglio, line 305: You could try doing
```
(local (include-book “std/typed-lists/character-listp” :dir :system)
```
instead of proving this theorem.

books/kestrel/utilities/system/numbered-names.lisp
```diff
@@ -123,17 +126,25 @@
   "{")
 
 (define get-numbered-name-index-start ((wrld plist-worldp))
-  :returns (start "A @(tsee numbered-name-index-start-p).")
-  :verify-guards nil
+  :returns (start numbered-name-index-start-p "A @(tsee numbered-name-index-start-p).")
```
acoglio, line 129: You can remove `”A @(tsee numbered-name-index-start-p).”` now that it’s actually code.

books/kestrel/utilities/system/numbered-names.lisp
```diff
@@ -123,17 +126,25 @@
...
    </p>"
-  (let ((pair (assoc-eq 'start (table-alist 'numbered-name-index-start wrld))))
-    (if pair (cdr pair) *default-numbered-name-index-start*)))
+  (let ((pair (assoc-eq 'start (table-alist+ 'numbered-name-index-start wrld))))
+    (if (and pair
+             (numbered-name-index-start-p (cdr pair)))
+        (cdr pair)
+      *default-numbered-name-index-start*)))
+
+(defrule stringp-of-get-numbered-name-index-start
```
acoglio, line 143: Using `defrule` is just fine. For even more compactness, you may be interested in `defret`.

books/kestrel/utilities/system/numbered-names.lisp
```diff
@@ -626,17 +662,16 @@
...
   :short "Retrieve the global index for numbered names."
   :long
   "<p>
    If the global index is not set yet, 1 is returned.
    </p>"
   (let ((pair (assoc-eq 'index
-                        (table-alist 'global-numbered-name-index wrld))))
-    (if pair (cdr pair) 1)))
+                        (table-alist+ 'global-numbered-name-index wrld))))
+    (pos-fix (cdr pair))))
```
acoglio, line 674: You may also consider throwing an internal error if the expected `posp` condition is violated.

acoglio (approved):
Thanks! This is good, but I left a few comments for consideration. Some of them apply to other similar instances in the file.

books/kestrel/utilities/system/numbered-names.lisp
gjurgensen: Isn't `(raise ...)` logically equivalent to `nil`? Presumably this would invalidate the returns lemma.

acoglio: Yes, great point. You can "adjust" the return type via
```
(prog2$ (raise ...) <anything of the right type>)
```
where `<anything of the right type>` will be never executed, but serves the logical purpose.

gjurgensen: Oh, of course, that makes perfect sense.

gjurgensen:
Thanks for the feedback, I think these changes address all of your notes.

PR #1417, opened 2022-08-10
[doc] Improve :DOC VERIFY-GUARDS-FOR-SYSTEM-FUNCTIONS.

acoglio:
Add a paragraph to clarify the use of VERIFY-TERMINATION and VERIFY-GUARDS.

Add a sentence motivating one of the steps.


books/system/doc/acl2-doc.lisp
```diff
@@ -132567,16 +132567,24 @@ introduction-to-the-tau-system) for more information about Tau.</dd>
...
+ typically to create a new file, perhaps named after the main function(s) whose
+ guards you want to verify, in directory @('books/system').  Community book
+ @('books/system/too-many-ifs.lisp') is a good example.  Ultimately, this file
+ should be a certifiable book that verifies the desired termination and guards.
+ Include this book in @('books/system/top.lisp').  Of course, you can instead
+ add to some existing file in the same directory that is already included in
  @('books/system/top.lisp'), rather than creating a new book and including it
  there.</p>
 
+ <p>When verifying termination and guards of system functions in interactive
```
ericwhitmansmith, line 132580: Doesn't verify-termination attempt to verify guards iff there is an explicit guard (or type declaration, etc.)?

MattKaufmann: Right; I don't think there's any "special way" that these files (books) are treated.  A link could be provided to :DOC set-verify-guards-eagerness, where it is explained that the default behavior is "yes if :guard, type or :stobjs is supplied".

acoglio:
Somehow this is new to me: I thought that `verify-termination` always only verifies termination. But I see from the documentation that it may also verifies the guards, as you guys say. So suppose that there is a program-mode system function `f` that does not have a guard declaration (so it has guard `t`), and that `(verify-guards f)` succeeds (i.e. `f` does not need a stronger guard). Does the file under `[books]/system/` need to have both a `(verify-termination f)` and a `(verify-guards f)` then?

ericwhitmansmith:
Yes, I think so.  Look at books/system/legal-variablep.lisp.

MattKaufmann:
Yes; below is an example.  Notice the "L" near the margin from :pe, meaning "logic mode but not guard-verified".  Although the example below is interactive, there's no difference between that situation and a book.

    ACL2 !>(defun f (x) (declare (xargs :mode :program)) x)

    Summary
    Form:  ( DEFUN F ...)
    Rules: NIL
    Time:  0.00 seconds (prove: 0.00, print: 0.00, other: 0.00)
     F
    ACL2 !>(verify-termination f)

    Since F is non-recursive, its admission is trivial.  We observe that
    the type of F is described by the theorem (EQUAL (F X) X).  

    Summary
    Form:  ( DEFUN F ...)
    Rules: NIL
    Time:  0.00 seconds (prove: 0.00, print: 0.00, other: 0.00)

    Summary
    Form:  ( MAKE-EVENT (VERIFY-TERMINATION-FN ...))
    Rules: NIL
    Time:  0.00 seconds (prove: 0.00, print: 0.00, other: 0.00)
     F
    ACL2 !>:pe f
     L         2:x(VERIFY-TERMINATION F)
		  \
    >L             (DEFUN F (X)
			  (DECLARE (XARGS :MODE :LOGIC))
			  X)

    Additional events for the logical name F:
     PL        1  (DEFUN F (X)
			 (DECLARE (XARGS :MODE :PROGRAM))
			 X)
    ACL2 !>

acoglio:
I've revised the text based on these discussions.

MattKaufmann (approved):
Looks great to me!  Thanks for the updates.

PR #1418, opened 2022-08-11
Turn doc strings into XDOC.

### LINKED ISSUE #319: Convert misc/hons-help.lisp documentation to xdoc
There are a bunch of common lisp documentation strings inside books/misc/hons-help.lisp.  If we think these functions are useful, we should convert them to xdoc-compatible primitives like defxdoc, define, etc.

This book is already included by doc/top.lisp, so all that needs to occur is cleanup of the book.

acoglio:
Closes #319.

That Issue suggested to turn the (now obsolete) doc strings into XDOC. The file already had some XDOC. This commit completes the job, so to speak.

I'm leaving this PR pending just in case anybody objects. It can be merged later.

### DIFF
--- a/books/misc/hons-help.lisp
+++ b/books/misc/hons-help.lisp
@@ -58,20 +58,21 @@ an existing fast alist like @('make-fal').</p>"
                        (cdar al)
                        (make-fal (cdr al) name))))))
 
-(defmacro ansfl (x y)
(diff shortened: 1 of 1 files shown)


MattKaufmann:
I see this hasn't yet been merged; for what it's worth, I have no objection.

PR #1421, opened 2022-08-24
[rp-rewriter] Remove include-books from portcullis book.

ericwhitmansmith:
Similar to a recent change.  The goal is to let me get the package from the portcullis without getting any books not necessary for that.


temelmertcan:
Thanks. It looks good to me. I don't have write access so I cannot accept the pull request.

ericwhitmansmith:
Thanks, Mert!  I merged this and will now delete the branch.

PR #1423, opened 2022-09-01
[sb-machine] Reduce dependencies of package.lsp.

ericwhitmansmith:
This allows the package to be obtained without also getting those books.  Add some includes to another book to compensate.

### DIFF
--- a/books/projects/sb-machine/package.lsp
+++ b/books/projects/sb-machine/package.lsp
@@ -3,11 +3,12 @@
 ; License: A 3-clause BSD license.  See the LICENSE file distributed with ACL2.
 
 (include-book "std/portcullis" :dir :system)
-(include-book "std/util/define" :dir :system)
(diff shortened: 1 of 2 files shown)


MattKaufmann:
This broke the testing branch.  Apparently std/basic/pos-fix was formerly included in an sb-machine book where it isn't now.

ericwhitmansmith:
Should be fixed now.  Sorry about that.  I'm pretty sure I tested this PR when I made it, but that was 2.5 weeks ago, and I guess things have evolved since then.

Anyway, all seems well.  I'm going to delete the branch now.

MattKaufmann:
Thanks, Eric.  Sorry it took so long to accept the pull request.  As far as I'm concerned, anyone with authorization to accept pull requests should be able to accept their own after waiting a couple of days or so.  Maybe someone has a different opinion....

PR #1424, opened 2022-09-02
[cert.pl] Attempt to fix cert.pl.

ericwhitmansmith:
Recent changes broke cert.pl for many of us.  The issue seems to be that it isn't finding the system books dir using the ACL2 environment variable, due to the function determine_acl2_dirs returning before it can reach the fourth way of setting acl2_books.  So I just conditionalized that return (in the code for the third way).

Big disclaimer: I don't know perl, or anything about cert.pl, so somebody should review this.

### DIFF
--- a/books/build/lib/Certlib.pm
+++ b/books/build/lib/Certlib.pm
@@ -234,11 +234,13 @@ sub determine_acl2_dirs {
 	# Check whether the SYSTEM dir was set when reading the acl2_projects file and set it if so.
 	$acl2_books = lookup_colon_dir("SYSTEM", {});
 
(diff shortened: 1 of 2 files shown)


solswords (approved):
Thanks, these both look correct. Apologies for the new bugs!

solswords:
Actually, if you get to it before merging, it might be better to use File::Spec->file_name_is_absolute($path) in the second patch rather than checking for an initial slash.

solswords:
If it's OK with people at Kestrel I will make that change and merge this -- seems important to unbreak this ASAP.

PR #1427, opened 2022-09-15
Add :undefined option to apt::isodata 

gjurgensen:
This option is processed in largely the same way as in `apt::restrict`. It affects what term is placed in the else branch of the top-level `mbt$` conditional of the new function.

The default value of the `:undefined` option is `:auto`, in which case it will default to the current behavior of using `nil` or `(mv nil ... nil)` for single-value and multi-value functions respectively. Note that one can use the quoted value `':auto` if one actually wishes the term `:auto` to be used as the "undefined" value.

### DIFF
--- a/books/kestrel/apt/isodata-doc.lisp
+++ b/books/kestrel/apt/isodata-doc.lisp
@@ -5,6 +5,7 @@
 ; License: A 3-clause BSD license. See the LICENSE file distributed with ACL2.
 ;
 ; Author: Alessandro Coglio (coglio@kestrel.edu)
+; Contributing Author: Grant Jurgensen (grant@kestrel.edu)
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
@@ -398,7 +399,21 @@
        "Any other term. It must be a term that only references logic-mode
         functions and that includes no free variables other than
         @('x1'), ..., @('xn'). This term must have no output
-        @(see acl2::stobj)s. This term must not reference @('old').")))
+        @(see acl2::stobj)s. This term must return the same number of results
+        as @('old'). This term must not reference @('old')."))
+     (xdoc::p
+      "If one wishes to use the term @(':auto') as the undefined result, this
+       may be accomplished by providing the quoted constant @('\':auto').")
+     (xdoc::p
+      "Even if the generated function is guard-verified
+       (which is determined by the @(':verify-guards') input; see below),
+       the undefined term need not be guard-verified.
+       Since the term is governed by the negation of the guard
+       (see the generated new function, below),
+       the verification of its guards always succeeds trivially.")
+     (xdoc::p
+      "In the rest of this documentation page, let @('undefined') be this term
+       specified by @(':undefined')."))
 
     (xdoc::desc-apt-input-new-name)
 
@@ -609,7 +624,7 @@
       "                 ..."
       "                 (newpn xn)))"
       "      old-body<(back1 x1),...,(backn xn)>"
-      "    nil)) ; or (mv nil ... nil)"
+      "    undefined))"
       ""
       ";; when old is not recursive,"
       ";; :predicate is nil,"
@@ -620,7 +635,7 @@
       "                 ..."
       "                 (newpn xn)))"
       "      (forth_r1 old-body<(back1 x1),...,(backn xn)>)"
-      "    nil))"
+      "    undefined))"
       ";; where forth_r1 is actually pushed into"
       ";; the 'if' branches of old-body if it starts with 'if',"
       ";; and recursively into their 'if' branches (if any)"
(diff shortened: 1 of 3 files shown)


acoglio (commented):
You can edit the file headers to say
```
; Main Author: Alessandro Coglio (coglio@kestrel.edu)
; Contributing Author: Grant Jurgensen (grant@kestrel.edu)
```
as we usually do in cases like this.

books/kestrel/apt/isodata-doc.lisp
```diff
@@ -382,6 +383,23 @@
...
+      "Denotes the value that the generated new function must return
+       outside of the new domain.")
+     (xdoc::p
+      "It must be one of the following:")
+     (xdoc::ul
+      (xdoc::li
+       "@(':auto'), to use @('nil') or @('(mv nil ... nil)') for single-value
+        and multi-value functions respectively.")
+      (xdoc::li
+       "Any other term. It must be a term that only references logic-mode
```
acoglio, line 398: Another requirement that we should state here is that the term returns the same number of results as the function.

books/kestrel/apt/isodata-doc.lisp
```diff
@@ -382,6 +383,23 @@
...
+        functions and that includes no free variables other than
+        @('x1'), ..., @('xn'). This term must have no output
+        @(see acl2::stobj)s. This term must not reference @('old').")))
+
```
acoglio, line 402: Just after the two `li` bullets, as last part of this `xdoc::desc`, we could add a paragraph `(xdoc::p ...)` saying that one can use a quoted `':auto` if they want to return the keyword `:auto` as undefined result.

acoglio (commented):
The 'Generated Events' section of the user documentation needs to be generalized in the parts where it shows `nil` or `(mv nil ... nil)` being returned by the generated function. You could say, at the end of the description of the `:undefined` input, something like "In the rest of this documentation page, let `undefined` be the term, determined by this `:undefined` input, that denotes the value returned outside the new domain by the generated new function." and then you can use `undefined` in the 'Generated Events' section.

books/kestrel/apt/isodata.lisp
```diff
@@ -1101,9 +1102,45 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(define isodata-process-undefined (undefined
+                                   (old$ symbolp)
+                                   ctx
+                                   state)
+  :returns (mv erp
+               (undefined "A @(tsee pseudo-termp) that is
```
acoglio, line 1110: I would use `undefined$` here, just for consistency with other code, and given that it differs from `undefined` in general.

books/kestrel/apt/isodata.lisp
```diff
@@ -1101,9 +1102,45 @@
...
+                           the translation of @('undefined').")
```
acoglio, line 1111: Maybe just say `"A @(tsee pseudo-termp)."`, because if `undefined` is `:auto`, then the result is not the translation of that.

books/kestrel/apt/isodata.lisp
```diff
@@ -2334,8 +2376,7 @@
      the resulting term is the code of the new function's body (see below).
      Then we construct an @(tsee if) as follows.
      The test is the conjunction of @('(newp1 x1)'), ..., @('(newpn xn)').
-     The `else' branch is @('nil') or @('(mv nil ... nil')),
-     depending on whether @('old') returns single or multiple results.
+     The `else' branch is @('undefined').
```
acoglio, line 2379: `@('undefined$')`

books/kestrel/apt/isodata-tests.lisp
```diff
@@ -491,6 +491,110 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(must-succeed*
```
acoglio, line 494: These are great tests, but I would move them down within the file. Now they are between the tests for `oldp` and `newp`, which should probably stay one after the other, since they are related. Perhaps these could go just after "Wildcard in OLD." and "Check NEW-NAME input."?

acoglio (commented):
Can you add `undefined` and `undefined$` to the two lists of names under `xdoc::evmac-topic-implementation`, at the beginning of the files just after the `include-book`s? (Ask me more if unclear.)

acoglio:
In the user doc that described the `:undefined` input, you could add something like this, similar to what the description of `:undefined` in `restrict` says:
```
     (xdoc::p
      "Even if the generated function is guard-verified
       (which is determined by the @(':verify-guards') input; see below),
       the term needs not be guard-verified.
       Since the term is governed by the negation of the guard
       (see the generated new function, below),
       the verification of its guards always succeeds trivially.")
```

gjurgensen:
Thanks Alessandro! I believe these new commits incorporate all of your feedback.

acoglio (approved):
This PR looks great, thanks!

PR #1433, opened 2022-09-29
Add base case option to `isodata` `:undefined` parameter

gjurgensen:
This adds `:base-case` and `:base-case-then` as options to `isodata`'s `:undefined` parameter. When either of these options are provided, `isodata` will choose the undefined value by looking for a base case within the newly generated function. A base case of a term may either be the whole term when the term does not include a recursive call, or if the term is an `if`, then it may be a base case of the `then` or `else` branch. A term may have several base cases; the `:base-case` option is biased toward selecting base cases in `else` branches, while `:base-case-then` is biased toward `then` branches.

The motivation for this feature is that a base-case will likely have a desirable type, and the result may be amenable to combining nested `if`s.

### DIFF
--- a/books/kestrel/apt/isodata-doc.lisp
+++ b/books/kestrel/apt/isodata-doc.lisp
@@ -398,6 +398,15 @@
        "@(':auto'), to use @('nil') or @('(mv nil ... nil)') for single-value
         and multi-value functions respectively.")
       (xdoc::li
+       "@(':base-case'), to search for a base-case within the domain of the new
+        function. A base-case of a term may be be the whole term when the term
+        does not include any recursive calls, or it may be a base-case of the
+        `then' or `else' branch when the translated term is an `if'. This
+        search for a base-case is biased toward `else' branches.")
+      (xdoc::li
+       "@(':base-case-then'), to search for a base-case with a bias toward
+        `then' branches.")
+      (xdoc::li
        "Any other term. It must be a term that only references logic-mode
         functions and that includes no free variables other than
         @('x1'), ..., @('xn'). This term must have no output
@@ -405,7 +414,8 @@
         as @('old'). This term must not reference @('old')."))
      (xdoc::p
       "If one wishes to use the term @(':auto') as the undefined result, this
-       may be accomplished by providing the quoted constant @('\':auto').")
+       may be accomplished by providing the quoted constant @('\':auto'). The
+       same applies for @(':base-case') and @(':base-case-then').")
      (xdoc::p
       "Even if the generated function is guard-verified
        (which is determined by the @(':verify-guards') input; see below),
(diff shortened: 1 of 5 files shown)


books/kestrel/apt/utilities/find-a-base-case.lisp
```diff
@@ -0,0 +1,154 @@
...
+
+(include-book "std/util/bstar" :dir :system)
+(include-book "kestrel/utilities/forms" :dir :system)
+(include-book "kestrel/terms-light/expr-calls-fn" :dir :system)
+(include-book "kestrel/utilities/translate" :dir :system)
+(include-book "kestrel/utilities/magic-macroexpand" :dir :system)
+(local (include-book "kestrel/typed-lists-light/pseudo-term-listp" :dir :system))
+(local (include-book "kestrel/typed-lists-light/symbol-listp" :dir :system))
+
+(defthm pseudo-termp-of-if-condition
```
acoglio, line 13: Can these three theorems be made local, so they do not affect the books that they include this tool directly or indirectly?

books/kestrel/apt/isodata-doc.lisp
```diff
@@ -397,6 +397,15 @@
       (xdoc::li
        "@(':auto'), to use @('nil') or @('(mv nil ... nil)') for single-value
         and multi-value functions respectively.")
+      (xdoc::li
+       "@(':base-case'), to search for a base-case within the domain of the new
```
acoglio, line 401: I would make the two options more symmetric, i.e. `:base-case-then` and `:base-case-else`, also switching them in the two `xdoc::li`s (i.e. first describing `:base-case-then`, then `:base-case-else` briefly).

books/kestrel/apt/utilities/find-a-base-case.lisp
```diff
@@ -0,0 +1,154 @@
...
+;; is `nil`, we are biased toward "else" branches.
+
+;; Returns `(mv erp all-base largest)`, where `erp` is `nil` when a base-case
+;; is found, `all-base` is `t` when the entire term is a base-case, and
+;; if `erp` and `all-base` are both `nil`, then `largest` is the largest base-case.
+(defun find-a-base-case-translated-aux (term fns prefer-then)
+  (declare (xargs :guard (and (pseudo-termp term)
+                              (symbol-listp fns)
+                              (booleanp prefer-then))))
+  (b* (((unless (consp term)) (mv nil t nil))
```
acoglio, line 44: Why is the 3rd result `nil` and not `term`?

books/kestrel/apt/utilities/find-a-base-case.lisp
```diff
@@ -0,0 +1,154 @@
...
+       ((unless (eq 'if (ffn-symb term)))
```
acoglio, line 45: It may be cleaner to have an explicit case/check for a quoted constant, between the previous check (for variables) and this check (where the use of `ffn-symb` suggests that we are in the case of a call term.

gjurgensen:
Thanks Alessandro, I made some updates. I had meant for the pseudo-termp theorems to be local originally; it was an oversight that they were not.

I also tried to improve the documentation of the find-a-base-case functions. Hopefully it is clearer now.

acoglio (approved):
Looks good, thanks!

PR #1439, opened 2022-10-30
[jenkins] Simplify Makefile.

ericwhitmansmith:
Complete the removal of old code for classic (non-hons) builds.  Non-hons builds are not really supported any more, and that code has been commented out for 5 years.

Avoid setting ACL2_HONS when calling make, since that has no effect.  Also tweak the messages printed to not mention 'h' in the ACL2 variant names.

Add more example usages in comments at top of file.


ericwhitmansmith:
I'm going to merge this, as David Rager approved it via email.

PR #1440, opened 2022-10-30
[rtl] Put macros like local-defun in ACL2 package (but import into RT…

ericwhitmansmith:
…L package).

This will allow references to remain working as books are moved to the ACL2 package.

Also, avoid having 8 copies of these macros by centralizing them to a new book.


ericwhitmansmith:
Approved by Russinoff via email.

PR #1441, opened 2022-10-31
Move several projects from RTL to ACL2 package.

ericwhitmansmith:
Move the following projects to the ACL2 package:

books/projects/quadratic-reciprocity
books/projects/curve25519
books/projects/shnf
books/workshops/2022/russinoff-groups

The rationale is that they are not about RTL and so should not be in that package.

This change means that some function and theorem names, including primep and divides, are now in the ACL2 package.

Also renamed the following (to avoid name clashes with pre-existing names in the ACL2 package): perm -> permutationp
sublistp -> sub-listp
sum$ -> sums$


acoglio (approved):

ragerdl:
My personal view is that moving things into acl2 is borderline backwards
progress, but I know there are philosophical differences here.  What about
moving them into another package?

MattKaufmann:
I had the same initial reaction as @ragerdl  .  I'm guessing though that the point is to standardize some things, kind of reserving them in the "ACL2" package.  @ericwhitmansmith , is that right?

ericwhitmansmith:
I'd be happy to put things in some other package(s), and I agree that that might be the ideal solution.  But then we get into issues of what the new package(s) should be called, who would "own" them, how the relevant libraries (which aren't mine) should be structured, and even how many new packages are needed (the main new one could be some abbreviation of "number theory" but then what about the group theory stuff and the Hebrew calendar stuff?).  Since Russinoff's email request about this asked me to the put those books into the ACL2 package, I decided to just do that, as a first step.  To me, it's a strict improvement; having primep live in the RTL package is crazy, whereas having it in the ACL2 package may not be ideal but at least has a precedent with other arithmetic libraries.

ericwhitmansmith:
BTW, by "those books" I really mean quadratic-reciprocity/, which was Doc's original request.  The other directories then came along for the ride; it seemed better to move them too than to make the fixes required to keep them in RTL after moving quadratic-reciprocity/ (which defines primep) to the ACL2 package.

russinoff:
As Eric says, "having primep live in the RTL package is crazy".  Of course, I am to blame for this.  Whenever I contribute a book that uses the rtl library in any way I put it in that package.  This applies to the following directories, which use that library only tangentially:

projects/quadratic-reciprocity
projects/curve25519
projects/shnf
workshops/2022/russinoff-groups
workshops/2022/russinoff-calendar

I should not have put any of these in the RTL package.  The most interesting are quadratic-reciprocity and russinoff-groups.  They are both likely to grow (I already have several books that I would like to add to the latter) and their interdependence is likely to increase.  They should be in the same package, something other than RTL, and perhaps something other than ACL2. (Eric has already identified some potential name conflicts between functions in those directories and existing ACL2 functions.)  I don't know what I'd call such a package; MATH seems presumptuous, but DMR has a nice ring.

None of the other three is likely to change or grow, and could probably be moved to ACL2.  If there were to be a new package created for q-r and groups, then I guess curve25519 and shnf could go there, but in any case, calendar should be in ACL2.

There are two other issues. First, quadratic-reciprocity is a bad name (the content is broader than that) and will be increasingly bad as the directory grows.  Can we change it to something like number-theory (or nt)?  Second, although russinoff-groups is appropriately a workshop directory, it doesn't make sense for such a directory to grow.  Should I replicate it in a separate directory where I can add new books?

MattKaufmann:
Regarding the final question, about replicating a workshop directory: This is a good, standard thing to do.  Then the workshop directory is static, which is good since then it will continue to correspond to the workshop paper.  A README in the workshop directory can direct people to the replicated, evolving directory.  In an extreme case the workshop directory might even stop being certified; see for example the top part of books/workshops/2018/kaufmann/README.

MattKaufmann:
@acoglio Your arguments are persuasive.  I'll just mention a weakness of the version control argument: a few years ago it may have seemed like we'd be using svn indefinitely.  But I don't feel strongly about this -- just wanted to mention the potential drawback.

russinoff:
Regarding the question of replicating the workshop books:

I think I'm in favor of leaving workshops/2022/russinoff-gorups/ as it is (after the package change), even though it will not quite match the corresponding books in the new evolving directory. One reason is that the latter will not correspond to the paper's treatment of the symmetric and alternating groups, which is greatly improved in the new version.

Regarding the more important question of packaging:

I don't like dealing with packages (I suppose because I don't understand them), and I resisted the creation of the RTL package years ago, but I became accustomed to dealing with it.  I don't like name conflicts either, so I'm warming to the idea of creating a new package.  I've struggled to find a name that works for both quadratic-reciprocity (the name of which I still want to change to something like nt) and groups, but I think I like DM (replacing my not-so-serious suggestion of DMR) for Discrete Mathematics.  This works for both of those directories as well as curve25519, shnf, and other areas that I hope to get to (linear algebra, algebraic fields).

ericwhitmansmith:
After email discussion with Russinoff, I've switched various dirs to use the DM package.  I'll probably push these changes in a few days if nobody objects (that push needs to be coordinated with lots of changes to Kestrel's proprietary books, which I have ready to also push).

I've tried to make sure acl2r builds as well, but, if I didn't do it perfectly, I have a Jenkins project that will send me an email.

PR #1444, opened 2022-11-08
[Std/util] Add a theorem and disable a function.

acoglio:
The theorem is disabled by default because `true-listp` is a common function.

### DIFF
--- a/books/std/util/support.lisp
+++ b/books/std/util/support.lisp
@@ -392,13 +392,16 @@ saying whether the key was bound.  Returns (mv value boundp)."
           (split-/// ctx (cdr x))))
       (mv (cons (car x) pre) post)))
 
-  (defthmd true-listp-of-split-///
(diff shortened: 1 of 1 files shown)


books/std/util/support.lisp
```diff
@@ -390,7 +390,15 @@ saying whether the key was bound.  Returns (mv value boundp)."
           (mv nil (cdr x)))
          ((mv pre post)
           (split-/// ctx (cdr x))))
-      (mv (cons (car x) pre) post))))
+      (mv (cons (car x) pre) post)))
+
+  (defthmd true-listp-of-split-///
+    (b* (((mv pre-/// post-///) (std::split-/// ctx x)))
+      (implies (true-listp x)
```
solswords, line 397: I think the hyp could apply only to the conclusion about post -- pre should be a true-list regardless

solswords (approved):
fine with me, but maybe you can strengthen the theorem as I suggested in the one comment.

PR #1448, opened 2022-12-29
fix bugs where backslashes in names were not escaped

bendyarm:
Also moved bar-escaping code to a more general location and package so that it can be more easily used for other thiings than xdoc.

### DIFF
--- a/books/misc/symbol-print-full-escapes.lisp
+++ b/books/misc/symbol-print-full-escapes.lisp
@@ -1,3 +1,4 @@
+; Originally part of...
 ; XDOC Documentation System for ACL2
 ; Copyright (C) 2009-2015 Centaur Technology
 ;
@@ -28,14 +29,11 @@
 ;
 ; Original author: Jared Davis <jared@centtech.com>
 
-(in-package "XDOC")
+; In 2022, moved from books/xdoc/full-escape-symbol.lisp
+; and from XDOC to ACL2 package, and fixed escaping of backslash.
+; Eric McCarthy (mccarthy@kestrel.edu).
 
-(include-book "misc/symbol-print-full-escapes" :dir :system)
-
-#||
-This code has been moved to books/misc/symbol-print-full-escapes.lisp
-at the same time the code was fixed to escape backslashes in the string.
-- Eric McCarthy (mccarthy@kestrel.edu)
+(in-package "ACL2")
 
 (defun bar-escape-chars (x)
   (declare (xargs :guard (character-listp x)))
@@ -61,12 +59,13 @@ at the same time the code was fixed to escape backslashes in the string.
       (coerce (bar-escape-chars (coerce x 'list)) 'string)
     x))
 
-(defun full-escape-symbol (x)
-  (declare (type symbol x))
-  (concatenate 'string "|" (bar-escape-string (symbol-package-name x)) "|::|"
-               (bar-escape-string (symbol-name x)) "|"))
-||#
+(defun bar-escape-name (x)
+  (declare (type string x))
+  (concatenate 'string "|" (bar-escape-string x) "|"))
 
 (defun full-escape-symbol (x)
   (declare (type symbol x))
-  (acl2::full-escape-symbol x))
+  (concatenate 'string
+               (bar-escape-name (symbol-package-name x))
+               "::"
+               (bar-escape-name (symbol-name x))))
--- a/books/xdoc/top.lisp
+++ b/books/xdoc/top.lisp
@@ -58,8 +58,6 @@
            (type string x))
   (bootstrap-revappend-chars-aux x 0 (length x) acc))
 
-
-
 (defun revappend-bar-escape-string (x n xl acc)
   (declare (xargs :mode :program)
            (type string x)
@@ -67,7 +65,7 @@
   (if (eql n xl)
       acc
     (let* ((char (char x n))
-           (acc  (if (eql char #\|)
+           (acc  (if (or (eql char #\|) (eql char #\\))
                      (cons #\\ acc)
                    acc)))
       (revappend-bar-escape-string x (+ n 1) xl (cons char acc)))))
@@ -412,7 +410,7 @@
                       (cons `(local (xdoc::set-default-parents ,name))
                             new-args))
             new-args))
-         
+
          (wrapper
           ;; ACL2 wants an encapsulate to have at least one event, so things
           ;; like (encapsulate nil) cause an error.  To make empty defsections


ericwhitmansmith (commented):
Can you do the code moving and the bug fix in separate commits?

MattKaufmann:
@bendyarm Could you provide an example that shows the undersired behavior that you've fixed?

bendyarm:
> Can you do the code moving and the bug fix in separate commits?

Originally I was intending to do that but I realized that each change causes a lot of recertification so I combined them.  Of course the commits could be separate and in the same PR but at this point why do the commits need to be separated?

bendyarm:
> @bendyarm Could you provide an example that shows the undersired behavior that you've fixed?

Sure.  
Buggy:
```
ACL2 !>(xdoc::full-escape-symbol '| \\ \| |)
"|ACL2|::| \\ \\| |"
```
Fixed:
```
ACL2 !>(xdoc::full-escape-symbol '| \\ \| |)
"|ACL2|::| \\\\ \\| |"
```
I also fixed a similar bug in `revappend-bar-escape-string` but I didn't look to see what that was being used for.

MattKaufmann:
@bendyarm Thanks, but I don't know the spec for full-escape-symbol.  I was hoping for a user-level example, e.g., where :DOC prints something incorrect.  OK, I've just now tried to make my own, given that your new code does something for \\ as was already done for \|.  But I didn't see any difference in the behavior of the two :DOC commands below.  Could you come up with an example like this that shows your fix?

```(include-book "xdoc/init" :dir :system)
(defpointer |AB\|C| io)
:doc |AB\|C|
(defpointer |AB\\C| io)
:doc |AB\\C|

**Comment by ericwhitmansmith** [comment]
> Of course the commits could be separate and in the same PR but at this point why do the commits need to be separated?

Separate commits would let us clearly see what changed vs what was just moved (or refactored).

**Comment by ericwhitmansmith** [comment]
I think the spec for full-escape-symbol is to create a string which, after being printed, can reliably be read in as the symbol.  Yes?

**Comment by bendyarm** [comment]
> I think the spec for full-escape-symbol is to create a string which, after being printed, can reliably be read in as the symbol. Yes?

Yes.  I have written doc for it but will make that a separate PR.
@MattKaufmann  If you do 
```
:trans (defpointer abc |AB\\C|)
```
You will see the bug:
```
...
(MAKE-EVENT-FN
                  '(XDOC::DEFXDOC-FN 'ABC
                                     '(POINTERS)
                                     "See @(see |ACL2|::|AB\\C|)."
                                     NIL NIL NIL STATE)
...
```
after the fix:
```
...
            (MAKE-EVENT-FN
                 '(XDOC::DEFXDOC-FN 'ABC
                                    '(POINTERS)
                                    "See @(see |ACL2|::|AB\\\\C|)."
                                    NIL NIL NIL STATE)
...
```
Also, there is an appeal to first principles,
that if slash is used as an escape character, then a literal slash must be escaped.

**Review by acoglio — APPROVED** [review]
This looks good to me. I agree that, next time, this PR could be split into two commits (running only one regression for both commits), but on the other hand I wouldn't go back and re-do it now; the change looks clear to me. Besides escaping vertical bar with backslash, we also need to escape backslash with backslash.

**Review by gjurgensen — APPROVED** [review]

**Comment by acoglio** [comment]
@bendyarm You might consider adding the commit for the extended doc to this PR. Up to you, of course.

**Review by MattKaufmann — APPROVED** [review]
Thanks for the example.  I see the issue now; thanks, @bendyarm, for the examples.  I'm approving even though I haven't looked at the code (I'm not familiar with it); if there's a problem then it's probably subtle and will come out when the documentation is built.

**Comment by bendyarm** [comment]
> > Of course the commits could be separate and in the same PR but at this point why do the commits need to be separated?
> 
> Separate commits would let us clearly see what changed vs what was just moved (or refactored).

Right, sorry.  The diff is below.  
There is probably a little more string wastage that would get garbage collected
that we could avoid by not using `bar-escape-name` but I don't think it would make much difference.
```
$ cd acl2/books
...
$ git diff master:./xdoc/full-escape-symbol.lisp fix-bar-escapes:./misc/symbol-print-full-escapes.lisp
diff --git a/books/xdoc/full-escape-symbol.lisp b/books/misc/symbol-print-full-escapes.lisp
index e689c41dbe..bb11fea69c 100644
--- a/books/xdoc/full-escape-symbol.lisp
+++ b/books/misc/symbol-print-full-escapes.lisp
@@ -1,3 +1,4 @@
+; Originally part of...
 ; XDOC Documentation System for ACL2
 ; Copyright (C) 2009-2015 Centaur Technology
 ;
@@ -28,7 +29,11 @@
 ;
 ; Original author: Jared Davis <jared@centtech.com>
 
-(in-package "XDOC")
+; In 2022, moved from books/xdoc/full-escape-symbol.lisp
+; and from XDOC to ACL2 package, and fixed escaping of backslash.
+; Eric McCarthy (mccarthy@kestrel.edu).
+
+(in-package "ACL2")
 
 (defun bar-escape-chars (x)
   (declare (xargs :guard (character-listp x)))
@@ -36,6 +41,8 @@
          nil)
         ((eql (car x) #\|)
          (list* #\\ #\| (bar-escape-chars (cdr x))))
+        ((eql (car x) #\\)
+         (list* #\\ #\\ (bar-escape-chars (cdr x))))
         (t
          (cons (car x) (bar-escape-chars (cdr x))))))
 
@@ -46,13 +53,19 @@
 
 (defun bar-escape-string (x)
   (declare (type string x))
-  ;; Dumb optimization: don't need to escape anything unless there's a #\|
+  ;; Dumb optimization: don't need to escape anything unless there's a #\| or #\\
   ;; somewhere.
-  (if (position #\| x)
+  (if (or (position #\| x) (position #\\ x))
       (coerce (bar-escape-chars (coerce x 'list)) 'string)
     x))
 
+(defun bar-escape-name (x)
+  (declare (type string x))
+  (concatenate 'string "|" (bar-escape-string x) "|"))
+
 (defun full-escape-symbol (x)
   (declare (type symbol x))
-  (concatenate 'string "|" (bar-escape-string (symbol-package-name x)) "|::|"
-               (bar-escape-string (symbol-name x)) "|"))
+  (concatenate 'string
+               (bar-escape-name (symbol-package-name x))
+               "::"
+               (bar-escape-name (symbol-name x))))
$ 
```

ericwhitmansmith:
This seems ok to me, mostly based on the example.  I wonder if we should avoid the name bar-escape for the functions that are now also escaping backslashes.

acoglio:
Maybe `bar/bslash-escape-...`? On the other hand, the bar is the main thing, while the backslash comes along for the ride.

bendyarm:
If you look at the CLtL2 table 22-1: Standard Character Syntax Types
you will see that `\` has syntax single escape and `|` has syntax `multiple escape`.
You can enter the symbol with print name `"#A"` with `\#a` or `|#A|`.
I think "bar escaping" is a good term for the latter, so I kept that nomenclature.
The "bar" refers to the type of escaping, which means you surround the name with bars
and then add backslashes preceding both literal bars and literal backslashes.
This info is in my xdoc but I can't add the xdoc to the PR yet because it depends on
other stuff.

ericwhitmansmith (approved):

solswords (approved):

PR #1449, opened 2023-01-01
[R1CS] Remove stale ref & fix typos.

acoglio:


ericwhitmansmith (commented):
I think this PR is no longer necessary.  I added the dense r1cs stuff to the community books, so that comment now makes sense, and I fixed the if->is typo.  (Not sure about un-hyphenating true-list; there seem to be precedents both ways.)  Thanks for pointing out these issues.

PR #1450, opened 2023-01-11
Xdoc include book form

bendyarm:
There are two commits here that do the same thing for two ways of rendering xdoc.   The file name at the top changes so that it is a copyable, executable include-book form.  For example, for the topic `crypto-hdwallet`, this text:
```
kestrel/hdwallet/wallet.lisp :DIR :SYSTEM
```
changes to
```
(include-book "kestrel/hdwallet/wallet" :dir :system)
```

The first commit handles the html rendering.  For example, what you would see at 
https://www.cs.utexas.edu/users/moore/acl2/manuals/latest/index.html?topic=HDWALLET____CRYPTO-HDWALLET

The second commit handles the rendering used by the `:doc` command.

It is possible we will want another commit to handle the rendering used by the `m-x acl2-doc` Emacs command; that is TBD.

### DIFF
--- a/books/xdoc/display.lisp
+++ b/books/xdoc/display.lisp
@@ -693,6 +693,21 @@
                                             state)))
                         (value fal))))))))
 
+;; Corresponds to includeBookFormForFrom in fancy/xdoc.js
+(defun include-book-form-for-from (from)
+  ;; Check the exact pattern expected for `from`, so that the worst that could
+  ;; happen if the pattern changes is to revert to the previous behavior.
+  (if (or (not (stringp from))
+          (let ((from-len (length from)))
+            (or (not (> (length from) 18))
+                (not (equal (subseq from (- from-len 18) from-len)
+                            ".lisp :DIR :SYSTEM")))))
+      from
+    (concatenate 'string
+                 "(include-book \""
+                 (subseq from 0 (- (length from) 18))
+                 "\" :dir :system)")))
+
 (defun topic-to-text (x all-topics state)
   "Returns (MV TEXT STATE)"
   (b* ((name (cdr (assoc :name x)))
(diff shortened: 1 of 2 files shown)


solswords (changes requested):
I'm fine with the format change, but it'd be good if it worked for other :dir entries besides system (i.e. files under the directories of other project-dir-alist entries) and for books outside of any project dirs (which I think would just appear with no :dir).

MattKaufmann:
I completely agree with @solswords .  As I mentioned in a Jan. 6 posting on zulip, I introduced the format we have currently to be able to deal with project directories besides just the system books.  Any change that eliminates this feature would be going backwards.  I'll wait on my review until I'm told that this is fixed.

bendyarm:
@MattKaufmann This change does not eliminate that feature.  Any string that does not end in `.lisp :DIR :SYSTEM` still comes out the same way as it did before these commits.  

I agree that there are other improvements we can make.  Do you have test cases for the other forms?  Hopefully when we make the other improvements, we can make them consistently when setting up `:form` instead of adjusting the strings when outputting the display doc.

acoglio (approved):
This looks fine to me: it turns entries that end in `.lisp :DIR :SYSTEM` into entries with `include-book`, and leave all other entries unchanged. I guess @solswords was proposing to extend this improvement to additional cases, not just things in `:dir :system`. I think it would be generally a good idea, which could be done in this PR or in a separate PR. This PR is taking care of the most common case, I believe.

MattKaufmann:
Even though I slightly prefer the existing format, I don't mind this "include-book" change, provided :dir :system is handled like other project directories.  Otherwise, any potential aesthetic improvement is outweighed by that inconstency, in my opinion.

@bendyarm No, I don't have test cases.  Also, you say each of the following.

  "The second commit handles the rendering used by the :doc command."

  "It is possible we will want another commit to handle the rendering used by the m-x acl2-doc Emacs command; that is TBD."

But the :doc command and acl2-doc use the same rendering, so the second comment doesn't quite make sense to me.  Perhaps you are addressing my comment on zulip a few days ago in response to your query there, where I said: "I'm OK with that change, provided it doesn't break acl2-doc or :DOC at the terminal."  I'll be quite disappointed if there is a push to master that breaks an acl2-doc command (which I think is unlikely, but is possible even if :doc works at the terminal, since acl2-doc does more than just display topics).

PR #1451, opened 2023-01-12
[acl2s] Add some additional proof builder macros to acl2s/utilities

mister-walter:
We've found these proof builder macros useful when automatically generating proof-builder instructions.

Ping @pmanolios


MattKaufmann:
On quick glance those seem very reasonable to me, for what it's worth.  (I'd merge, except I don't think it's my place to merge into testing-acl2s.)

PR #1455, opened 2023-01-18
[defaggregate] Allow a defaggregte to be local in most cases.

ericwhitmansmith:
This avoids putting in (logic) or (program) when not needed, since those prevent the defaggregate from being local.  Defaggregate uses progn instead of encapsulate (see comment in defaggregate.lisp), and ACL2 doesn't allow things like (local (progn (logic))).

### DIFF
--- a/books/std/util/defaggregate.lisp
+++ b/books/std/util/defaggregate.lisp
@@ -1282,9 +1282,12 @@ would have had to call @('(student->fullname x)').  For instance:</p>
                  doc-events
                '())
 
-           ,(if (eq mode :logic)
-                '(logic)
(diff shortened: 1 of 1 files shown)


acoglio:
Regarding [the comment you mention](https://github.com/acl2/acl2/blob/37d50b11702bf1c18ac34ad57a516164c1b3941b/books/std/util/defaggregate.lisp#L1272),  I wonder if this points to a more generally solvable problem (or just a performance improvement) in `encapsulate`.

acoglio (approved):

solswords (approved):

PR #1456, opened 2023-01-19
[Seq] Improve `:s=` and `:w=`.

acoglio:
Instead of returning the original stream if the `mbt` check fails, return the remaining stream from the call that fails the `mbt` check. The advantage is that often the stream processing functions fix the stream to its type, and therefore unconditionally return a remaining stream of the right type. However, if `:s=` and `:w=` return the initial stream, which is not fixed to the right type, then functions using these operators need conditional return theorems. Note that the exact stream returned if the `mbt` check fails does not matter in guard-verified code, because it is provably never returned. But I'm not sure if there are good use cases for returning the original stream instead, when the code is not guard-verified (e.g. in program mode).

### DIFF
--- a/books/misc/seq.lisp
+++ b/books/misc/seq.lisp
@@ -427,7 +427,7 @@ creation of warnings while processing the stream.</p>")
                                         (len !!!stream))))
                              (prog2$ (er hard? "SEQ count failed for (~x0 ~x1.)~%"
                                          ',(car x) ',action)
-                                     (mv "SEQ count failure." nil !!!stream)))
+                                     (mv "SEQ count failure." nil ,stream)))
                             (t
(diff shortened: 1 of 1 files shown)


solswords (approved):
Seems good, assuming the regression passed. I'm not too worried about execution of non-guard-verified code since this case is guarded by both a `(not (mbt ...))` and a `(prog2$ (er hard? ...) ...)` .

acoglio:
Yes, the regression passes. I'm assuming that you've already checked, or will check after this is on master, any Intel or other books that you may have.

And you are right that, even in non-guard-verified code, the `(er ...)` stops execution so the stream is never returned — I should have realized that. So it's really a non-issue.

PR #1457, opened 2023-01-25
[ATC] Disable rules which depend on choice of bit width values

gjurgensen:
These disabled rules are collected in a static theory, 'bit-width-value-choices, so that they may be collectively enabled if one wishes to reason specifically over the chosen bit width values.

In addition, new rules have been added mirroring these now disabled rules (except the unsigned to signed maxima comparisons), but are generalized to arbitrary bit width choices consistent with the C standard.


gjurgensen:
I encountered useless rune failures after making these changes, which I resolved by removing the useless rune files, as per the first advice on the doc page: https://www.cs.utexas.edu/users/moore/acl2/manuals/latest/index.html?topic=ACL2____USELESS-RUNES-FAILURES). Let me know if I should instead attempt to update the useless rune files.

acoglio (approved):

acoglio:
It might good to re-generate the useless runes file, if you are willing:
```
ACL2_USELESS_RUNES=write cert.pl <the file>
```
And then you can add the commit here and we merge.

Issue #1454, opened 2023-01-17
[Proposal] Improve doc and code links in doc


bendyarm:
## 💥 Proposal

Improve documentation links and code links in xdoc.  How do we want them to look?

I see this proposal as a place we can discuss the issues and when there is some consensus we can improve the doc.

If you would prefer to let me know your opinions by a DM (Zulip, etc.) that is also fine.

### Background

In xdoc, if the doc was defined in a community book, under the topic there is currently a code link.  For example, for the doc for `dm::primep`  the code link text looks like
```
kestrel/number-theory/primep-def-doc.lisp :DIR :SYSTEM
```
and the link target looks like
```
https://github.com/acl2/acl2/tree/master/books/kestrel/number-theory/primep-def-doc.lisp
```

There have been some requests for making the link text into an include-book form that you could cut and paste into your program.  One problem with that is that the current source file link goes to the source of the xdoc, not necessarily the source of the code.  Including that source file with include-book will not bring in the definition in the documentation, and so it could be considered a documentation bug. 

### Considerations and Objectives

There are three main ways to read the doc:
1. web interface
2. `:doc` at the ACL2 command line
3. `m-x acl2-doc` in Emacs

The kinds of link targets are currently:
* no link target available (xdoc defined interactively)
* containing file was a system community book; the link text adds `:DIR :SYSTEM` to the relative filename
* containing file was in a project directory in the sense of `project-dir-alist`; the link text adds `:DIR <project-dir>` to the relative filename
* containing file was somewhere else; the link text passes through the file name unchanged and does not add a link target (if the developer wants a link target, they can define a project; see `:doc project-dir-alist`.

Some design principles:
* make the defaults convenient for most developers
* allow developers to override the defaults
* make it clear to users what the links are for

### Questions

Are there more xdoc topics documenting definitions where (A) the xdoc is defined in the same source file as the
definition being documented, or (B) the xdoc is defined in a different source file than the definition being documented?

### Proposal 1

* First, change the source link text by prefixing it with "doc source: ".   This makes the current behavior more clear.
* Add a feature to xdoc-carrying definitions where the developer can add a separate "code source: " link.  It could be called `:code-source`, and would be relative to the same system/project as the xdoc source file.  (*).  We could also allow `:code-source :none` to mean there is no source code to link to.
* Make a list of places where the xdoc describes a definition that is in a different source file, and encourage the owners to add "code source: " links, or make PRs for them to review.
[...]

PR #1459, opened 2023-01-26
defxdoc :code-source definitional support

bendyarm:
Adds the `:code-source` argument to `defxdoc` and `defxdoc+`.    
There is an example of its use in `ubi-doc.lisp`.  
Does not yet push the value into the doc entry and does not yet handle rendering.  
See also a few inline comments in the code and in this PR.


books/kestrel/utilities/ubi-doc.lisp
```diff
@@ -8,6 +8,7 @@
 
 (defxdoc ubi
   :parents (kestrel-utilities history)
+  :code-source "kestrel/utilities/ubi :DIR :SYSTEM"
```
bendyarm, line 11: The first rendering would just make this text into a link with the appropriate link target to github as the doc links currently do.  If we want to allow lowercase `:dir :system` to mean the same thing when rendering, let me know, that would be easy.
Also, eventually I would like to be able to add a `copy include-book form to clipboard` button to the html xdoc rendering that follows the code-source link.

books/system/doc/render-doc-base.lisp
```diff
@@ -236,16 +237,23 @@
...
 ; more convincing reason to make this change.
 
-    (mv (list* (rendered-symbol name)
+    (mv (list ;list*
+               (rendered-symbol name)
                (rendered-symbol-lst parents)
                terminal
+               from
+               #|| EM: review later
                (if (equal from "ACL2 Sources")
```
bendyarm, line 246: I took this out for simplicity, since in my test it bloated only about 0.6%
However, once the downstream rendering is working I will put it back in later.

MattKaufmann:
Just as for the related pull request #1454, I don't want to be involved in this project personally, so please remove me as a reviewer.

Please make sure that :DOC at the terminal or meta-x acl2-doc continue to work with these and subsequent xdoc changes.

books/xdoc/base.lisp
```diff
@@ -118,7 +120,18 @@
                           (cons :parents parents)
                           (cons :short short)
                           (cons :long long)
-                          (cons :from bookname)))
+                          ;; :from is the book where the doc is defined,
+                          ;; (usually relativized to acl2/books and with
+                          ;; :DIR :SYSTEM added)
+                          (cons :from bookname)
+                          ;; :code-from is specified by the xdoc writer,
+                          ;; and is not currently not modified/normalized
```
acoglio, line 128: not currently not => currently not

books/kestrel/utilities/ubi-doc.lisp
acoglio: I think it would be good to use lowercase `:dir :system`.

acoglio (approved):

books/xdoc/base.lisp
solswords: I kind of think we ought to fix this before this enters widespread usage. It would seem strange to me to type in `:code-from "my/book/path :DIR :SYSTEM"` -- I suggest we allow both `:code-from "path"` and `:code-from ("path" :dir :dirname)`, normalize both to full-book-names using ACL2 system functions (parse-book-name, or something like it) and then use xdoc's normalize-bookname to get it into the form we want.

books/xdoc/base.lisp
```diff
@@ -118,7 +120,18 @@
...
+                          ;; so when specified, the writer should self-normalize
+                          ;; it to a form similar to the :from (for now at
+                          ;; least)
+                          ; [Pending change]
+                          ; (cons :code-from code-source)
```
solswords, line 133: why is this still commented out? Shouldn't this change be tested with this uncommented?

books/xdoc/base.lisp
```diff
@@ -72,11 +72,13 @@
       (and long (not (stringp long))
            ":long is not a string (or nil)")
       (and pkg (or (not (stringp pkg)) (equal pkg "") (not (pkg-witness pkg)))
-           ":pkg is not a string that is a known package (or nil)")))
+           ":pkg is not a string that is a known package (or nil)")
+      (and code-source (not (stringp code-source))
+           ":code-source is not a string (or nil)")))
```
solswords, line 77: see my other comment -- I think we should accept `("path" :dir :dirname)` as well as just a string here

solswords (approved):
I guess my code comments here are mostly relative to planned work, not what's done here -- this PR seems pretty harmless in that it just adds the argument to defxdoc/defsection.

books/xdoc/base.lisp
bendyarm: That is a good idea, thanks!

books/kestrel/utilities/ubi-doc.lisp
bendyarm: The proposal now is that if you say
```
  :code-source "ubi.lisp"
```
it would be a path relative to the current book, but if you say
```
  :code-source ("kestrel/utilities/ubi" :dir :system)
```
it would be relative to the community books.

If "ubi.lisp" exists, then either "ubi" or "ubi.lisp" would find it.

books/xdoc/base.lisp
bendyarm: Let's explore this a little bit.  I like allowing `("path" :dir :system)` with `"path"` relative to `"acl2/books"`.

When you write `("path" :dir :dirname)` do you really mean `("path" :dir <project-keyword>)`?
If so, then I think we should interpret this `"path"` relative to the the project directory.
It would be confusing if `"path"` were relative to the current book containing the `defxdoc`.

Since the `:code-source` is supplied by the doc developer for doc purposes only, it is good to have clear rules on how the text is processed and what ends up in the rendered doc.  I think if you say
```
  :code-source ("something" :dir :myproject)
```
then it would be most clear if the `"something"` is passed unchanged (*) to the sysfile-p `(:myproject . "something")`, and then rendered in the html doc as 
```
  Code:  (include-book "something" :dir :myproject)
```

(*) There might be some minor modifications of adding and removing `".lisp"`, since `full-book-name`s are supposed to have the `".lisp"` extensions, but I am mainly referring to the directory of the `"something"` here.

I don't think that this makes code movement harder.  If a doc file is referring to a source file with a path relative to the doc file that ought to remain in the same relative place upon movement, then specify it as the string form, not the list form.

For the string version with a relative path
```
  :code-source "subdir/file"
```
the developer story becomes slightly more complex.  I think it is reasonable if we want this to be interpreted relative to the current book's path, since it saves the doc developer a lot of typing and makes it easier to move code.  If the current book is under a project directory, subdirectory "currdir", then this could still be rendered in the html doc as something like
```
  Code:  (include-book "currdir/subdir/file" :dir :myproject)
```

@solswords @ericwhitmansmith What do you think?

books/xdoc/base.lisp
bendyarm: Because if the xdoc table entry changes now, all the consumers will need to be OK with that, and I was probably missing one when I tried it before since I got errors that were hard to debug.  I'm trying for a staged approach to make it easier.

books/kestrel/utilities/ubi-doc.lisp
ericwhitmansmith: Looks good.  Would just` :code-source "ubi"` also work?

books/xdoc/base.lisp
ericwhitmansmith: This all seems good to me.  I think we should make the `:code-source` as easy as possible to write and maintain (and so support relative paths there).  When rendered, the `include-book` form should, whenever possible, be something that can be used from anywhere (and so should have a `:dir` option whenever possible).

books/kestrel/utilities/ubi-doc.lisp
ericwhitmansmith: Oddly, the email github generated with my comment "Looks good" quoted the old example of the `:code-source` option!  What I think actually looks good is Eric's "The proposal now ...".

bendyarm: @ericwhitmansmith Yes, `:code-source "ubi"` would also work, and because the `defxdoc` is under `acl2/books`, the code info would render as
```
  Code:  (include-book "kestrel/utilities/ubi" :dir :system)
```

books/xdoc/base.lisp
solswords: That sounds good. I could go either way (in the :dir-relative, `("something" :dir :myproject)` case) as to whether we do minimal processing and just produce a string directly from that object, or canonicalize using parse-book-name or something similar and then generate the string.  I think the main differences would be that the former might allow 1) totally bogus paths, and 2) paths relative to include-book-dirs as well as project dirs, I think 1) is disadvantageous. 2) could be nice except that include-book-dirs aren't global, so potentially a user might have to determine what cert.acl2 file (or whatever) to load to get that include-book-dir.  So I guess I'm leaning toward full processing and canonicalization with parse-book-name even for the :dir-relative case.

acoglio:
Should we merge this?

mister-walter:
Is this PR awaiting review from @ericwhitmansmith, or is there more to do?

MattKaufmann:
Please do not accept this PR until after Version 8.6 is released.  That should happen soon (probably within a week, maybe sooner, and definitely this month unless something goes terribly wrong).  I don't want to take a chance on it breaking anything.  I figure that since it's been sitting dormant for several months, it can wait another week or even another month.  Thanks.

MattKaufmann:
By the way, I agree with the thrust of the comment @mister-walter -- it would be good to get this resolved before long (just not till after the 8.6 release).

PR #1460, opened 2023-01-31
apply-renaming-to-hints: handle a few more types of hints

westfold:


ericwhitmansmith:
Thanks for this PR.  For the record, I think it is now covered by the changes we discussed on Zoom today.  So this can be closed if you are satisfied with them.

PR #1461, opened 2023-02-04
[FTY] Enable `alistp` in fixtype macros.

acoglio:
Disabling built-in functions, including `alistp`, reveals that `fty::defprod` relies on `alistp` to be enabled in order for the proofs generated by `fty::defprod` to work. This commit enables `alistp` locally inside the events generated by `fty::defprod`. Enabling this locally to the events makes `fty::defprod` more robust.

### DIFF
--- a/books/centaur/fty/deftypes.lisp
+++ b/books/centaur/fty/deftypes.lisp
@@ -119,6 +119,7 @@
     hons
     open-member-equal-on-list-of-tags
     alistp-compound-recognizer
+    alistp
     ;;  len
     ;; equal-of-plus-one fix
     prod-car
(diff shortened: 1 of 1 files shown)


acoglio:
@solswords I'm not sure if the place where I put the local enable is the best one for this macro. I see that the macro does some theory manipulations, so there might be a better place or structure for this.

solswords (approved):
I'd slightly prefer to add it to deftypes-theory, but NBD either way

acoglio:
@solswords Agreed, thank you for pointing me to that theory. I moved the enablement there. The regression still passes.

PR #1464, opened 2023-02-23
[RAC]: improve error report

voskh0d:
A not that small commits to improve error detection: now if the certification fails we report it, add errors in the parser for duplicate id. Also, fix some memory issues and, in a more general way, made some change to improve safety. Start refactoring to be able to implement more checks in the future.


MattKaufmann:
I'll accept this pull request (as requested by David Russinoff) when it is made to the testing branch instead of master.  See: https://www.cs.utexas.edu/users/moore/acl2/manuals/latest/?topic=ACL2____GITHUB-COMMIT-CODE-USING-PULL-REQUESTS

voskh0d:
L
Oupsie, I've missed this part. This should be good now :)

PR #1465, opened 2023-02-25
[Std/typed-lists] Add a rule and rename another rule.

acoglio:
The new rule backchains from `true-listp` to `symbol-listp` without limit; it is disabled by default.

The renamed rule is one that backchains at most once, and is enabled.

It seems that using the unqualified `true-listp-when-symbol-listp-rewrite` for the first rule is more consistent with other libraries. This is the reason for renaming the second rule in a way that describes its specific characteristics.

### DIFF
--- a/books/std/typed-lists/symbol-listp.lisp
+++ b/books/std/typed-lists/symbol-listp.lisp
@@ -62,8 +62,7 @@ std::deflist).</p>"
     ;; It is disabled by default because true-listp is a common match;
     ;; it can be enabled where needed.
     (implies (symbol-listp x)
-             (true-listp x))
-    :rule-classes ((:rewrite :backchain-limit-lst 1)))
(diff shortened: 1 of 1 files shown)


solswords (approved):

ericwhitmansmith (changes requested):
It looks like the backchain limit didn't actually get removed from the second rule.

acoglio:
Yes, no kidding (!). I just pushed a commit to rectify that.

ericwhitmansmith (approved):

acoglio:
@ericwhitmansmith Now you can rename this:
https://github.com/acl2/acl2/blob/5c5902e96c39239634c6c5fc1490160813c293db/books/kestrel/typed-lists-light/symbol-listp.lisp#L124-L127

PR #1466, opened 2023-02-27
[RAC] Now use C++17

voskh0d:
C++ 17 is now required to compile the RAC parser but in the previous commits, the makefile was still using C++11.


MattKaufmann:
Accepting as requested by David Russinoff.

PR #1467, opened 2023-03-16
Deflist compound recognizer

ericwhitmansmith:
Deflist generates rules like true-listp-when-nat-listp, but these are :compound-recognizer rules, not regular :rewrite rules, as might be expected.  This PR adds "-compound-recognizer" to such rule names.

A future improvement to deflist would be to have it generate the rewrite rules as well (perhaps disabled), now that the names are available.


solswords (approved):
OK -- Seems like a reasonable convention and it looks like you've gone through all the public books to fix the breakage.

books/kestrel/c/atc/defstruct.lisp
```diff
@@ -1427,7 +1427,7 @@
        (fixtype-arrayp-of-fixtype-array-of
         (pack fixtype '-arrayp-of- fixtype '-array-of))
        (valuep-when-fixtype-arrayp (pack 'valuep-when- fixtype '-arrayp))
-       (true-listp-when-fixtype-listp (pack 'true-listp-when- elem-type-listp))
+       (true-listp-when-fixtype-listp (pack 'true-listp-when- elem-type-listp '-compound-recognizer))
```
acoglio, line 1430: I would also change the name of the variable to include `-compound-recognizer` or a slight abbreviation, so we can use `true-listp-when-fixtype-listp` for the variable holding the name of the rewrite rule later, but we can always do this later.

acoglio (approved):

PR #1469, opened 2023-03-22
Modified VWSIM tests

vivekr123:
Modified VWSIM tests to perform rational instead of float comparisons.


MattKaufmann:
I'd like to accept this, but all files except one are marked with "A" rather than "M", so I don't know what would happen.  Will it confuse git if I accept this?

MattKaufmann:
Oh, and I forgot to mention: These files are exactly those from pull request #1468, which was already accepted by @solswords .  Ah, and maybe only the two "Conflicting files" have changed.  What should be done next?

PR #1474, opened 2023-04-03
[ATC] Add quoted theories to some generated defstruct forms.

gjurgensen:
Quoted theories are added to generated forms to reduce proof time.

This commit also adds a static theory to be used by forms which do not include explicit theories. The thought here is that there are many proofs generated by the macro that are already fast, but I don't think we want their success to be dependent on the current user's theory.

Finally, another struct example is added to the `defstruct` tests. This example is large and uses all supported field types (integer members, fixed array members, flexible array members). The `defstruct` is also invoked in an empty theory to show that the macro is independent of the current theory.

### DIFF
--- a/books/kestrel/c/atc/defstruct.lisp
+++ b/books/kestrel/c/atc/defstruct.lisp
@@ -6,6 +6,7 @@
 ; License: A 3-clause BSD license. See the LICENSE file distributed with ACL2.
 ;
 ; Author: Alessandro Coglio (coglio@kestrel.edu)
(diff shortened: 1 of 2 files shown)


gjurgensen:
Oh, I realized that the `defstruct-base` is not actually a static theory right now (it is defined with `deftheory`, not `deftheory-static` as I had meant to). Perhaps I should make that change, as I understand `deftheory-static` to behave in a more consistent manner.

PR #1477, opened 2023-04-16
[file-io-light] define a theory useful for guard-verified file output…

bendyarm:


books/kestrel/file-io-light/output-theory-test.lisp
```diff
@@ -0,0 +1,100 @@
+; Simple regresstion tests for output-theory.lisp
```
acoglio, line 1: regression

books/kestrel/file-io-light/top.lisp
```diff
@@ -77,4 +77,8 @@
 (include-book "file-length-dollar")
 (include-book "file-is-newer-thanp")
 
+;; Books grouped together for a specific purpose:
+(include-book "output-theory")
+(include-book "output-theory-test")
```
acoglio, line 82: I wouldn't include test files into top files. The main purpose of top files is to ensure the absence of name conflicts, but that doesn't matter for tests.

acoglio (approved):
I'm not familiar with these rules, but the PR looks reasonable as far as I can tell.

books/kestrel/file-io-light/top.lisp
bendyarm: Done, thanks!

books/kestrel/file-io-light/output-theory-test.lisp
bendyarm: fixed.

PR #1480, opened 2023-05-03
moved projects/quadratic-reciprocity to projects/numbers

russinoff:


ericwhitmansmith:
I didn't review the changes to Doc's books, but I did look at the changes to the Kestrel books and to books belonging to third parties.  They all look fine, except the changes to the release notes.  I think it would be best to revert these, as those old release notes now talk about the "numbers" directory which did not exist back then.

MattKaufmann:
As far as I'm concerned it would be good to leave the old release notes as is, which will keep them accurate for those old releases, but to add a new release note in :DOC note-8-6-books about the change to the directory's name and contents.  (There's are mentions of quadratic-reciprocity in :DOC note-3-2 and :DOC note-3-2-1, which were written before the note-*-books notes, but I don't intend to change them.)

russinoff:
I don't have any opinion about (or understanding of) any of this.  Is there
something you want me to do?

MattKaufmann:
@russinoff Oh, it looks like you've already done as Eric asked; sorry I missed that.  OK.  I'd like you to add an entry to (defxdoc note-8-6-books ...) in books/doc/relnotes.lisp, but that can be done later.  I'll accept your pull request now.

ericwhitmansmith:
I discussed this with Matt and just pushed a commit to undo the changes to the release notes.

PR #1485, opened 2023-05-15
save-exec doc: Cmd line args will be appended

kazarmy:
Regarding save-exec, this pr clarifies what happens to command line arguments if inert-args is not nil. I think this should be clarified due to a potentially useful use case of having a set of flags in inert-args that are missing only the target file. The relevant ACL2 code for this is apparently https://github.com/acl2/acl2/blob/e8905b0f33e9c5d810ba4ef162175ba969005c22/acl2-init.lisp#L990-L995

This is a pr and not an issue for convenience. I have not attempted to build the docs.


kazarmy:
Sorry about (a) and (b). For (a), apparently I have drawn the wrong conclusion regarding commit ea869c9cfe46f6fda0118574bd6a904b86db82b6, and for (b), I simply forgot about it and have regressed to my usual practice regarding pull requests.

I think I have to open a new pull request since this pull request is merging into `acl2:master` and I don't think there is a way to change that after the pull request has been submitted.

PR #1496, opened 2023-06-07
Fix fastnumio to work with new versions of sbcl on M1/M2

westfold:


MattKaufmann:
I was hoping someone who uses an M1 or M2 Mac, or is at least familiar with the issue here, would accept this pull request.  But if nobody objects, I'll probably accept it on Tuesday (I'm happy to trust Stephen).

westfold:
I am pretty confident in the modification. It only affects #+sbcl code and the files have reasonable test cases for correctness.
The problem was its use of unsupported sbcl internal functions and I replaced these with CommonLisp functions:
sb-bignum::%ashl -> ash
sb-bignum::%digit-logical-shift-right 32 -> ldb (byte 32 32).   (applied to (unsigned-byte 64))

Bignums are essentially arrays of (unsigned-byte 64). Some care needs to be taken when dealing with (unsigned-byte 64) to not make temporary bignums as only (unsigned-byte 60) are implemented as fixnums. For writing, in the original implementation, each 64-bit word is split into two (unsigned-byte 32), each of which is then split into a dynamic-extent array of (unsigned-byte 4). I couldn’t resist the temptation to simplify things by doing this in one step. This gave a 25% speed improvement and temporary bignums were avoided by inlining. Unfortunately ARM-64 sbcl does not stack allocate the dynamic-extent array, although the x86-64 version does, but it is still a significant speed-up: ~2.5 times for ARM-64 vs ~3 times for x86-64 compared to the default output routine.

Stephen

PR #1502, opened 2023-06-26
[cert.pl] add newlines after setting CERT_PL_IMAGES; also turn off msg

bendyarm:
There are two new lines here.

The line `print $mf "\n\n";`
is needed to close off the list of images so the following comment doesn't start at the end of the previous line.

The `print $mf " \;\@:";` line turns off the message
```
make: Nothing to be done for 'all-cert-pl-images'.
```
that those of us that don't use images are now seeing on every call to `cert.pl`.
It does that by appending an empty recipe.

If you use images and you want to see this message when there aren't any,
then it will have to be done differently---in that case
maybe it could be conditioned on one of the image-related environment variables.

### DIFF
--- a/books/build/cert.pl
+++ b/books/build/cert.pl
@@ -1093,9 +1093,6 @@ unless ($no_makefile) {
 	    print $mf " \\\n     " . image_file_path($cert);
 	}
     }
-    # Turn off the "Nothing to be done for 'all-cert-pl-images' warning
-    # when there aren't any images.  No-op if there are images.
-    print $mf " \;\@:";
     # Close out the list of images
     print $mf "\n\n";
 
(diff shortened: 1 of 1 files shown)


solswords:
Obviously the two newlines are needed, but can you explain the  `\;\@:` thing a little more?  I'm fine with eliminating the "Nothing to be done" message, I just don't understand how adding those characters does that.

solswords:
Actually I think I mostly get it -- adding ";@" at the end of the dependencies of all-cert-pl-images ends the dependencies and gives a recipe which is just an empty unechoed shell command.  But what is the colon doing at the end?  Also I'd prefer leaving the CERT_PL_IMAGES variable as-is (a list of image file paths) and just adding the recipe explicitly to all-cert-pl-images, i.e.:
```
all-cert-pl-images : $(CERT_PL_IMAGES)
	@
```

acoglio (approved):
Making the output more concise sounds good to me, but I don't know enough Perl to provide specific comments on this code.

bendyarm:
@solswords The `@`preceding a make action means not to echo,
and the `:` is a no-op recipe.
I will push another commit on the branch to move it to `all-cert-pl-images`.

solswords (approved):
Looks good, thanks!

Issue #1508, opened 2023-07-02
preventing quicklisp bundle definition from reloading


bendyarm:
As a specific example of issue #1507
this form from `quicklisp/base.lisp`
```
(include-raw "bundle/bundle.lisp" :host-readtable t)
```
when reloaded, resets the asdf:already-loaded-systems info,
losing knowledge of which systems were previously loaded,
even though those systems remain in the environment.

So even if you load the asdf system with (asdf:require-system ..)
instead of (asdf:load-system ..) to prevent reloading, 
if this include-raw is reloaded then the require-system
would think it needs to load.

Usually this would not cause any problem since system loading
is generally idempotent, but it is suboptimal.

I will submit a PR for this issue that specifically prevents
the bundle definition from reloading.

If anyone can think of a scenario where you would want
the bundle info to reload in an ACL2 environment,
please comment.

bendyarm:
Here's a transcript that demonstrates the bug.
```
; 1. Load a quicklisp system and then undo.
ACL2 !>(include-book "quicklisp/zippy" :dir :system)
ACL2 !>:u
ACL2 !>:q
* (asdf:already-loaded-systems)
("asdf" "uiop" "asdf-package-system" "zippy" "documentation-utils" ...)
; Note that the :u did not remove the raw loaded asdf "zippy".
* (lp)

; 2. Now load the main quicklisp loader book.
ACL2 !>(include-book "quicklisp/base" :dir :system)
ACL2 !>:q
* (asdf:already-loaded-systems)
("asdf" "uiop" "asdf-package-system")
; ^^ that return value is a bug. ^^
; The raw loaded asdf "zippy" is in the environment but the asdf info was reset.
; E.g.:
* (find-package "ZIPPY")
#<PACKAGE "ZIPPY">
```

PR #1511, opened 2023-07-04
[quicklisp] make the asdf bundle info remember it was loaded and do nothing if reloaded

bendyarm:
Fixes issue #1508 .

After this PR, redoing (include-raw "bundle/bundle.lisp" ..)
(which is called from "quicklisp/base")
does not reset the asdf info since it recognizes it has already been done
in the current environment.
So the second part of the instructions in the comment on issue #1508  becomes:

; 2. Now load the main quicklisp loader book.
ACL2 !>(include-book "quicklisp/base" :dir :system)
ACL2 !>:q
* (asdf:already-loaded-systems)
(... "zippy" ...)

### DIFF
--- a/books/quicklisp/bundle/bundle.lisp
+++ b/books/quicklisp/bundle/bundle.lisp
@@ -5,6 +5,15 @@
   (unless (find-package '#:asdf)
     (error "ASDF could not be required")))
 
+; Prevent the bundle form from reloading.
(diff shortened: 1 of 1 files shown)


westfold (approved):
Looks good to me

solswords (commented):
This general scheme is ok with me, certainly an improvement on the status quo.

PR #1513, opened 2023-07-11
[Std/osets] Minor improvements

acoglio:
- Slightly simplified a theorem (along with two redundant instances of it):
  - In the first case, if `X` satisfies `empty`, its `sfix` is always `nil`.
  - In the second case, we know that `X` does not satisfy `empty`, and thus it always satisfies `setp`, and there's no need to fix it with `sfix`.
- Fixed and updated some documentation.


acoglio:
Merging after Sol gave approval via email.

PR #1515, opened 2023-07-11
[acl2s] Update ACL2s install instructions

mister-walter:
Update the ACL2s install instructions XDOC topic(s).


MattKaufmann:
It's been a couple of days and nobody has complained, so I'm merging this pull request.  (Please let me know if I'm violating protocol.)

PR #1517, opened 2023-07-15
[quicklisp osicat] do not define %open-temporary-file/fd-streams in ccl

bendyarm:
This PR works around the problem that, on CCL, every time you do
(asdf:load-system "osicat"), it needs to recompile and put a new `.lx64fsl`
file in the bundle cache.
This has caused occasional failures with regressions, especially on leeroy.
(I do not see an issue for it here, though.)
There was already an issue for the underlying problem in osicat:
https://github.com/osicat/osicat/issues/37
I added a comment on that issue.

### DIFF
--- a/books/quicklisp/bundle/software/osicat-20220220-git/src/osicat.lisp
+++ b/books/quicklisp/bundle/software/osicat-20220220-git/src/osicat.lisp
@@ -318,7 +318,11 @@ PATHSPEC exists and is a symlink pointing to an existent file."
         (pathname (concatenate 'string system-tmpdir "/")))))
 
 ;; Windows has no MKSTEMP.
-#-windows
+;; 2023-07-14 ccl added here for now to prevent a compilcation error on ccl,
+;; which caused (asdf:load-system "osicat") to always recompile.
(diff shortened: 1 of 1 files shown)


ericwhitmansmith (approved):

solswords (commented):
I don't know much about osicat so I'll assume the actual patch is fine. My one issue is that we need to keep track of this patch and make it easy to reapply (and to remember to reapply) when we update the quicklisp bundle as in update-libs.sh. I don't think we probably care to have update-libs apply it automatically but maybe add a note to what it prints at the end?

bendyarm:
@solswords Good idea!  I added a message to `update-libs.sh` for your review.

solswords (approved):

westfold:
Looks okay, but wouldn't it it clearer to  have the condition be  #+(and osicat-fd-streams (not windows)) like the one usage of %open-temporary-file/fd-streams. Also, could make (defun %open-temporary-file/no-fd-streams ...) be conditioned on  #-(and osicat-fd-streams (not windows)).

bendyarm:
@westfold Good idea, thanks!

westfold (approved):
Looks good

PR #1523, opened 2023-07-24
Adding modular pdf parser and artifacts

lwli11:
Adding the folder 'pdf-parser' into abnf, which contains a modular pdf parser implementation pdf-parser.lisp.


ericwhitmansmith:
How about making a new directory for this pdf work under books/projects/?

lwli11:
Thanks for working me with this merge request! I've moved the directory to books/projects. How does this look?

acoglio (approved):

acoglio:
@lwli11 Thank you for this contribution. Did you make sure that the ACL2 regression passes after these commits, according to the instructions in https://www.cs.utexas.edu/users/moore/acl2/manuals/latest/?topic=ACL2____GITHUB-COMMIT-CODE-USING-PULL-REQUESTS? I'm asking because the other PR, which I merged already, did not certify, and I had to fix a few things to unbreak the testing branch (fixes: https://github.com/acl2/acl2/commit/896c7be3e7ac2c57bce030c4d61f592df429fecb, https://github.com/acl2/acl2/commit/890e99bd166556c2e8b7d6dca587b4101b14b65a, https://github.com/acl2/acl2/commit/4d0edaf13f3f550b10fb1d4040aec902da2b62f5, https://github.com/acl2/acl2/commit/b19a087cb5dfb6116208b440a2ca0610a41fce27).

lwli11:
Thanks for merging in my changes! I only have ACL2s working, which I tested my changes in, so I wasn't able to run make. Sorry for missing those errors.

acoglio:
I see, but we need to make sure that the regression passes before we can merge this. Can you build (plain) ACL2 and run a regression from the branch that you are making this PR from? Since you have the repository already clone, it should be just a matter of a make to build ACL2 and a make for the regression.

lwli11:
I'm working on it, but it looks like installing ACL2 is going to take a while. It seems apt-get install clisp does not find a correct version of common lisp. Working through: https://www.cs.utexas.edu/users/moore/acl2/v8-5/HTML/installation/requirements.html, the gcl instructions don't work, and ccl seems to be the easiest install.

Also, the instructions for build may need to be changed as I got: Preparing to build saved_acl2 (log file make.log).

** ABORTING: Shell output has been redirected to the file make.log .
             But this is illegal, because this "make" invocation
             is already directing its output to that same file.
             To change where this "make" invocation directs its output,
             you may set "make" variable ACL2_MAKE_LOG to that desired
             filename, but you are advised to consider simply avoiding
             the redirection of shell output to make.log.
make[1]: *** [GNUmakefile:302: set-up-log] Error 1
make: *** [GNUmakefile:728: large] Error 2

real	0m0.083s
user	0m0.027s
sys	0m0.003s

lwli11:
Even without my changes, there's an error in make:


ACL2 Error in ACL2-INTERFACE:  When the ACL2 reader tries to read #\x,
then x must either be a single character or one of Space, Tab, Newline,
Page, Rubout, or Return (where case is ignored).  What follows x must
be a character in the list:
(#\Tab #\Newline #\Page #\Space #\" #\' #\( #\) #\; #\` #\,).
However, Space
 is neither a single character nor one of Space, Tab,
Newline, Page, Rubout, or Return (where case is ignored).

> Error: ACL2 Halted
> While executing: #<Anonymous Function #x30200604768F>, in process listener(1).
> Type :POP to abort, :R for a list of available restarts.
> Type :? for other options.
1 > make[2]: Entering directory '/mnt/hgfs/shared/acl2'
Initialization FAILED: acl2-status.txt should contain :INITIALIZED.

MattKaufmann:
@lwli11 Regarding your post about problems obtaining Lisp:

You say that "the gcl instructions don't work".  Could you be more specific?  Just now I followed these instructions from the website you mentioned, https://www.cs.utexas.edu/users/moore/acl2/v8-5/HTML/installation/requirements.html#Obtaining-GCL, and they worked fine on my Intel Mac.

```
git clone git://git.sv.gnu.org/gcl.git
cd gcl/gcl
git checkout Version_2_6_13pre
./configure --enable-ansi && make
```

You also said that "the instructions for build may need to be changed", and you provided the error message (thank you).  Could you say exactly the command you issued that resulted in that error?  If your command was something like make >& make.log, that would explain the error; you need to redirect to a different file, as the error message is attempting to explain.  Feel free to suggest wording that you think would be clearer.

All that said, I think it's good that you're using CCL, as that Lisp and SBCL are by far the ones most often used for building ACL2, as far as I know.  You mentioned clisp, but not that it's not suitable for building ACL2, as I guess you have inferred from the ACL2 installation requirements page mentioned above.

MattKaufmann:
@lwli11 I wonder if your build error is due to somehow related how ACL2 was fetched and differences in operating systems.  How did you fetch ACL2?  What operating system are you running on?

lwli11:
It's possible it's a corporate firewall issue:

git clone git://git.sv.gnu.org/gcl.git
Cloning into 'gcl'...
fatal: unable to connect to git.sv.gnu.org:
git.sv.gnu.org[0: 209.51.188.168]: errno=Connection refused
git.sv.gnu.org[1: 2001:470:142::168]: errno=Network is unreachable

lwli11:
Thanks! It might be a bad git pull. Our firewall often has issues with large repos. If instead I use the tar acl2-8.5.tar.gz, it builds correctly. I'll try the regression test next.

acoglio (changes requested):
Not necessarily a change request, but the regression needs to pass with the latest ACL2 (maybe master can be merged into this branch), not just with the 8.5 release, because things in ACL2 or the libraries may have changed a lot since then.

lwli11:
I cloned the repo, 'make' succeeded, but 'time nice make -j 8 regression-fresh)' has been running from 9am - 5pm and is still not complete. Is this normal?

MattKaufmann:
Whether it's normal depends on your computer.  That does seem like a long time if your computer really does have the 8 threads that I'd expect when you use "-j 8", though maybe not if, say, your computer is a basic sort of machine that's 10 years old.  You could send a log if you have one; that might suggest something.

lwli11:
We have virtual machines to run linux, so that may be a reason why it's so slow. There's progress, but it just takes a long time, and I can' tell if aborting a run makes it start over from the beginning.

Here's a part of the log:
Built /home/l/acl2/books/centaur/fgl/top-doc.cert (77.496s) at 03-Aug-2023 09:42:29.492
Making /home/l/acl2/books/coi/defung/phony-induction.cert on 03-Aug-2023 09:42:29.564
Built /home/l/acl2/books/coi/defung/phony-induction.cert (1.880s) at 03-Aug-2023 09:42:31.467
Built /home/l/acl2/books/projects/rp-rewriter/top.cert (13.068s) at 03-Aug-2023 09:42:31.537
Making /home/l/acl2/books/coi/quantification/quantification.cert on 03-Aug-2023 09:42:31.539
Making /home/l/acl2/books/coi/super-ihs/from-rtl.cert on 03-Aug-2023 09:42:31.606
Built /home/l/acl2/books/coi/super-ihs/from-rtl.cert (2.300s) at 03-Aug-2023 09:42:33.919
Making /home/l/acl2/books/coi/paths/hints.cert on 03-Aug-2023 09:42:33.986
Built /home/l/acl2/books/coi/quantification/quantification.cert (6.102s) at 03-Aug-2023 09:42:37.683
Making /home/l/acl2/books/projects/x86isa/machine/linear-memory.cert on 03-Aug-2023 09:42:37.749

lwli11:
Maybe this is a user preference, but a percentage complete would be helpful, or to know how long the log is supposed to be:
 cat make-regression-finish-1.log | wc -l
13248

Am I getting to the end?

ericwhitmansmith:
Aborting the build and restarting should only make it restart the books that were actively being certified at the moment of the abort.  I like the idea of adding a percentage indicator, but I don't know how to do that with 'make'.  (Also, various books are added or excluded from the build based on which tools are available on the particular machine.)  One thing you could do is look for the number of .cert files, e.g., with 'find . -name "*.cert" | wc -l'.  In my copies of ACL2, there are usually a bit over 9400 .cert files created a by 'make regression-everything'.

MattKaufmann:
Regarding whether "aborting a run makes it start over from the beginning": aborting and then running "make regression" won't recertify books whose certification already completed.

Regarding progress: you're probably pretty far along.  Here's my "wc -l" for a recent run, but I ran "make regression-everything", which certifies a few more books than "make regression":

```
~/acl2/acl2$ wc -l make-regression-everything-ccl-quicklisp-j-16.log
   19073 make-regression-everything-ccl-quicklisp-j-16.log
~/acl2/acl2$ 
```

Regarding "percentage complete": I agree that this would be nice, but few people know enough about the infrastructure for book certification to be able to do this in a reasonable amount of time (and I'm not one of those, and they tend to be full-up with work responsibilities).  I don't want you to be discouraged from making suggestions, though: they often lead to improvements in ACL2 (recent ones are discussed for example in the latest release notes, https://www.cs.utexas.edu/users/moore/acl2/manuals/latest/?topic=ACL2____NOTE-8-6).

MattKaufmann:
[I submitted my comments before I saw Eric's -- and I agree with Eric's.]

ericwhitmansmith:
BTW, the machine used matters a lot.  I was able to do a full everything regression in 52 minutes using 32 cores on a very new machine.  (This also used SBCL, which is faster than CCL.)

MattKaufmann:
Eric, that's amazing!  On my 2019 Intel-based MacBook Pro, certification of books/workshops/2011/krug-et-al/support/MinVisor/setup-nested-page-tables.lisp alone recently took just over 80 minutes (admittedly using CCL and running 16 threads on an 8-core hyperthreaded machine).  That might be the slowest single book, though.  Are you using an M2-based MacBook Pro?

ericwhitmansmith:
This was on an AMD Ryzen Threadripper PRO 3975WX with 32 cores plus hyperthreads.  Actually, I just tried again using 64 simultaneous jobs, which reduced the time to 47 minutes.  The book you mention took about 33 minutes:

Built /home/ewsmith/acl2/books/workshops/2011/krug-et-al/support/MinVisor/setup-nested-page-tables.cert (2009.111s) at 03-Aug-2023 17:50:21.659

On an M2 Macbook it's much faster (about 21 minutes):

Built /Users/ewsmith/acl2/books/workshops/2011/krug-et-al/support/MinVisor/setup-nested-page-tables.cert (1282.563s) at 03-Aug-2023 17:37:59.302

though that machine has many fewer cores.

ericwhitmansmith:
(BTW, I think these build times do not include building the ACL2 executable, just the books.)

lwli11:
make finally finished running if I leave it overnight! However, it seems ACL2 sedan mode admits more theorems than make. Has anyone found a way to identify the hints/lemmas that ACL2s used that I can add so the same theorems admit on command line?

Thanks!

ericwhitmansmith:
I've never used ACL2s, but regular ACL2 prints a summary after each proof that lists all the rules used.  I wonder if ACL2s might have something similar.

mister-walter:
@lwli11 Eric's comment above should hold for ACL2s as well - after each proof, it should print out a summary of all of the rules it used. There are a few other ways that I am aware of where ACL2s might be helpful in proving your theorems over base ACL2:
- ACL2s starts out with some books included, in particular `arithmetic-5`. You'll be able to see this in the list of rules that are printed.
- ACL2s defines some additional "type-like" theorems about certain functions. You'll be able to see this in the list of rules that are printed.
- ACL2s' counterexample generation can sometimes help the theorem prover backtrack when it has made a bad generalization.

Feel free to ping me if you're having trouble getting a theorem that is proved in ACL2s to go through in base ACL2.

lwli11:
@mister-walter Thanks, I just wrapped up some programs so I have more time this fall to finish up this push. I'll try out your suggestions and see if I can get all the functions admitted.

lwli11:
@mister-walter Thanks, I'm just getting back to this, and trying your advice. I took a look at the rules from command-line ACL2 vs ACL2sedan, and found some differences. I've looked all around the acl2 documentation for hints, but didn't see anything about how to indicate which theorem to use.

As an example, this is the ACL2sedan rules used:
Summary
Form:  ( DEFTHM INDEX-OF-SUBSEQ-SELF ...)
Rules: ((:DEFINITION DOUBLE-REWRITE)
        (:DEFINITION INDEX-OF-SUBSEQ)
        (:EXECUTABLE-COUNTERPART EQUAL)
        (:FAKE-RUNE-FOR-TYPE-SET NIL)
        (:REWRITE PREFIXP-WHEN-EQUAL-LENGTHS))
Time:  0.07 seconds (prove: 0.02, print: 0.00, proof tree: 0.00, other: 0.06)
Prover steps counted:  47

And then this is the command-line ACL2 rule output which does not admit:


Form:  ( DEFTHM INDEX-OF-SUBSEQ-SELF ...)
Rules: ((:DEFINITION DOUBLE-REWRITE)
        (:DEFINITION ENDP)
        (:DEFINITION INDEX-OF-SUBSEQ)
        (:DEFINITION NOT)
        (:ELIM CAR-CDR-ELIM)
        (:EXECUTABLE-COUNTERPART EQUAL)
        (:FAKE-RUNE-FOR-TYPE-SET NIL)
        (:INDUCTION INDEX-OF-SUBSEQ)
        (:TYPE-PRESCRIPTION INDEX-OF-SUBSEQ)
        (:TYPE-PRESCRIPTION PREFIXP))

I added arithmetic-5 as a book. Are there any others you would use?

Any resources you can point me through would be appreciated. I think I'm missing a few hints on how to direct the theorem prover in command-line ACL2, and then I can work through all the theorems.

mister-walter:
It looks like the relevant lemma that ACL2 uses to prove this theorem in ACL2s is `PREFIXP-WHEN-EQUAL-LENGTHS`. I used grep to search the ACL2 repo for this lemma - it is part of the `std/lists` books, so you should be able to get the same theorem proved in vanilla ACL2 by adding an appropriate `include-book` form. Try `(include-book "std/lists/top" :dir :system)`.

lwli11:
Thanks @mister-walter ! This is great information, and adding lists and acl2s/base-theory. It'll take a while to work through the full list, but I'll update when it's complete.

lwli11:
@mister-walter : Sorry to bother you again, but if I can't find a book that the command line ACL2 isn't using, but there are rules that are different, is there a way to enable/disable rules in guard-hints? I know of :hands-off, but not how to enable rules. Thanks for all your help, and we should record all this somewhere for anyone else with these same questions.

MattKaufmann:
If you look at :DOC guard-hints, you will see a link to :DOC xargs, which discusses the :guard-hints keyword.  There, you are directed to :DOC hints, which explains the :in-theory hint.  If that doesn't answer your question, then follow the link there to :DOC in-theory.  Oh, and all that documentation (":DOC") can be accessed by using the "Jump to" box near the top of the page, http://www.cs.utexas.edu/users/moore/acl2/manuals/latest/index.html.

lwli11:
I think, the issue is the incredible amount of resources available on ACL2 in the documentation, but which doesn't show up on google. From the name alone, I wouldn't have guessed that in-theory meant enable/disable theorems.

But before I get to that, I tried to combine what I had so far:
   | ACL2 Warning [Compiled file] in ( INCLUDE-BOOK "acl2s/ccg/ccg" ...):
   | Unable to load compiled file for book
   |   /home/l/acl2/books/acl2s/ccg/ccg.lisp
   | because the compiled file does not exist.  See :DOC include-book. 
   | Here is the sequence of books with loads of compiled or expansion files
   | that have led down to the printing of this message, where the load
   | for each is halted during the load for the next:
   | 
   |   /home/l/acl2/books/acl2s/base-theory.lisp
   |   /home/l/acl2/books/acl2s/properties.lisp
   | 
   | 
   | ACL2 Error in ( INCLUDE-BOOK "acl2s/base-theory" ...):  The ttag 
   | :EV-FNCALL-W-OK associated with file /home/l/acl2/books/acl2s/cgen/type.lisp
   | is not among the set of ttags permitted in the current context, specified
   | as follows:
   |   NIL.
   | See :DOC defttag.
   | 

Are ACL2s books allowed to be included without declaring "(in-package "ACL2s")"?

Thanks!

ericwhitmansmith:
It looks like your book depends on something that requires a trust tag.  If so, you need to tell ACL2 that this is okay when certifying your book.  Try creating a file, named <your-book-name>.acl2, containing the following line:

; cert-flags: ? t :ttags :all

See https://www.cs.utexas.edu/users/moore/acl2/manuals/latest/?topic=BUILD____CUSTOM-CERTIFY-BOOK-COMMANDS

P.S.  If you would like more interactive help getting your books to certify, we could hop on a Zoom call.  Just email me if so.

ericwhitmansmith:
(Edited my comment because text within angle brackets got removed!  The new file should be called your-book-name.acl2.)

mister-walter:
That error seems to indicate that the `ccg` book is not certified. The easiest way to do this is by running `books/build/cert.pl books/acl2s/top.lisp` from inside of the ACL2 repo root.

Depending on your setup, you may need to perform some additional steps to get `cert.pl` to work - see the `build::preliminaries` [XDoc topic](https://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/?topic=BUILD____PRELIMINARIES) for more information if you run into any issues.

Separately, if you are trying to certify some books that include ACL2s' books, you will need to tell `cert.pl` that it's OK (because ACL2s books use trust tags) using the technique that Eric discussed above. Typically we will do this by adding a `cert.acl2` file in the same directory as the files to be certified. Here is an example `cert.acl2` from one of our projects, in case it ends up being useful!

```
(include-book "acl2s/portcullis" :dir :system)
(include-book "acl2s/base-theory" :dir :system :ttags :all)
(include-book "acl2s/ccg/ccg" :dir :system :ttags :all)
(include-book "acl2s/custom" :dir :system :ttags :all)
(acl2::acl2s-common-settings)
(include-book "acl2s/acl2s-sigs" :dir :system :ttags :all)

#+acl2-par
(set-waterfall-parallelism nil)

#+acl2-par
(set-parallel-execution nil)

; cert-flags: ? t :ttags :all
```

lwli11:
Thanks, both you of you! I really do appreciate all this step-by-step advice on how to get this book certified, and I had no idea I was going to have to bother you this much. 

I have gotten the ACL2 books to be included, but now not everything that admits in ACL2s compatible mode is accepted when certifying the books. Maybe I'm at the point where this book needs professional help, so I appreciate the offer for an interactive meeting.

lwli11:
Thanks to all your help, I got the file to admit and certify. It's great to finally finish this! Please find the cert.out and make-regression log attached. [make-regression-finish-1.log](https://github.com/acl2/acl2/files/13231356/make-regression-finish-1.log)
[pdf-parser.cert.out.txt](https://github.com/acl2/acl2/files/13231362/pdf-parser.cert.out.txt)

I've learned alot from this conversation and I really appreciate all the time you all took

. Thanks for walking me through acl2 vs acl2s compatibility and all the local books which can be added, and hope to contribute more in the future.

PR #1524, opened 2023-07-26
Magic macroexpand

ericwhitmansmith:


MattKaufmann (approved):
That all looks fine to me (as long as books using magic-macroexpand still certify).  Here are some minor comments (feel free to ignore) on magic-ev-term.lisp.

The following is a minor stylistic oddity: 

```
      (let ((fn (ffn-symb term)))
        (if (eq 'quote fn)
            (mv nil (unquote term))
          ...
```

Normally ffn-symb isn't used on quoted terms, so a preferred style is as follows.

```
      (if (fquotep term)
          (mv nil (unquote term))
        (let ((fn (ffn-symb term)))
          ...
```

Also, usual style for terms is to use fargn instead of car, cadr, etc., e.g., (fargn term 3) instead of (cadddr x).  (I realize that the original magic-ev didn't do that either.)

I'm not sure about the following optimization, even though it's logically correct, since if (cadr x) and (caddr x) are both, say translations of (prog2$ (cw "Hi") a), then "Hi" would only be printed once.

                (if (equal (cadr x) (caddr x))
                    ;; OR -- don't re-evaluate the same term
                    (mv nil test)

ericwhitmansmith:
Thanks, Matt.  I may address some of the stylistic issues in a later commit, though I deliberately avoid fquotep to avoid re-doing the car (I realize any speedup is unlikely to matter in practice).

Note that the optimization about OR is in the original magic-ev, not in my new version.  But, actually, I am considering including it, since it would prevent the first argument of an OR from bring evaluated twice, thus avoiding double side effects in that case.  I don't see any way to tell whether something of the form (IF x x y) actually came from (OR x y).  Is there any better way of handling this?

PR #1536, opened 2023-09-05
[ihs] Drop redundant guard from logtail.

ericwhitmansmith:
Equivalent claims are in the extended formals.


ericwhitmansmith:
Forget to say: I'll merge this in a few days if nobody objects.

acoglio (approved):

PR #1538, opened 2023-09-13
[jenkins] list osicat cache dir and quicklisp/base.* files before and…

bendyarm:
… after every books build, to get more information about where the osicat bundle cache binaries might be going missing


acoglio (approved):
I'm not very familiar with this, but it looks good as far as I can see.

ericwhitmansmith:
This looked good to me, so I merged it.

PR #1540, opened 2023-09-14
Removes hard-coded python name from expt example config

chchen:
Example 2 shows how to extend smtlink's functionality with a custom python module. Unfortunately, the custom configuration also hardcodes the python program name, which is user-configurable. This change restores the expected behaviour, which uses the user-defined python executable with the custom module.


MattKaufmann:
I'm happy to trust the change, but it shouldn't be merged into master; the usual thing would be to merge into testing.  See :DOC github-commit-code-using-pull-requests.

chchen:
Happy to resubmit the right way.

PR #1543, opened 2023-09-28
Fmt support

ericwhitmansmith:
Speeds up a few proofs in fmt-support.  Also changes the whitespace, but that was done in separate commits.


MattKaufmann:
Looks great.  Did you do an acl2-devel build and run "make devel-check", as described in a comment in *system-verify-guards-alist*?  If not, would you please do that (or send me an email asking me to do it)?

MattKaufmann:
I'm accepting after email from Eric confirming the devel check.

PR #1550, opened 2023-10-22
adding cert_pl_exclude

ankitku:
So that the testing branch is no longer broken


ericwhitmansmith:
Can these books be made to certify in standard ACL2 (rather than ACL2s), perhaps using the suggestions in the comment here:

https://github.com/acl2/acl2/commit/9955384cfb295c18d7045c6e10d5837ef420c72e

?

If not, I am afraid that this directory will bit rot, because it won't be tested (unless it is part of some ACL2s test suite that I don't know about).

PR #1552, opened 2023-10-25
Fix ankits book (part 2)

maxvonhippel:
In which I delete cert.image.


ericwhitmansmith:
I merged this since testing is already broken, so it couldn't hurt.  But I'm not optimistic that things will work now.  My own testing discovered several other issues with these files, including mismatched parens and a function being given the wrong number of arguments.

ericwhitmansmith:
It looks like a cert.acl2 file may be needed in the attacks/ directory as well.  And there are files in that directory that are not certifiable books (they do not start with in-package forms, and they contain things like :q and calls of time$); these should probably be renamed to have .lsp extensions.

ericwhitmansmith:
More work is needed to get these books to work.  For example:

demo.lisp has an extra right-paren after update-msgs-state2.  Also, the string "State Definitions" can't appear at the top level of a book.

unittests.lisp calls the function params on 18 arguments but it requires 26

scoring-filecoin.lisp gives this error:
   | ACL2s Error in CHECK=: Check failed (values not equal).
   | First value:  T
   | Second value: NIL

ankitku:
Sorry, some of these files were not meant to be pushed. I am working on a fix for these issues.

maxvonhippel:
Ok, I'll leave it to @ankitku to fix the remaining issues since he knows this code better than I do.

PR #1557, opened 2023-11-12
Update README to include DRLA workshop books

rubengamboa:


MattKaufmann:
Normally pull requests should target testing, not master.  In this case I imagine it's fine to accept into master, since it's just a text file -- but I'll leave that to someone who is ocnfident about that.

ericwhitmansmith:
I changed the base to testing and will now merge.

PR #1561, opened 2023-11-13
[walter-acl2-workshop-2023] Remove extended package prefix syntax from acl2s-setup.lisp

mister-walter:
I removed [extended package prefix syntax](https://www.sbcl.org/manual/#Extended-Package-Prefix-Syntax) from acl2s-setup.lisp, as it picked up in the regression-everything build that Leeroy performs with CCL. CCL does not support extended package prefix syntax, which is why this caused a failure on Leeroy.

I also added a comment in the README that SBCL is required for building this system locally, as the non-ACL2-book parts of the system use this syntax extensively.

### DIFF
--- a/books/workshops/2023/walter-etal/README.md
+++ b/books/workshops/2023/walter-etal/README.md
@@ -49,11 +49,17 @@ example as noted previously in this README.
 ## Running the CLI locally
 
 ### Prerequisites
+- ACL2 built using SBCL (NOT CCL)
 - Java 17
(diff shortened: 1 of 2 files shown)


MattKaufmann:
ACL2 has something like extended package prefix syntax that might serve your purposes; see https://www.cs.utexas.edu/users/moore/acl2/manuals/latest/?topic=ACL2____SHARP-BANG-READER.

PR #1562, opened 2023-11-17
[X86ISA] Fix typo in code.

acoglio:
For the double-precision case, the result must be turned into a 64-bit quantity.

This issue was noticed by Eric Smith.

I checked that against the Intel instruction manual.


shigoel (approved):
LGTM! Thanks!

PR #1563, opened 2023-11-18
[X86ISA] Fix two typos in code.

acoglio:
These are similar to the one in a previous commit.

Eric Smith first noticed these.

I checked these against the Intel manual.


shigoel (approved):
Thanks!

PR #1565, opened 2023-11-21
adding MIT License

ankitku:


MattKaufmann:
Normally pull requests should target the testing branch, but I'll accept this one that targets master since it doesn't cause any certification.  Thanks, @ankitku .

PR #1569, opened 2023-12-01
Quicklisp changes to support zippy

bendyarm:
Update the quicklisp bundle, except for osicat (due to a compile problem) and a small reversion to dexador.
The fixes in bz3 and zippy were needed when using zippy to do random access to a jar file.


bendyarm:
After tests run by @acoglio, merged by request from @ericwhitmansmith .

PR #1581, opened 2024-03-01
[xdoc] Fix guards on XDOC `extend-links-fal`

mister-walter:
The guards on the `extend-links-fal` function in `books/xdoc/importance.lisp` were inconsistent with comments in that file, and the actual types of arguments that were being passed in. This caused `xdoc::save` to fail for me. I updated the guards so that they are consistent with comments and the actual types of arguments being passed in.

I'm not 100% sure why this wasn't caught by anyone else previously - maybe something about the environment in which I call `xdoc::save`?

### DIFF
--- a/books/xdoc/importance.lisp
+++ b/books/xdoc/importance.lisp
@@ -185,8 +185,8 @@
    targets   ; string list, keys of the pages linked to from source
    links-fal ; the fal being constructed
    )
-  (declare (xargs :guard (and (stringp source)
-                              (symbol-listp targets))))
+  (declare (xargs :guard (and (symbolp source)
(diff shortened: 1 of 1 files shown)


ericwhitmansmith:
BTW, I tried doing (set-guard-checking :all) before building the manual, and it does cause the guard violation that Andrew is seeing and that his patch fixes.

ericwhitmansmith:
Okay, I was able to build the whole manual with the patch, even with (set-guard-checking :all).  So I am going to merge this.

ericwhitmansmith:
Nevermind.  Matt merged it an hour ago.

PR #1583, opened 2024-03-02
[Std/osets] Rename `empty` to `emptyp`.

acoglio:
The new name is in line with the typical Lisp/ACL2 convention for predicates. The old name has caused confusion to some users/readers.

This change was proposed some time ago on the ACL2 Books mailing list, and there were no objections.


ericwhitmansmith:
Seems like a good change to me!  And the changes to my files look good.

solswords (approved):

PR #1588, opened 2024-03-08
[xdoc] Part 1 of XDOC fancy viewer improvements

mister-walter:
This PR contains the first batch of improvements to XDOC's fancy viewer implementation. I am submitting these improvements in smaller batches to (hopefully) reduce the burden on anyone reviewing these changes.

This batch contains a fix to long-broken Google Fonts links and contains minor updates to the list of allowed entities (see the [Entities XDOC topic](https://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/?topic=XDOC____ENTITIES)). This PR should not change the functionality of XDOC's fancy viewer.

The bulk of this PR consists of refactoring of the `xdata` and `xindex` data structures as well as the code surrounding XSLT, in preparation for additional refactoring of the main XDOC viewer code. I added JSDoc-style documentation to the majority of the functions that I added or substantially changed.

I removed two libraries:
- LazyLoad.js, which is unmaintained and was easy enough to get rid of
- jquery.base64.js, which was easily replaced with JavaScript's built-in `atob` function. `atob` has been supported for a while now, and was introduced prior to some of the problematic features listed in **Browser Support Changes** in any case.

### Browser Support Changes
The changes I have made here have resulted in dropped support for Internet Explorer 11 and below, Chrome v41 and lower, Safari 10 and lower, Edge 13 and lower, and Firefox 44 and lower. The problematic features for supporting earlier browser versions are:
- [The Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)
- [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
- `const` and `let`

I think that the Fetch API is the "least supported" of the above features, and[ the browser data from caniuse.com](https://caniuse.com/fetch) indicates that 97.31% of the users users tracked by caniuse use browsers with support for the Fetch API.

It may be possible to polyfill Fetch and Promises (that is, load a library that provides these features on older browsers), but this is not possible for `const` and `let`. We could instead add a transpilation step using something like [Babel](https://babeljs.io/) to translate the XDOC fancy viewer source code into something that works on older browsers.

If supporting these old browser versions is highly valued, I can look into eliminating the usage of the above features, though there is a readability cost to doing so. I hope to eliminate the use of JQuery at some point in the future, which is much easier to do if I have access to more modern JavaScript functionality.


ericwhitmansmith:
Thanks for doing this!  I am probably not qualified to review the javascript changes, but I did have a question about browser compatibility.  I seems that we are only losing support for browers older than the 2015-2016 timeframe.  Is that right?  If so, that seems fine with me as browsers that old should probably not be used for security reasons anyway.

MattKaufmann:
I see that you changed function wrapXdocFragment in file
books/xdoc/fancy/xslt.js.  A comment just above its definition is as
follows, but it doesn't look like you made changes as suggested in
that comment.

```
// Warning: Keep this in sync with *entity-strings* and
// *entitytok-as-plaintext-fal* in ../parse-xml.lisp,
// *xml-entity-stuff* in ../prepare-topic.lisp, and
// (defxdoc entities ...) in topics.lisp.
```

solswords:
Thanks for working on this! I'm in favor of modernizing the javascript and getting rid of unnecessary library dependencies, etc., and I'm not too broken up about losing support for 8+-year-old browsers.  I imagine you're aware of both ways of serving xdoc pages, described in [this topic](https://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/?topic=XDOC____DEPLOYING-MANUALS) -- serving all the data at once vs serving it lazily from a SQLite database.  Have you tested your changes with both types of servers?

mister-walter:
@MattKaufmann thanks for pointing that out - I've converted this PR to a draft until I ensure all the entity data is in sync.

mister-walter:
@solswords I tested in three ways:
- Loading a built manual directly (e.g. by using "Open File" in a browser)
- Loading a built manual via a local HTTP server
- Loading a built manual, while pointing it at the `https://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/xdataget.cgi` server. This requires some finagling to get around CORS issues (I use the "CORS Everywhere" Firefox add-on).

I haven't deployed the Perl server locally yet for additional testing. I probably will want to do that at some point, so I might as well do this now.

mister-walter:
I have made sure the entity data is in sync across the various locations mentioned in that comment in the `xslt.js` file, and I also added a Dockerfile that makes it easier to run an XDOC server locally for testing. Should be ready for review.

mister-walter:
The main place where the entity data needed updating was `xdata2html.pl`, indeed the rest of the locations were in-sync (modulo the reordering to make it easier to compare across them)

MattKaufmann:
Ah, thanks for the explanation.  It makes me wonder if a similar "Keep in sync" comment should be added in other places.  My head's not in this, but if you think that makes sense and care to add such comments, that would be great.

mister-walter:
Great point, I'll add `xdata2html.pl` as one of the files that should be updated to the comments in the other places where entity data is specified.

mister-walter:
I updated those comments to refer to `xdata2html.pl` as well.

MattKaufmann:
Excellent; thank you!  Does anyone object to accepting this pull request?  (I don't know enough about the "fancy" manual to be someone who is suitable to review it.)

mister-walter:
I deployed the version of the fancy manual in this PR at http://xdoc.atwalter.com/. This deployment uses the lazy SQLite serving of topics. I will ensure that this deployment is kept up-to-date with my future XDOC PRs.

mister-walter:
Let me know if there's anything I can do to push this PR along/make it easier to review!

MattKaufmann:
On March 13 I commented, "Does anyone object to accepting this pull request?"  Nobody has objected, so I plan to accept it tomorrow or Wednesday if nobody objects first.  As I commented earlier, I'm not a suitable reviewer; however, I see that Sol has done some testing and that's good enough for me if nobody objects.

solswords:
I wasn't the one who tested it, but I'm fine with accepting.

MattKaufmann:
Oops.  I see now that it was Drew who did the testing -- anyhow, I'm still happy to accept the pull request.

PR #1591, opened 2024-04-25
[acl2s] Replace ACL2s' PRNG with the Thrust PRNG

mister-walter:
I replaced the PRNG used by ACL2s with the Thrust PRNG, which is designed to work with 63-bit state (the seed) and 64-bit outputs. My implementation is an adaptation of Tommy Ettinger's C implementation, which Tommy dedicated to the public domain. The implementation is here: https://gist.github.com/tommyettinger/e6d3e8816da79b45bfe582384c2fe14a

Some quick-and-dirty benchmarking indicates that this implementation is about 2x faster than the previous one.

I also needed to update some of the theorems that were being proved about `genrandom-seed` and `genrandom-state`, as they often used `car` to access the first element of the MV that those functions return. I replaced `car` with `mv-nth 0` in those theorems, which eliminated failures that were occurring in random-state.lisp.

I tested this on Linux x86-64 with SBCL 2.4.3. I will mark this PR as a draft until my local regression-all completes.

### DIFF
--- a/books/acl2s/defdata/random-state-basis1.lisp
+++ b/books/acl2s/defdata/random-state-basis1.lisp
@@ -8,18 +8,16 @@
 (set-verify-guards-eagerness 2)
 (include-book "std/util/bstar" :dir :system)
 (include-book "acl2s/utilities" :dir :system)
+(include-book "ihs/basic-definitions" :dir :system)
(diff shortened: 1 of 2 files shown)


mister-walter:
This passed `regression-everything` on SBCL and `regression` on CCL on Linux.

PR #1595, opened 2024-05-28
Add test cases for c$ preprocessing utility

gjurgensen:
Since the preprocessing utility is open-ended in its system dependencies, I've added the cert_param flag "uses-preprocessor" as well as the Makefile variable "USE_PREPROCESSOR" to control certification of books using the tool, such as this new test book.

The feature does not attempt to auto-detect dependencies, and defaults to "off". Thus, you must explicitly set `USE_PREPROCESSOR=1` when invoking make to build the relevant books.

The names of these variables is easy to change. We might consider renaming it to "USE_C_PREPROCESSOR" to be more specific. Or "USE_C_UTILITIES" if we expect other similar utilities to require conditional certification.

### DIFF
--- a/books/kestrel/c/syntax/preprocess-file.lisp
+++ b/books/kestrel/c/syntax/preprocess-file.lisp
@@ -54,8 +54,8 @@
    (xdoc::p
      "These tools appeal to a configurable C preprocessor, and the prerequisite
       dependencies may not be present on all systems. For books which use the
-      default preprocessor \"cpp\", certification may controlled with the the
-      @(see build::uses-cpp) @(see build::cert_param) flag.")
+      default preprocessor \"cpp\", certification may be controlled with the
+      the @(see build::uses-cpp) @(see build::cert_param) flag.")
    (xdoc::p
      "The community books Makefile will autodetect whether \"cpp\" is
       available, and exclude books appropriately. Certification of such books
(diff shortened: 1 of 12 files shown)


acoglio (changes requested):
Since 'preprocessor' sounds generic, I suggest using 'cpreprocessor' for clarity/specificity, in the Makefile, environment variable, and certification flag.

The rest looks fine.

acoglio:
Adding two more reviewers since this affects the books makefile as well. @solswords @ericwhitmansmith

solswords (approved):

gjurgensen:
@acoglio 
> Since 'preprocessor' sounds generic, I suggest using 'cpreprocessor' for clarity/specificity, in the Makefile, environment variable, and certification flag.

I changed the names to "uses-c-preprocessor" and "USE_C_PREPROCESSOR".

gjurgensen:
For simplicity, I switched to "uses-cpp" and "OS_HAS_CPP", which is set by autodetecting availability of "cpp".

books/kestrel/c/syntax/preprocess-file.lisp
```diff
@@ -51,7 +50,22 @@
   :long
   (xdoc::topstring
    (xdoc::p
-     "A collection of tools to invoke an external C preprocessor."))
+     "A collection of tools to invoke an external C preprocessor.")
+   (xdoc::p
+     "These tools appeal to a configurable C preprocessor, and the prerequisite
+      dependencies may not be present on all systems. For books which use the
+      default preprocessor \"cpp\", certification may controlled with the the
```
acoglio, line 57:
```suggestion
      default preprocessor \"cpp\", certification may be controlled with the the
```

acoglio (approved):

ericwhitmansmith (approved):

PR #1601, opened 2024-07-07
Add `:verify-guards :after-returns` option to `defines`

gjurgensen:
This mirrors the existing option provided by `define`.

The only notable implementation change is that `make-fn-defsection` is modified to return two sets of events. The produced `defsection` is split right after the returns theorems (if `:returns-no-induct t` and `:verify-guards :after-returns`) to allow for the `verify-guards` event to be place before the functions' additional events.

### DIFF
--- a/books/std/util/defines.lisp
+++ b/books/std/util/defines.lisp
@@ -28,6 +28,7 @@
 ;
 ; Original authors: Jared Davis <jared@centtech.com>
 ;                   Sol Swords <sswords@centtech.com>
+; Modified by Grant Jurgensen <grant@kestrel.edu> to add :after-returns support
 
 (in-package "STD")
 (include-book "define")
@@ -163,7 +164,15 @@ failures.  You can enable @(':verbosep t') for better debugging.</dd>
 
 
 <dt>:mode</dt>
-<dt>:guard-hints, :guard-debug, :verify-guards</dt>
+<dt>:guard-hints, :guard-debug</dt>
+
+<dt>:verify-guards val</dt>
+
+<dd>The value @('val') may be one of the following: @('t'), @('nil'), or
+@(':after-returns'). The first two correspond to what is described in @(see
+xargs). The keyword @(':after-returns') indicates that the guards of the
+functions are to be verified after the @(see returns-specifiers).</dd>
+
 <dt>:well-founded-relation, :measure-debug, :hints, :otf-flg</dt>
 <dt>:ruler-extenders</dt>
 
@@ -431,7 +440,7 @@ encapsulate), and is mainly meant as a tool for macro developers.</dd>
 ;;     (append (defguts->rest-events (car gutslist))
 ;;             (collect-rest-events-from-guts (cdr gutslist)))))
 
-(defun make-fn-defsection (guts cliquename process-returns)
+(defun make-fn-defsections (guts cliquename process-returns divide-defsections)
   (b* (((defguts guts) guts)
        (short      (getarg :short          nil guts.kwd-alist))
        (long       (getarg :long           nil guts.kwd-alist))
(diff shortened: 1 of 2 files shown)


books/std/util/defines.lisp
```diff
@@ -733,6 +768,12 @@ encapsulate), and is mainly meant as a tool for macro developers.</dd>
        ((unless (consp (cdr gutslist)))
         (raise "Error in ~x0: expected more than one function." name))
 
+       (verify-guards-after-returns
+         (eq (cdr (assoc :verify-guards kwd-alist)) :after-returns))
+       (kwd-alist (if verify-guards-after-returns
+                      (put-assoc :verify-guards 'nil kwd-alist)
```
solswords, line 774: I'd prefer that the kwd-alist stored in the defines-guts isn't changed like this -- if defines was called with :verify-guards :after-returns then it should reflect that

books/std/util/defines.lisp
```diff
@@ -440,34 +449,59 @@ encapsulate), and is mainly meant as a tool for macro developers.</dd>
...
-(defun collect-fn-defsections (gutslist cliquename process-returns)
+                   (let* ((world (w state))
+                          (events (returnspec-thms ',guts.name
+                                                   ',guts.name-fn
+                                                   ',guts.returnspecs
+                                                   world)))
+                     `(with-output :stack :pop (progn . ,events))))))
+        ,@(and (not divide-defsections)
+               rest-events))
+     `(,@(and divide-defsections
```
solswords, line 477: Do we have to divide it into two defsections to support putting verify-guards as a separate event?  Why not insert the verify-guards into the one defsection, between the returnspec-thms and the rest-events?

solswords (commented):
I'd rather keep it to one defsection if possible, but if there's a reason for it I can be convinced.

books/std/util/defines.lisp
gjurgensen: Sure, I can avoid modifying `kwd-alist` and just do the `put-assoc` when calling `get-xargs-from-kwd-alist`.

books/std/util/defines.lisp
gjurgensen: I think we need a single `verify-guards` after all of the returnspec-thms (from all the functions in the `defines`) but before their collective rest-events.

To be concrete, say we have a `defines` likes this:
```lisp
(defines foo
  :returns-no-induct t
  :verify-guards :after-returns
  (define foo (...)
    ...
    ///
    foo-other-events)
  (define bar (...)
    ...
    ///
    bar-other-events))
```
If we don't split the defsections, we'll generate:
```lisp
...
(defsection foo
  ...
  foo-return-thms
  foo-other-events
  ...)
(defsection bar
  ...
  bar-return-thms
  bar-other-events
  ...)
...
```
We want to put the `(verify-guards foo)` (verifying guards for both `foo` and `bar`) after `foo-return-thms` and `bar-return-thms`, but before `foo-other-events` and `bar-other-events`. With the current order of things, we can't do that. Hence why I thought we should generate (just when `:returns-no-induct t` and `:verify-guards :after-returns`):
```lisp
(defsection foo
  ...
  foo-return-thms)
(defsection bar
  ...
  bar-return-thms)
(verify-guards foo)
(defsection foo
  :extension foo
  foo-other-events
  ...)
(defsection bar
  :extension bar
  bar-other-events
  ...)
```

solswords: I see, that makes sense. How about additionally checking whether the function has rest-events at all before producing a second defsection for it? I think it might be fairly rare for the individual functions in the clique to have rest-events.

gjurgensen: Oh good point. I pushed a change to only divide sections if there are rest-events.

gjurgensen:
I realized the verify-guards event was being placed below all of the individual rest-events when `:returns-no-induct nil`. Fixed in the most recent commit.

solswords (approved):

PR #1612, opened 2024-07-24
[pdf-parser] Drastically speed up certification of pdf-parser

mister-walter:
Add hints for CCG that drastically speed up termination analysis for some functions in projects/pdf-parser/pdf-parser. Should resolve regression timeouts related to this book.

### DIFF
--- a/books/projects/pdf-parser/pdf-parser.lisp
+++ b/books/projects/pdf-parser/pdf-parser.lisp
@@ -1020,6 +1020,7 @@
 
 (defund parse-objects (prefix contents )
         (declare (xargs :guard (character-listp contents)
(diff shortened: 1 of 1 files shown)


MattKaufmann:
@mister-walter Thanks for doing this.  I usually put in comments when I change others' books.  For example, you could say something like "; Updated 7/24/2024 by Drew Walter to speed up CCG analysis".  Would you be willing to do that?

mister-walter:
Will do!

MattKaufmann:
To @lwli11 -- would you please add copyright and license information to books/projects/pdf-parser/pdf-parser.lisp or, if you prefer, the README.md file in that directory?  You can find lots of examples in the books, for example at the top of books/projects/apply-model/apply.lisp.  Thanks.

PR #1614, opened 2024-07-25
[c2c] Add `split-fn` transformation

gjurgensen:
This transformation splits a function after a certain number of block items, factoring out the remainder of the function into a new function.

It also provides a couple test cases. The transformation is not yet fully robust. I expect it may currently fail when multiple variables are declared at once, when variables are declared but not immediately initialized, or when variables are declared with type/storage specifiers that cannot or should not be duplicated in the new function's parameter declarations.

### DIFF
--- a/books/kestrel/c/transformation/split-fn.lisp
+++ b/books/kestrel/c/transformation/split-fn.lisp
@@ -32,7 +32,7 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (defxdoc+ split-fn
-  :parents (utilities)
+  :parents (transformation-tools)
   :short "A C-to-C transformation to split a function in two."
   :long
   (xdoc::topstring
--- a/books/kestrel/c/transformation/split-fn.lisp
+++ b/books/kestrel/c/transformation/split-fn.lisp
@@ -86,7 +86,7 @@
 
 (define expr-ident-list
   ((idents ident-listp))
-  :short "Map @(tsee expr-list) over a list."
+  :short "Map @(tsee c$::expr-list) over a list."
   :returns (exprs expr-listp)
   (if (endp idents)
       nil
@@ -204,7 +204,7 @@
       declarations of the original function which appear free in the block
       items, which will constitute the body of the new function. (Note that
       arguments are not currently taken by reference, but this may be necessary
-      for general correctness. It may be sufficient to take an argument by
+      for general equivalence. It may be sufficient to take an argument by
       reference when its address is taken in an expression of the new function
       body.)"))
   :returns (mv er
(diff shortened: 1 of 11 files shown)


books/kestrel/c/transformation/split-fn.lisp
```diff
@@ -0,0 +1,444 @@
...
+   (pointers tyqual-list-listp)
+   (items block-item-listp)
+   (decls ident-decl-mapp))
+  :returns (mv er
+               (idents ident-listp)
+               (new-fn fundefp))
+  (b* (((reterr) nil (c$::irr-fundef))
+       (idents (free-vars-block-item-list items nil))
+       (decls (ident-decls-map-filter decls idents))
+       (idents (strip-cars decls))
```
acoglio, line 168: It's better to use `(omap::keys decls)`, since `decl` is an omap.

books/kestrel/c/transformation/split-fn.lisp
```diff
@@ -0,0 +1,444 @@
...
+(define decl-to-paramdecl
+  ((decl declp))
+  :returns (mv erp
+               (paramdecl paramdeclp))
+  (b* (((reterr) (c$::irr-paramdecl)))
+    (decl-case
+      decl
+      :decl (b* (((when (endp decl.init))
+                  (reterr t))
+                 ((initdeclor initdeclor) (first decl.init)))
```
acoglio, line 107: Since it's only using `(first decl.init)`, I suppose that there may be an assumption that the `decl` passed as input should declare just one identifier? It would be good to document this kind of things.

books/kestrel/c/transformation/split-fn.lisp
```diff
@@ -0,0 +1,457 @@
...
+(defxdoc+ split-fn
+  :parents (transformation-tools)
+  :short "A C-to-C transformation to split a function in two."
+  :long
+  (xdoc::topstring
+    (xdoc::p
+      "This transformation takes the identifier of the function it is to split,
+       the name of the new function to be generated, and the location of the
+       split, represented as a natural number corresponding to the number of
+       statement block items in the function body before the split.")
```
acoglio, line 43: Is this the number of block items that are statements, or just the number of block items?

books/kestrel/c/transformation/split-fn.lisp
```diff
@@ -0,0 +1,457 @@
...
+    (paramdeclor-case
+      paramdecl.decl
+      :declor (omap::update (declor-get-ident paramdecl.decl.unwrap)
+                            (paramdecl-fix paramdecl)
+                            nil)
+      :otherwise nil)))
+
+(define paramdecl-list-to-ident-paramdecl-map
+  ((paramdecls paramdecl-listp))
+  :short "Fold @(tsee decl-to-ident-paramdecl-map) over a list."
```
acoglio, line 119:
```suggestion
  :short "Fold @(tsee paramdecl-to-ident-paramdecl-map) over a list."
```

acoglio (approved):

PR #1615, opened 2024-08-05
[doc] Update the pre-built-binary-distributions XDOC topic

mister-walter:
Remove an old broken link to ACL2s, update that section to link to the ACL2s installation XDOC topics.

Add a section on Docker images, populate with a few of mine.

Add a link to the Homebrew package, and update the MacPorts reference to link to the appropriate package.

Should resolve the remaining issue from #1602.


MattKaufmann:
I'm about to accept this pull request (thanks @mister-walter), but I wonder about the following.  The instructions formerly mentioned how to fetch a "Windows system that contains a pre-built binary of (pure) ACL2".  With a quick read of the changes, I didn't see any replacement way to get a Windows image -- but maybe that's just because of my ignorance about Windows.  If a Windows user can still obtain a pure ACL2 image by fetching ACL2s stuff, it might be good to mention that.

mister-walter:
Note that we no longer provide binaries for Windows, instead opting to use Windows Subsystem for Linux to run what is basically a normal Linux ACL2 installation inside of Windows.

PR #1617, opened 2024-08-07
[doc] Remove broken ACL2s links, update ACL2s windows-installation text

mister-walter:
Remove broken links to the old ACL2s website and replace with links to the acl2-sedan XDOC topic.

Update the reference to ACL2s in the windows-installation XDOC topic.

Remove text couching WSL from windows-installation XDOC topic (as we have used it successfully for ACL2s).


MattKaufmann:
Thanks, @mister-walter !  It's great that these topics have been updated (by you).

PR #1620, opened 2024-08-23
Fixes for merging updated x86isa model into acl2

yaso9:
This has two commits. One updated support-axe.lisp and marks books/projects/x86isa/virtualization/top.lisp as ccl-only. These are necessary to get the regression to succeed.

The other one marks both the concrete and abstract stobjs created by bigmem-asymmetric as non-executable. They each require 6 GB of memory, so this change drastically reduces the amount of memory necessary to certify books. This wasn't done before because :non-executable was not supported on abstract stobjs until recently. I can't certify the doc book on my machine without this change, but it shouldn't be necessary on a machine with more memory.

This has been tested on ACL2 built with CCL.

CC: @solswords @ericwhitmansmith


ericwhitmansmith:
The Axe change looks good to me, thanks.

PR #1628, opened 2024-09-11
[acl2s] Tell macOS users to preemptively re-sign Eclipse

mister-walter:
In the installation instructions for ACL2s, we now tell macOS users to preemptively re-sign the Eclipse app after installing the ACL2s plugin, to make this issue more visible to users.


pmanolios:
Thanks

PR #1631, opened 2024-09-16
Make bigmem attachable and allow one to attach bigmem-asymmetric

yaso9:
*Note*: This makes some trivial changes to Kestrel's books and some slightly more involved changes to Intel's books. I think someone from Intel may want to approve these.

This change is larger than one might naively expect because to allow for
attaching bigmem-asymmetric to bigmem, the logical definitions of the
stobjs must be the same (i.e. the same symbol). Previously, these
logical definitions, while identical, were repeated in each book under
their own package.

Thus, a centaur/bigmems/logic book was created with the logic
definitions. The bigmem(-asymmetric) directories were moved into that
directory as well for better organization and the doc topics were also
updated to reflect this hierarchy. The stobjs were also updated to use
the shared definitions in the logic book.

Thanks to Matt for implementing attachable stobjs.


MattKaufmann:
Looks great to me!  I'd accept it now, but I see that @ericwhitmansmith was a suggested reviewer and I see the comment that "someone from Intel may want to approve these" (maybe @solswords ).  So I'll assume that one of those two two will accept it after the other has commented.

solswords (approved):
From what I see there shouldn't be any disruption to our usage of the Centaur/Intel libraries.

ericwhitmansmith (approved):
Looks fine to me except that I see a line being added twice to books/kestrel/big-data/packages.lisp.

yaso9:
My bad, that is now fixed.

PR #1634, opened 2024-10-03
Update `x86isa` debugging tools and fix bt(s/r) instructions

yaso9:
- Before this fix, if the RIP pointed to unmapped memory the virtualization based
debugging tool would segfault.
- `bt`, `bts`, and `btr` instructions now decode their operand properly when the
bitBase is one of the r8-15(d) registers.

See the relevant commit messages for more information.


MattKaufmann:
About to merge, as requested by Warren Hunt.

PR #1635, opened 2024-10-04
Use query params for xkey in SEO friendly xdoc

yaso9:
The SEO friendly version of xdoc put the xkey in the path component of
the url (i.e. URLs would look like `index-seo.php/ACL2___TOP`). While
this is more idiomatic than having them be a query parameter, this
caused mod_evasive to recognize each page as separate and thus ratelimit
them separately instead of together.

This moves the xkey to an `xkey` query parameter and updates the URLs
accordingly.

### DIFF
--- a/books/xdoc/fancy/index-seo.php
+++ b/books/xdoc/fancy/index-seo.php
@@ -7,7 +7,7 @@
    $index = 'DOCINDEX'; // index page
    
    $basedir=pathinfo($_SERVER["SCRIPT_NAME"],PATHINFO_DIRNAME)."/";
-   $scriptdir=$_SERVER["SCRIPT_NAME"].'/';
+   $scriptname=$_SERVER["SCRIPT_NAME"];
    error_reporting(E_ALL);
    ini_set('display_errors', 1);
    
@@ -38,8 +38,8 @@
       exit();
    }
    
-   if(isset($_SERVER["PATH_INFO"]) && (strlen($_SERVER["PATH_INFO"])>1))
-        $key = pathinfo($_SERVER["PATH_INFO"],PATHINFO_BASENAME);
+   if(isset($_GET['xkey']) && (strlen($_GET['xkey'])>1))
+        $key = $_GET['xkey'];
    else $key = $top;
    $showIndex=false;
    if($key=="*") {$showIndex=true;$key=$index;unset($_GET['path']);}
@@ -54,12 +54,12 @@
      $page = $ret->fetchArray(SQLITE3_ASSOC);
    }else 
    if(!$showIndex){
-     $preferedURL=$scriptdir.$page['XKEY'];
+     $preferedURL=xkey_url($page['XKEY']);
      header("Content-Location: ".$preferedURL);
    }
    
    if(!$page) { // fallback when not even the 404 page was found
-     header("Location: ".$scriptdir);
+     header("Location: ".$scriptname);
      die();
    }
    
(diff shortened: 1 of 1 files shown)


MattKaufmann:
I have two questions below about possible changes I need to make.  Whoever answers, please bear in mind that I know essentially nothing about "seo" or other non-user-level web stuff -- others (mostly, David Rager) have told me what to do.  So if there are changes I need to make, I'd appreciate being told explicitly what they are -- no explanation necessary.  Thanks in advance!

By the way: I think this pull request may be necessary (due to some recent changes at UT CS) and I hope it's accepted today -- but I'd prefer that additional changes be postponed till after the ACL2 Version 8.6 release, which I may get to any time after today.

(1) Most evenings, at my UT CS account, I run the script bin/make-fancy-manual.sh (which is in github) to put up a new copy of the manual at the usual place, http://www.cs.utexas.edu/users/moore/acl2/manuals/latest/index.html. Is there anything in that script that needs to change?  In particular, there are these lines.

```
# Create version of the manual optimized for search engines
# (instructions from David Rager):
pushd $books/doc/manual
chmod +x ./xdata2sql4seo.pl
perl xdata2sql.pl
./xdata2sql4seo.pl
chmod o+r ./xdata-seo.db
chmod ugo+r xdata.db
chmod ugo+rx index-seo.php
popd
```

(2) I have a long set of (private) instructions for the release, which includes the following.  Does any of it need to change (other than replacing "v8-5" by "v8-6")?  Note that I've been doing this step for many years.

```
# --- Rager says the following about the manual (email 6/20/14).

#     Perform the following steps from David Rager, for the combined
#     manual, in
#     /u/www/users/moore/acl2/v8-5/combined-manual/
#     but first:
#       cp -p /u/www/users/moore/acl2/v8-5/combined-manual/index-seo.php /u/www/users/moore/acl2/v8-5/combined-manual/index-seo.php.orig-released
#       cp -p /u/www/users/moore/acl2/v8-5/combined-manual/sitemap.xml /u/www/users/moore/acl2/v8-5/combined-manual/sitemap.xml.orig-released
#       chmod o-rx /u/www/users/moore/acl2/v8-5/combined-manual/index-seo.php.orig-released
#       chmod o-rx /u/www/users/moore/acl2/v8-5/combined-manual/sitemap.xml.orig-released

#     Now edit index-seo.php and sitemap.xml as requested below by
#     Rager.  (As is clear below, the two files are under directory
#     /u/www/users/moore/acl2/v8-5/combined-manual/.  Note that
#     "three" seems to be for both files put together: one reference
#     in one file, two in the other.  Also note that instead of
#     current/combined-manual we should look for current/manual.))

#        There are three absolute references to
#        "http://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/"
#        in the files doc/manual/index-seo.php and doc/manual/sitemap.xml.  After you've
#        copied over the manual to the official release directory (
#        "http://www.cs.utexas.edu/users/moore/acl2/v8-5/combined-manual/"),
#        change the
#        link in the released files index-seo.php and sitemap.xml to reference
#        the manual for that particular release (
#        "http://www.cs.utexas.edu/users/moore/acl2/v8-5/combined-manual/").
```

I appreciate the help!

yaso9:
While I don't know much about the release or deployment procedure, after looking over these instructions, I don't think this change shouldn't affect any of them.

mister-walter:
I also don't have any knowledge regarding deployment outside of what I've picked up through reading the XDOC sources, but based on the knowledge I do have, I agree with @yaso9 above.

PR #1636, opened 2024-10-04
Update XDOC Dockerfile to add support for PHP

yaso9:
While this change was made to support the development of #1635, it is independent from it.

CC: @mister-walter

The old XDOC fancy viewer Dockerfile did not support PHP. The seo friendly version of the manual relies on PHP. This adds support for PHP so that one can use the Dockerfile to test the seo friendly XDOC viewer.

### DIFF
--- a/books/xdoc/fancy/Dockerfile
+++ b/books/xdoc/fancy/Dockerfile
@@ -1,6 +1,7 @@
-FROM httpd:bookworm
+FROM php:apache-bookworm
 
 # Copyright (C) 2024 Andrew Walter
+# Copyright (C) 2024 Yahya Sohail
 #
 # License: (An MIT/X11-style license)
 #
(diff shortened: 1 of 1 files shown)


mister-walter:
Thanks @yaso9! I'll test tomorrow and post an update here afterwards but barring any issues with that, this PR has my seal of approval :+1:

mister-walter:
After testing, looks good!

If you have time, it would be great if you could add a comment to the Dockerfile saying something about how to access the SEO-friendly version, e.g. expressing that one just needs to navigate to http://localhost:8080/xdoc/index-seo.php.

To be honest I was unaware that the SEO-friendly version of XDOC was available in this way; I had assumed that `xdata2html.pl` was used to generate SEO-friendly HTML versions of all of the pages, which were then hosted as static files.

yaso9:
I've added the comment.

I really do wish the seo-friendly xdoc was just a bunch of static HTML files. It would probably be far lighter than what we have right now. I thought about replacing it with something like that (instead of #1635), but I decided that was more work than I was willing to put in at the moment and I wanted the seo friendly xdoc up as soon as possible. It's not the most straightforward since, as it stands, the breadcrumbs are handled server side, so making it static would require moving that logic client side.

What would be really nice would be to unify the seo friendly xdoc (which is sort of a second class citizen at the moment) with the fancy xdoc viewer. It could be a folder of static files that do the initial load from the server and then handle all subsequent routing client side, giving us the best of both worlds. That is a lot of work though and what we have right now works well enough, so I doubt we'll see that any time soon.

mister-walter:
The existing `books/xdoc/fancy/xdata2html.pl` script generates individual HTML files for each topic in the xdata/xindex files, though they are not formatted as nicely as the regular viewer.

I have some in-progress improvements to the fancy XDOC viewer that I've been meaning to prepare for PRing, so maybe once that's done I can take a look at improving the SEO-friendly version/merging it with the regular fancy viewer. I'll see if I can find time soon.

MattKaufmann:
Please see my comments on PR #1635.

mister-walter:
Note that this PR only modifies a file I wrote for use in debugging/development, and shouldn't affect deployment of XDOC in any way.

PR #1637, opened 2024-10-09
Speed up certification of x86isa's x87.lisp (fxsave/fxrstor defs)

yaso9:
The x86isa/machine/instructions/x87.lisp book, which contains
definitions for the instructions fxsave and fxrstor, was very slow to
certify (~15 minutes on my machine with CCL). This change disables some
runes and adds some lemmas to make this much faster (~200 seconds).

Thanks to Eric for suggesting I look at the reported splitter runes.

### DIFF
--- a/books/projects/x86isa/machine/decoding-and-spec-utils.lisp
+++ b/books/projects/x86isa/machine/decoding-and-spec-utils.lisp
@@ -2201,7 +2201,29 @@ reference made from privilege level 3.</blockquote>"
                      (the (unsigned-byte 8) (prefixes->opr prefixes)))))
         (if (= cs.d 1)
             (if p3? 2 4)
-          (if p3? 4 2))))))
+          (if p3? 4 2)))))
+  ///
+  (defthm select-operand-size-range
+          (and (<= 1 (select-operand-size proc-mode byte-operand? rex-byte imm?
+                                          prefixes default64? ignore-rex? ignore-p3-64? x86))
+               (<= (select-operand-size proc-mode byte-operand? rex-byte imm?
+                                        prefixes default64? ignore-rex? ignore-p3-64? x86)
+                   8))
+          :rule-classes :linear)
+
+ (defthm select-operand-size-values
+         (or (equal (select-operand-size proc-mode byte-operand? rex-byte
+                                         imm? prefixes default64? ignore-rex? ignore-p3-64? x86)
+                    1)
+             (equal (select-operand-size proc-mode byte-operand? rex-byte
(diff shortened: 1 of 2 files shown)


MattKaufmann:
The defthm select-operand-size-values looks like it's stored as a rewrite rule, but it's a very odd rewrite rule (as OR calls expand to IF calls).

MattKaufmann:
Sorry, I should have added that if the rule doesn't get in the way for the book it's in, then making it local would be a reasonable enough change.

yaso9:
It's definitely an odd rule. It probably won't ever fire normally. I should probably just make it a `defthmd`, so it's disabled by default and can be used in `:use` hints (which is how I used it). It can't be local since it is used outside of that book (at the moment, only in x87.lisp, but I wouldn't be surprised if it finds use elsewhere in the future).

I'm not sure if there's a better way to write this rule. I wrote it because I had a goal in a guard proof that looked like:
```lisp
(implies (and ...snip...
              (not (equal 1 (select-operand-size ...)))
              (not (equal 2 (select-operand-size ...)))
              (not (equal 4 (select-operand-size ...)))
              (not (equal 8 (select-operand-size ...)))
              ...snip...)
         (equal 32 (select-operand-size ...)))
```
This is vacuously true by `select-operand-size-values`. The other way I tried to encode this property is using `member` or `member-equal`. Neither approach fired automatically, but that's probably a better way to write this lemma. I suspect it would work in this context too.

I'd love to hear suggestions, especially if it's possible to get this to fire automatically.

yaso9:
Oh it's probably also relevant to add the above goal was generated because a function being called with argument `x` bound to `(select-operand-size ...)` that had a guard including something like `(member x '(1 2 4 8 ... 32 ...))`.

ericwhitmansmith:
I haven't looked at the theorem, but here some ideas based on the discussion above:  Maybe you can define (or re-use) a predicate operand-sizep and keep it closed most of the time.  Or you could introduce a rule to strengthen an assumption like (member x '(1 2 4 8 ... 32 ...)) by removing an impossible value (e.g., when you know x is not some constant that is member of the list, like 4).  The rule might fire multiple times and eventually rewrite the member call to false.

MattKaufmann:
I'm accepting this PR now, since it seems like a nice change and I'm about to start the final release process.  I think a change to defthmd would be fine, or if it's not used as a rewrite rule, then just make it :rule-classes nil.  But none of this seems like a big deal; better to get this in now and it can be changed in github later.

yaso9:
Sounds good. I'll look into it later then and see if it can be further improved.

PR #1641, opened 2024-10-18
[RAC] Replace getopt by basic shell

voskh0d:
getopt on Mac and linux does not have the same behavior so, and rewrite it shell.


MattKaufmann:
Merging at David Russinoff's request.

PR #1643, opened 2024-11-11
[books] add "make clean_manual"

bendyarm:
When doing xdoc development, having to do `make clean` followed by `make manual`
takes a long time since too much is deleted and rebuilt.  
  Usually it is sufficient to remove the manual files and do `make manual`.
This PR adds a `make clean_manual` target to speed up the xdoc 
edit/build/view cycle.

### DIFF
--- a/books/GNUmakefile
+++ b/books/GNUmakefile
@@ -492,7 +492,7 @@ endif
 
 QUICKLISP_DIR := quicklisp
 
-.PHONY: quicklisp
+.PHONY: quicklisp manual
 
 ifeq ($(REALLY_USE_QUICKLISP), 0)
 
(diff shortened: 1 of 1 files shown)


acoglio (approved):
Good idea! I'm not a `make` expert, but the code looks fine as far as I can tell.

bendyarm:
> The name "clean_manual" suggests to me that the manual is only being cleaned, not built. I wonder if "rebuild_manual" would be a more descriptive name. (But I can live with "clean_manual", since I'll get used to it if I use it.)

How about `make new_manual`, does that work?

MattKaufmann:
Yes, `make new_manual` seems fine to me.  Thanks.

solswords (approved):
I don't have any objections, but I don't get why cleaning should be necessary. Are there missing dependencies so that something doesn't get rebuilt when it should?

bendyarm:
@solswords Thanks for asking that question!

In the past I was frustrated with not seeing my xdoc changes in the local html
manual after doing 'make manual', even after refreshing the browser.
A brief examination led me to believe the html manual construction process was
hard to describe with dependencies.  So for awhile my dev cycle became
'make moreclean' followed by 'make manual', which was too slow.
That's when I decided to try cleaning less stuff to see if it could rebuild
the local html manual, leading to this PR (PR#1643).

However, when I now do a simple test, just changing a single xdoc string
and doing 'make manual', the local html manual gets rebuilt fine.
I'm closing this PR and I will revisit the matter if I see the original
problem again.

PR #1647, opened 2024-11-16
x86 inst bugfixes

yaso9:
- **Fix bts/r instructions truncating results to incorrect number of bits**
- **Fix shld-spec computing incorrect result**


acoglio (approved):

ericwhitmansmith:
FWIW, all my tests/proofs still work after these fixes.

PR #1648, opened 2024-11-18
Implement more instructions in `x86isa`

yaso9:
- **Add BTC x86 instruction**
- **Implement BSWAP x86 instruction**
- **Implement movhlps x86 instruction**
- **Implement packuswb x86 instruction**
- **Implement PSLL, PSRA, PSRL x86 instructions**
- **Implement movq x86 instruction**
- **Implement punpckh and punpckl x86 instructions**
- **Implement pcmpeq x86 instruction**
- **Implement the PSHUF x86 instruction**


books/projects/x86isa/machine/instructions/bit.lisp
```diff
@@ -777,4 +777,46 @@
...
+  :guard-hints (("Goal" :in-theory (disable reg-indexp unsigned-byte-p)))
+
+  :body
+
+  (b* (;; Only 32 and 64 bit registers are allowed
+       ((the (integer 2 8) operand-size)
+        (select-operand-size
+         proc-mode nil rex-byte nil prefixes nil nil nil x86))
+
+       ;; SDM says undefined behavior for 16-bit register
```
acoglio, line 803: What is "SDM"? Can you spell it out?

books/projects/x86isa/machine/instructions/bit.lisp
```diff
@@ -777,4 +777,46 @@
...
+       ((when (equal operand-size 2)) (!!ms-fresh :bswap-16-bit))
```
acoglio, line 804: The Intel manual says that the result is undefined in this case. I believe that, ideally, we should handle this similarly to how undefined flags are handled after certain operations. But maybe for now we could add a comment saying all of this. Would you be willing to add that comment?

books/projects/x86isa/machine/instructions/fp/mov.lisp
```diff
@@ -744,6 +744,75 @@
...
+       (seg-reg (select-segment-register proc-mode p2 p4? mod r/m sib x86))
+
+       (inst-ac? ;; Exceptions Type 5
+        t)
+       ((mv flg0
+            (the (unsigned-byte 128) xmm)
+            (the (integer 0 4) increment-RIP-by)
+            (the (signed-byte 64) ?addr)
+            x86)
+        (x86-operand-from-modr/m-and-sib-bytes proc-mode
```
acoglio, line 780: This function reads from memory unless Mod = 11. Does the decoder enforce that? If not, we should read directly from xmm2, obtaining the register index from the R/M bits.

books/projects/x86isa/machine/instructions/bit.lisp
yaso9: By SDM I meant Intel's Software Developer Manual. The acronym SDM is used in comments throughout the code, but I think it generally specifies "Intel's SDM." I'll update it to mention "Intel's SDM."

books/projects/x86isa/machine/instructions/bit.lisp
yaso9: Using the mechanism used for undefined flags etc. is definitely the proper way to do it, but I am not very familiar with that mechanism. I'll look into it and use that if it isn't too difficult. Otherwise, I'll add the comment.

books/projects/x86isa/machine/instructions/fp/mov.lisp
yaso9: Line 2749 of `inst-listing.lisp` requires that Mod = 0x3

acoglio (changes requested):
These look good overall. I left some comments.

books/projects/x86isa/machine/instructions/fp/simd-integer.lisp
```diff
@@ -342,4 +342,91 @@
...
+        (add-to-*ip proc-mode temp-rip increment-rip-by x86))
+       ((when flg) (!!ms-fresh :rip-increment-error flg))
+
+       ;; Ensure the instruction is not too long.
+       (badlength? (check-instruction-length start-rip temp-rip 0))
+       ((when badlength?)
+        (!!fault-fresh :gp 0 :instruction-length badlength?)) ;; #GP(0)
+
+       ;; Calculate the result.
+       (result (packuswb src1 src2)) ; unreachable
```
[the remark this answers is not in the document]
yaso9: I suspect this is a copy pasta error. I must have copied another instruction implementation, which had a `cond` there with an unreachable branch. When I removed the `cond` I probably left the comment on accident.

books/projects/x86isa/machine/instructions/pcmp.lisp
```diff
@@ -0,0 +1,149 @@
+; X86ISA Library
```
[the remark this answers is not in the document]
yaso9: This new instruction definition implements `pcmpeq{b,w,q}`, but it is only used for `pcmpeq{w,q}`, I think because I didn't notice `pcmpeqb` was already implemented until I was done. I think it would be reasonable to use this implementation for `pcmpeqb` as well and then remove the other definition.

yaso9:
I think this addresses everything.

Some notes:

- I've unified the `pcmpeqb` implementation with the other sizes using my implementation for all of them. This *may* require some work to accommodate if you have proofs depending on the old definition of `pcmpeqb`.
- I now generate `psll`, `psrl`, and `psra` using a macro since the definitions are similar.

books/projects/x86isa/machine/instructions/fp/simd-integer.lisp
```diff
@@ -429,4 +429,550 @@
        ;; Update the instruction pointer.
        (x86 (write-*ip proc-mode temp-rip x86)))
       x86))
+
+(define psll ((result-width natp)
```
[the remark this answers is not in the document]
acoglio: Yes, it's also surprising to me that they use a whole xmm register for the count.

books/projects/x86isa/machine/instructions/pcmp.lisp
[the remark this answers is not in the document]
acoglio: Good idea. The new definition is simpler, and more uniform.

acoglio (approved):

PR #1650, opened 2024-11-24
Add `b*` binder to introduce `macrolet`s

gjurgensen:
This new binder, `patbind-macro`, is based on the existing `patbind-fun` binder, but expands to a `macrolet` instead of an `flet`. It accepts forms like so:

```lisp
(b* (((macro (name arg1 ... argn)) body-form))
  result-form)
```

which expands to:

```lisp
(macrolet ((name (arg1 ... argn) body-form))
  result-form)
```

Like `patbind-fun`, it allows an optional `inline`/`notinline` declaration, even though the documentation suggests this is unlikely to have any effect.


solswords (approved):
Looks good, thanks!

PR #1651, opened 2024-12-02
Add theorem and theory invariant to osets.

acoglio:


gjurgensen (approved):
LGTM.

gjurgensen:
I'll mention that the prevailing style in osets seems to be capitalizing variables representing sets, but plenty of theorems are already breaking that convention.

acoglio:
> I'll mention that the prevailing style in osets seems to be capitalizing variables representing sets, but plenty of theorems are already breaking that convention.

Yes, I noticed that, but I was actually thinking of proposing a PR to lowercase all of them at some point 😄

ericwhitmansmith:
LGTM. (I see that you are matching the existing naming scheme, which could be improved later.)

ericwhitmansmith (approved):

PR #1653, opened 2024-12-10
[X86ISA] Reduce dependence on other-non-det.lisp.

ericwhitmansmith:
This reduces the dependence on other-non-det, which has a ttag and is only used for a few instructions that currently live in machine/instructions/top.lisp.  Future commits could reduce mentions of the :other-non-det ttag.  This is a step towards building x86-related tools (e.g., Axe) without needing (so many) ttags.

### DIFF
--- a/books/projects/x86isa/machine/decoding-and-spec-utils.lisp
+++ b/books/projects/x86isa/machine/decoding-and-spec-utils.lisp
@@ -43,8 +43,12 @@
 
 (in-package "X86ISA")
 
-(include-book "other-non-det"
-              :ttags (:include-raw :undef-flg :syscall-exec :other-non-det))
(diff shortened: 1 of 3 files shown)


yaso9:
The trust tags used by x86isa are generally only needed for execution. Maybe it's worthwhile to have the model only include the logical definitions (without ttags) and then have a book which includes the model and all the executable definitions (perhaps `tools/execution/top.lisp`). This would allow reasoning without using ttags to include the raw definitions.

acoglio (approved):

ericwhitmansmith:
That would be great.  It would break the normal pattern where foo.lisp calls include-raw on foo-raw.lsp.  But maybe that is just a convention and the include-raw commands could be done much later (or never, when execution is not needed).

PR #1654, opened 2024-12-10
x86isa: Add a listing of implemented/unimplemented instructions by category

solswords:
This mostly adds a file (x86isa/machine/catalogue.lisp) that builds a set of xdoc pages based on the categories of instructions listed in the Intel SDM volume 1 chapter 5, describing which instructions are implemented/unimplemented in the x86 model.  These listings are based on the opcode maps, so they are incomplete when the opcode maps are incomplete, but I've done my best to note where they are incomplete. It also has notes about what I think are some bugs in the opcode maps.

(The change to cert.pl is incidental, nothing to do with the rest of the PR.)

I added a bit of code to inst-listing.lisp to allow printing opcodes more completely, similarly to how they're shown in the SDM.  This could be moved (along with the rest of the xdoc generation stuff) to a separate file so that most of the model wouldn't need to depend on it.


solswords:
@yaso9

yaso9:
Very cool. Was the information compiled by hand from the manual? Must have been a ton of tedious work.

acoglio (approved):

ericwhitmansmith (approved):
Very impressive!  Looks great!

PR #1657, opened 2024-12-13
[X86ISA] Drop many unnecessary mentions of the :other-non-det ttag.

ericwhitmansmith:


acoglio (approved):

yaso9 (approved):

ericwhitmansmith:
My understanding is that if you give a :ttags argument to include-book, it must include all the relevant ttags (or be :all) but you can choose to supply no :ttags argument at all (which is what I always do).

I was going to ask, separately from this commit, whether we think having explicit :ttags in the include-books is worth it.  It does suppress some warnings (but I still see ttag notes).  If we decide it's worth it, we could add :ttag arguments to the other include-books in the project (found by Sol's grep command), for consistency.

PR #1659, opened 2024-12-13
X86 -- fix some opcode map bugs, add more categories to instruction catalogue, split files

solswords:
Unless anyone objects to splitting out the xdoc generation portion of inst-listing.lisp, what mainly needs review here is the changes to the opcode maps, which would be pretty tedious to check thoroughly. You could spot check a few of them against either the SDM or, say, XED datafiles ([here](https://github.com/intelxed/xed.git)). I think the changes are unlikely to break existing proofs/executions, because they're mainly just in the mnemonics/features, with a few more substantive changes to instructions that we don't yet implement anyway.


solswords:
It's easier to see the changes in the opcode maps by looking at the diff of inst-listing in just the [first commit](https://github.com/acl2/acl2/pull/1659/commits/0a007e917a9ac0603879ec46e1066fa21188ef4d), rather than the "Files Changed" since that includes all the cut/pasting between files.

yaso9 (approved):

yaso9:
I didn't check exhaustively, but the few changes I sampled to look through checked out.

acoglio (approved):

ericwhitmansmith:
I can confirm that all of Kestrel's proofs still work after these changes.

ericwhitmansmith (approved):

PR #1661, opened 2024-12-17
[arithmetic-5] Change find-constant-addend to avoid certain loops.

ericwhitmansmith:
See the comment I added in the file.  Previously, for (equal 5 (+ 0 m)), the rule would add -5 to both sides, which could begin a loop if certain rules are disabled.  It seems better to refrain and hope that other rules simplify the (+ 0 m).

I ran a regression-everything after this change, with no failures found.


MattKaufmann:
Eric tells me he did a successful regression-everything with this change, so I see no problem with merging.

PR #1663, opened 2024-12-18
Fix read-bytes$ guard for 8-byte reads

yaso9:
While the definition of read-bytes$ supports 8 byte reads, the guard
disallowed them. This fixes the guard to allow 8 byte reads.


yaso9:
I'm not sure who's the correct person to add as a reviewer for std books.

ericwhitmansmith (commented):
This looks good to me.

acoglio (approved):

ericwhitmansmith:
I'm going to merge this.  Jared at one point gave me his blessing to maintain his files, and this seems like a clear improvement.

PR #1664, opened 2024-12-19
x86isa -- add asmtest/snippet testing framework, implement adcx, adox, blsi instructions

solswords:
The newly implemented instructions pass some snippet-based tests, but I could use some more experienced eyes on the operand decoding.


books/projects/x86isa/machine/instructions/arith-and-logic.lisp
```diff
@@ -295,6 +295,107 @@
 
     x86))
 
+
+
+(def-inst x86-adcx/adox
+
+  :parents (three-byte-opcodes)
+
+  :short "Operand Fetch and Execute for ADD, ADC, SUB, SBB, OR, AND,
```
yaso9, line 304: This documentation seems to be copied from `x86-add/adc/sub/sbb/or/and/xor/cmp-G-E`. Updating the documentation is sufficient, but I'd suggest adding `gpr-adcx/adox-spec` to `gpr-arith/logic-spec` and then adding support for those instructions to `x86-add/adc/sub/sbb/or/and/xor/cmp-G-E`.

books/projects/x86isa/machine/instructions/bit.lisp
```diff
@@ -822,3 +822,85 @@
...
+       ((the (unsigned-byte 32) input-rflags) (rflags x86))
+       (output-rflags (!rflagsBits->cf
+                       (b-not (bool->bit (eql src2-val 0)))
+                       (!rflagsBits->sf
+                        (logbit (1- (* 8 operand-size)) res)
+                        (!rflagsBits->zf
+                         (bool->bit (eql res 0))
+                         (!rflagsBits->of 0 input-rflags)))))
+       ;; SDM says AF and PF are undefined. Are they really undefined or just preserved?
+       (undefined-flags (!rflagsbits->pf 1 (!rflagsbits->af 1 0)))
```
yaso9, line 900: Flags that are undefined according to the SDM should be handled by setting them with `!flgi-undefined`. See `x86-tzcnt`.

books/projects/x86isa/machine/instructions/bit.lisp
```diff
@@ -822,3 +822,85 @@
...
+
+  :returns (x86 x86p :hyp (x86p x86))
+  :guard (vex-prefixes-byte0-p vex-prefixes)
+
+  :modr/m t
+  :vex t
+  :guard-debug t
+  :body
+  ;; placeholder
+  (b* ((operand-size (if (eql (vex->w vex-prefixes) 1) 8 4))
```
yaso9, line 858: The SDM says VEX.W1 is ignored in non-64 bit modes. Should that be checked here? I'm not very familiar with how 32-bit support works in the model.

yaso9 (changes requested):
I left some comments. I don't know how VEX/EVEX stuff works, so I didn't look too closely at how `blsi` is decoding the operands.

books/projects/x86isa/machine/instructions/arith-and-logic.lisp
solswords: Thanks, dumb copy-paste -- will fix the doc.  I actually tried doing what you suggested before defining a separate function for adox/adcx, but there were a few things that made me think adox/adcx didn't sufficiently fit the patterns assumed by that function. In particular, the operand size criteria for adox/adcx seems incompatible to me -- the x86-add/etc. function assumes the operand size is 1 if the low bit of the opcode is 0 (as in the code below), and it respects the 0x66 operand size override prefix which is a mandatory prefix for ADCX.
```
   (b* (...
       (byte-operand? (eql 0 (the (unsigned-byte 1)
                               (logand 1 opcode))))
       ((the (integer 1 8) operand-size)
        (select-operand-size
         proc-mode byte-operand? rex-byte nil prefixes nil nil nil x86))
```

books/projects/x86isa/machine/instructions/bit.lisp
solswords: I believe this is handled by write-user-rflags, called below -- at least, this idiom seems to be used in a few places in arith-and-logic.lisp and rotate-and-shift.lisp.

books/projects/x86isa/machine/instructions/bit.lisp
solswords: Thanks, good catch.

books/projects/x86isa/machine/instructions/arith-and-logic.lisp
yaso9: I see, that makes sense then.

books/projects/x86isa/machine/instructions/bit.lisp
yaso9: Oh I see, I completely misread that and thought you were setting rflags. Using `write-user-rflags` works.

yaso9 (approved):

books/projects/x86isa/machine/instructions/arith-and-logic.lisp
```diff
@@ -295,6 +295,102 @@
...
+
+  :modr/m t
+
+  :additional-args ((adcx booleanp))
+  
+  :body
+
+  (b* ((p2 (prefixes->seg prefixes))
+       (p4? (eql #.*addr-size-override* (prefixes->adr prefixes)))
+       (operand-size (if (logbitp *w* rex-byte) 8 4))
```
yaso9, line 334: Sorry, I just realized, this also doesn't account for 32-bit mode, where this is always 32-bit. Does `select-operand-size` not work here because the operand size shouldn't be less than 32-bits?

yaso9 (changes requested):

books/projects/x86isa/machine/instructions/arith-and-logic.lisp
solswords: I looked into this and I believe rex-byte will always be 0 if not in 64-bit mode -- get-prefixes is responsible for detecting the rex byte and it only does so if 64-bit mode. I think select-operand-size doesn't work unless we fake some stuff -- as you said it won't be less than 32 bits, plus the operand size for these instructions is dictated only by rex.w, not the default operand size and not the 0x66 prefix (which is a mandatory prefix for ADCX).

yaso9: I see, cool. Looks good then.

yaso9 (approved):

solswords:
Sorry if the force-push causes any trouble. I realized I had accidentally added the asmtest executable somehow. Everything else should be the same.

books/projects/x86isa/machine/dispatch-macros.lisp
```diff
@@ -318,9 +320,11 @@
                       (eql proc-mode #.*compatibility-mode*)
                       (eql proc-mode #.*protected-mode*)))))
     :ud)
-   ((eq type-id :type-vex-gpr)
+   ((or (eq type-id :type-vex-gpr)
+        ;; BOZO Sol -- it seems like maybe this is a synonym?
+        (eq type-id :type-13))
     ;; from table 2.5.1 from volume 2.. we only need to additionally check
-    ;; the cpuid requirements:
+    ;; the cpuid requirements: (BOZO Sol -- maybe this is now table 2.6.1?)
```
acoglio, line 327: Yes, in the June 2024 version:
- All the 2.4.x sections in Volume 2 are now 2.5.x.
- All the 22-x tables in Volume 3 are now 23-x.

books/projects/x86isa/machine/instructions/add-spec.lisp
```diff
@@ -389,3 +389,98 @@
 (make-event (gpr-adc-spec-gen-fn 8))
 
 ;; ======================================================================
+;; SPECIFICATION: ADCX/ADOX
+;; ======================================================================
+
+;; ADCX: Similar to ADC but sets only CF, preserves other flags
```
acoglio, line 395:
```suggestion
;; ADCX: Similar to ADC but sets only CF, preserves other flags.
```

books/projects/x86isa/machine/dispatch-macros.lisp
```diff
@@ -318,9 +320,11 @@
                       (eql proc-mode #.*compatibility-mode*)
                       (eql proc-mode #.*protected-mode*)))))
     :ud)
-   ((eq type-id :type-vex-gpr)
+   ((or (eq type-id :type-vex-gpr)
+        ;; BOZO Sol -- it seems like maybe this is a synonym?
```
acoglio, line 324: I'm not familiar with this part of the code, so I don't think I can provide an informed opinion. I can't even find the pre-existing `:type-vex-gpr` anywhere in the code other than here (perhaps it occurs in generated code?).

acoglio (approved):

ericwhitmansmith:
I can confirm that these changes don't break any of our proofs.

ericwhitmansmith (approved):

PR #1666, opened 2024-12-21
Make some stobjs non-executable

yaso9:
- **Make the concrete stobj generated by defrstobj2 non-executable**
  The concrete stobj generated by defrstobj2 is abstracted away from the
  user. Thus, it is never directly used. Making it non-executable saves
  memory.
  

- **Make bigmem::mem and bigmem::mem$c non-executable**
  While bigmem::mem itself doesn't use much memory, if one attaches the
  bigmem-asymmetric::mem to it, the memory usage can be substantial.
  bigmem::mem is generally not used directly, so making it non-executable
  to save on memory makes sense and if needed one can use
  `add-global-stobj` to use it at the top level.
  
  bigmem::mem$c is only used to back bigmem::mem so making it
  non-executable as well shouldn't be an issue and saves some (although
  probably insignificant) memory.

### DIFF
--- a/books/centaur/bigmems/bigmem/bigmem.lisp
+++ b/books/centaur/bigmems/bigmem/bigmem.lisp
@@ -197,6 +197,7 @@
 
 (acl2::defabsstobj-events mem
 
+    :non-executable t
     :attachable t
     :foundation mem$c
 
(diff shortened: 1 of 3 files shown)


yaso9:
I ran a regression-everything to make sure everything still builds with these changes.

solswords (approved):
Seems fine, thanks!

PR #1669, opened 2024-12-23
x86isa: implement pcmpgt and some instruction bugfixes

yaso9:
- **Fix some x86isa shifts**
  The flag computation for SAR treated the CF as undefined when the shift
  count was greater than the operand size, but that is incorrect.
  
  Additionally, the count mask was computed incorrectly for byte operands
  with a REX prefix
  

- **Fix x86isa cmpxchg computing incorrect flags**
  

- **x86isa Fix large SSE2 shift counts causing Lisp errors**
  The SSE2 packed shift instructions which take the count as an XMM
  register could have a huge (128-bit wide) shift count. These cause CCL
  to throw when passed to `ash`.
  
  Clamping the shift count to the el-width resolves this issue. This
  commit also cleans up some of the SSE2 shift implementation code.
  

- **x86isa Implement pcmpgt(b/w/d) instructions**

### DIFF
--- a/books/projects/x86isa/machine/instructions/pcmp.lisp
+++ b/books/projects/x86isa/machine/instructions/pcmp.lisp
@@ -147,3 +147,108 @@
        ;; Update the instruction pointer.
        (x86 (write-*ip proc-mode temp-rip x86)))
       x86))
+
+(define pcmpgt ((result-width natp)
+                (el-width posp)
+                (a natp)
+                (b natp))
+  :prepwork ((local (include-book "arithmetic-5/top" :dir :system)))
+  :guard (equal (mod result-width el-width) 0)
+  :returns (result (unsigned-byte-p result-width result)
+                   :hyp (and (natp result-width)
+                             (<= (pos-fix el-width) result-width)
+                             (equal (mod result-width (pos-fix el-width)) 0))
+                   :hints (("Goal" :in-theory (disable unsigned-byte-p
+                                                       bitops::logapp-of-i-0
+                                                       acl2::prefer-positive-addends-<))))
+  :measure (nfix result-width)
+  (b* ((result-width (nfix result-width))
+       (el-width (pos-fix el-width))
+       (a (nfix a))
+       (b (nfix b))
+       ((when (zp result-width)) 0)
+       (el-greater? (> (logext el-width a)
+                     (logext el-width b))))
+      (logapp el-width (if el-greater? -1 0)
+              (pcmpgt (- result-width el-width) el-width
+                      (logtail el-width a) (logtail el-width b)))))
+
+(def-inst x86-pcmpgt-sse
+  :parents (two-byte-opcodes)
+  :long
+  "<code>
+  PCMPGTB xmm1, xmm2/m128
+  PCMPGTW xmm1, xmm2/m128
+  PCMPGTD xmm1, xmm2/m128
+  </code>"
+
+  :modr/m t
+
+  :returns (x86 x86p :hyp (x86p x86))
+  :guard-hints (("Goal" :in-theory (disable unsigned-byte-p)))
+
+  :body
+
+  (b* ((p2 (prefixes->seg prefixes))
+       (p4? (eql #.*addr-size-override* (prefixes->adr prefixes)))
+       (seg-reg (select-segment-register proc-mode p2 p4? mod r/m sib x86))
+
+       ;; The operand size is always 128 bits, i.e. 16 bytes.
+       (operand-size 16)
+
+       ;; The first source operand (Operand 1 in the Intel manual)
+       ;; is the XMM register specified in Reg.
+       ;; This is also the destination operand,
+       ;; and thus we obtain the index for later use.
+       ((the (unsigned-byte 4) src1/dst-index)
+        (reg-index reg rex-byte #.*r*))
+       ((the (unsigned-byte 128) src1)
+        (xmmi-size operand-size src1/dst-index x86))
+
+       ;; The second source operand (Operand 2 in the Intel manual)
+       ;; is the XMM register, or memory operand, specified in Mod and R/M.
+       (inst-ac? t) ; Intel Manual Volume 2 Table 2-21 (Dec 2023)
+       ((mv flg
+            (the (unsigned-byte 128) src2)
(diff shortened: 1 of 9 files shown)


yaso9:
The bugs were discovered and fixes tested using a **very** crude test generator for asmtest, which I hope to get cleaned up and pushed out sometime in the near future. Please pay particular attention to the fixes to the scalar shift instruction flag computations. I think I have them right after staring at the SDM for a long time. Testing is difficult for the flags in the shift instruction because which flags are undefined is dependent on the input.

solswords (approved):
I'm on vacation this week so I'm not going to be able to give this the thorough look I'd like to til after I get back. Approving for now and will revisit in more detail in a week or so.

yaso9:
Sorry to push even more stuff while people are on vacation. These are just some more bugfixes.

I don't need these merged soon, so if you're on vacation, don't worry about looking at these until you're back.

yaso9:
It seems the change to SAR breaks `SAR-SPEC-8-ALT-DEF` in `books/kestrel/x86/x86-changes.lisp`. I think the flags are computed wrong there in the same way the old spec in the model computed it wrong.

ericwhitmansmith:
Probably that stuff was just copied from the old code.  I'll check out your branch and see what's going on.

ericwhitmansmith:
Ok, I made a PR that fixes the issues in x86-changes.lisp:
https://github.com/yaso9/acl2/pull/1

I can also confirm that this PR (with my fixes to x86-changes.lisp) doesn't break any of our x86 proofs.  My head's not into the instruction models yet, so I am going to approve this PR without reviewing the details there.

ericwhitmansmith (approved):

books/projects/x86isa/machine/instructions/shifts-spec.lisp
```diff
@@ -149,14 +149,14 @@
                       (mbe
                        :logic (part-select
                                dst
-                               :low ,size-1 ;; (- ,size src)
+                               :low (- ,size src)
```
acoglio, line 152: Is this necessary? We are under the case `src = 1` (line 144), so `(- ,size src)` should be the same as `,size-1`. Perhaps the comment in the original line should be `;; = (- ,size src)`?

books/projects/x86isa/machine/instructions/shifts-spec.lisp
```diff
@@ -149,14 +149,14 @@
...
                                :width 1)
                        :exec
                        (the (unsigned-byte 1)
                             (logand 1
                                     (the (unsigned-byte ,size)
                                          (ash (the (unsigned-byte ,size) dst)
-                                              ,neg-size-1))))))
+                                              (- src ,size)))))))
```
acoglio, line 159: Same comment as line 152 above applied here.

books/projects/x86isa/machine/instructions/shifts-spec.lisp
```diff
@@ -692,12 +682,17 @@ set to the most-significant bit of the original operand.</p>"
...
-                                              (unsigned-byte-p
-                                               (tau-system)))))
+       :guard-hints
+       (("Goal" :in-theory (e/d* (rflag-RoWs-enables)
+                                 (unsigned-byte-p
+                                  bitops::logbit-to-logbitp
+                                  (tau-system))))
+        (and stable-under-simplificationp
+             '(:in-theory (enable bitops::logbit-to-logbitp))))
+       :guard-debug t
```
acoglio, line 692: I think that `:guard-debug t` makes things a little slower, so it would be good to remove.

books/projects/x86isa/machine/instructions/shifts-spec.lisp
```diff
@@ -790,96 +795,62 @@ set to the most-significant bit of the original operand.</p>"
...
-                       (pf (general-pf-spec ,size result))
-                       ;; AF is undefined.
-                       (zf (zf-spec result))
-                       (sf (general-sf-spec ,size result))
-                       ;; OF is undefined.
-
-                       (output-rflags (mbe :logic
-                                           (change-rflagsBits
+                 (b* ((cf (if (<= ,size src)
+                            (mbe :logic (part-select dst :low ,(1- size) :width 1)
```
acoglio, line 799: Could use `size-1` instead of `(1- size)`.

books/projects/x86isa/machine/instructions/pcmp.lisp
```diff
@@ -147,3 +147,108 @@
...
+                   :hints (("Goal" :in-theory (disable unsigned-byte-p
+                                                       bitops::logapp-of-i-0
+                                                       acl2::prefer-positive-addends-<))))
+  :measure (nfix result-width)
+  (b* ((result-width (nfix result-width))
+       (el-width (pos-fix el-width))
+       (a (nfix a))
+       (b (nfix b))
+       ((when (zp result-width)) 0)
+       (el-equal? (> (logext el-width a)
```
acoglio, line 170: `el-greater?` may be a better name for this variable.

books/projects/x86isa/machine/instructions/shifts-spec.lisp
yaso9: Good catch, left that in on accident. There are a handful of them throughout the code, which we should probably remove.

books/projects/x86isa/machine/instructions/shifts-spec.lisp
yaso9: Yeah looks like it doesn't do anything. I must have not realized that didn't change anything.

books/projects/x86isa/machine/instructions/pcmp.lisp
yaso9: Whoops, copy pasta error

ericwhitmansmith:
In testing these latest changes I am seeing a guard proof failure in SAR-SPEC-8.

yaso9:
I've seen that failure. I started working on resolving it, but got busy with something else. I'll likely get to it sometime later today (I'm in UTC+5 at the moment, so probably going to be tomorrow morning for you).

acoglio (approved):

yaso9:
This should fix the guard issue

PR #1670, opened 2024-12-27
Paco package

ericwhitmansmith:
This splits out the PACO package, so I can bring it into the big-data stuff without getting anything else.  That lets me avoid making a copy of it.


MattKaufmann:
Eric asked J if he's OK with this change, then J said yes, so now I'm accepting this PR.

PR #1671, opened 2024-12-27
Have makefile print OS_HAS_CPP.

ericwhitmansmith:
(The order of the printing matches the order of the checks, or we could alphabetize both.)


MattKaufmann (approved):
I don't know anything about OS_HAS_CPP, but if this works then I can't see any reason not to proceed with this change.

gjurgensen (approved):
LGTM. I suppose I should have done this when adding the Makefile variable?

ericwhitmansmith:
Yeah, maybe.  Not sure how I noticed it was missing.  Okay, I will merge.

PR #1681, opened 2025-01-01
[X86ISA] Reduce inclusion.

ericwhitmansmith:
The book about initializing the x86 state doesn't need all the instruction semantic functions.  As is common with this sort of change, include-books had to be added to some downstream books, but overall the dependencies were reduced.

(This change was enabled by an earlier change that collected rules about XR.)


yaso9 (approved):
Was this meant to merge into `testing` instead of `master`?

acoglio (approved):

ericwhitmansmith:
Yes, and I just changed it to testing (thanks!).  BTW, I tried to find a setting to make all PRs default to using testing as the base, but I couldn't find one.

PR #1686, opened 2025-01-03
[X86ISA] Remove subsumed rule.

ericwhitmansmith:
This is subsumed by x86p-xw, which is much more general.  That rule is:

(defthm x86p-xw
  (implies (x86p x86)
           (x86p (xw fld index val x86))))

My only hesitation in removing x86p-!rip-when-val-is-canonical-address-p is that it has a force on the x86p hyp (so x86p-xw may not be strictly better), but the forcing doesn't seem to be needed anywhere.

### DIFF
--- a/books/kestrel/x86/support-x86.lisp
+++ b/books/kestrel/x86/support-x86.lisp
@@ -61,8 +61,8 @@
   (implies (x86p x86)
            (x86p (xw x86isa::fld x86isa::index value x86))))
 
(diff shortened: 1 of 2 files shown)


acoglio (approved):
If everything works without the forcing, it seems fine to remove it.

solswords (approved):

yaso9 (approved):

PR #1695, opened 2025-01-05
[X86ISA] Speed up bitstruct proofs.

ericwhitmansmith:
Each call to defbitstruct locally includes some helper books.  To avoid doing those includes over and over, we do them (locally) at the start of each book that calls defbitstruct repeatedly.

This saves about 14 seconds of the 60 seconds needed for this directory, on a very fast machine.

### DIFF
--- a/books/projects/x86isa/utils/basic-structs.lisp
+++ b/books/projects/x86isa/utils/basic-structs.lisp
@@ -40,6 +40,11 @@
 
 (include-book "centaur/fty/bitstruct" :dir :system)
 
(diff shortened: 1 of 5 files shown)


yaso9 (approved):
I don't really like that this breaks the abstraction boundary of defbitstruct. That being said, I can't think of a better solution.

acoglio (approved):

ericwhitmansmith:
It's good that the defbitstruct forms work either way, so this is just an optimization.  Alessandro suggests that defbitstruct could be improved to provide a macro that does the includes.  Maybe I'll do that in the future.

PR #1697, opened 2025-01-06
x86isa: Introduce `testgen` a test generator for asmtest

yaso9:
`testgen` uses XED to iterate over x86 instructions and generate tests
for them. It has some restrictions on instructions it supports. The big
ones are it only works with instructions that operate on flags, gprs,
xmm registers, and immediates. Beyond introducing `testgen`, this also
introduces some scripts to make it easy to run all snippets with random
inputs.

At the moment there are a few issues:
  - XED errors when trying to encode some instructions. I don't know why
    exactly that is.
  - `asmtest` crashes with various errors when running some tests.
  - Some instructions fail tests because some flags are undefined. We
    handle undefined flags correctly but flags which are undefined
    only sometimes can't easily be dealt with in an automated fashion.

That being said, these issues are handled gracefully and I'm comfortable
having this merged into the community books.

### DIFF
--- a/books/projects/x86isa/tools/execution/asmtest/.gitignore
+++ b/books/projects/x86isa/tools/execution/asmtest/.gitignore
@@ -1,0 +1,12 @@
+check-test
+check-test.*
+!check-test.lsp
+inputs
+results
+outputs
+snippets/testgen.nasm
+snippets-generated.txt
+snippets-combined.txt
+asmtest
+snippets.lsp
+snippets.h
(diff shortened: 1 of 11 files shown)


yaso9:
Btw there appear to be some fp instructions failing tests right now. I haven't looked closely at that yet.

acoglio (approved):

solswords (approved):
This sounds really cool! A future possibility is to build multiple asmtest executables with different sets of snippets -- that way one with only the handcrafted tests wouldn't need to depend on xed sources/an internet connection to build it. But that's a minor quibble.

ericwhitmansmith:
My head's not really in this stuff yet, but I also think that splitting out the handwritten tests would be good.

ericwhitmansmith (approved):

PR #1710, opened 2025-01-17
x86isa: Misc Linux Related Changes

yaso9:
- **x86isa Update Linux loader memory layout**
  This updated layout provides more space for larger initrd images
  

- **x86isa Allow setting :reuse-address on tty-raw TCP socket via env var**
  Normally one can't open a TCP socket on a port that was recently closed.
  Setting :reuse-address allows one to open the socket despite recent use,
  which is useful in debugging, but is not standards compliant.
  
  Setting the TTY_DEBUG environment variable allows one to set the
  :reuse-address option for use in debugging.
  

- **x86isa linux-load read disk-image without intermediate byte-list**
  Previously, linux-load would first read a disk image into a byte-list
  and then write it to memory. For large disk images, this could take up a
  lot of memory. Now, we write the disk image directly into the model's
  memory.
  

- **Update x86isa Linux docs and patch**


yaso9:
Merging since Warren would like to get it from `master`, and this is Linux related stuff, so I don't think anyone else will care. I can back stuff out if necessary.

PR #1717, opened 2025-01-25
[X86ISA] Disable some rules generated by defbitstruct.

ericwhitmansmith:
These backchain from unsigned-byte-p to the specific bitstruct predicates.  So they can introduce reasoning about the bitstructs out of nowhere in unrelated contexts.  Thus, I think we don't want them enabled by default.  I re-enabled them only where needed.

Example of problematic rule:

(DEFTHM UNSIGNED-BYTE-P-WHEN-MXCSRBITS-P
  (IMPLIES (MXCSRBITS-P X)
           (UNSIGNED-BYTE-P 32 X)))

If people approve this change, I can make similar changes for other bitstructs.  Even better would be to have defbitstruct (optionally) disable these rules.

### DIFF
--- a/books/kestrel/x86/floats.lisp
+++ b/books/kestrel/x86/floats.lisp
@@ -522,10 +522,12 @@
   (integerp (!MXCSRBITS->IE$INLINE bit mxcsr)))
 
 (defthm unsigned-byte-p-32-of-!MXCSRBITS->IE
-  (unsigned-byte-p 32 (!MXCSRBITS->IE$INLINE bit mxcsr)))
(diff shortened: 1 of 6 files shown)


acoglio (approved):

yaso9 (approved):
This is good, but maybe a nicer solution is some syntaxp hyps for the offending bitstruct rules that keep them from triggering in unrelated contexts.

solswords (approved):
Fine by me. I thought at first you were disabling all such rules by default when issued by defbitstruct. I would agree with that in principle, but it would be almost certain to break some things not in the public books.

PR #1718, opened 2025-01-28
Add XDOC for `fty::defoption`

gjurgensen:
The existing XDOC page for `defoption` is only a placeholder with a "TODO" message.

The PR adds a basic description of `defoption` and demonstrates its syntax and keyward options.


acoglio (approved):

solswords (approved):
Looks good. One additional thing you could consider documenting is that this won't work if the base type includes NIL.

PR #1722, opened 2025-01-30
[x86isa] Improve guard of rml08.

ericwhitmansmith:
Previously its guard included both (canonical-address-p lin-addr) and (signed-byte 48 lin-addr), which mean the same thing.  I added :split-types so that the signed-byte-p claim is used only as a :type, not in the guard.  I then had to add (member-eq r-x '(:r :x)) to the guard so that it implies the :type.

If people approve this PR, I can make similar changes to other functions.  A goal is to be able to keep canonical-address-p disabled more (e.g., to speed up some slow proofs involving linear-memory).

### DIFF
--- a/books/projects/x86isa/machine/linear-memory.lisp
+++ b/books/projects/x86isa/machine/linear-memory.lisp
@@ -1337,7 +1337,9 @@
                (x86))
 
   :parents (linear-memory)
-  :guard (canonical-address-p lin-addr)
+  :guard (and (canonical-address-p lin-addr)
+              (member-eq r-x '(:r :x)))
+  :split-types t
   :guard-hints (("goal"
                  :in-theory (e/d* (ifix rvm08) ())))
 


acoglio (approved):

yaso9 (approved):
It would be nice to modify `define` to allow adding a value to a table so that every argument marked as a `something-p` can automatically be given a corresponding type.

There is some precedent for this. We have post define hooks that alter the behavior of `define` based on a table. Support for pre-define hooks, which can modify the submitted `defun`, would be a more general mechanism upon which the above can be implemented.

That would make code like this far cleaner. The way it is now is pretty ugly.

ericwhitmansmith:
That is an interesting idea.  I almost want defun to do it (forward-chain from the guard and harvest any type information from the result as declares to help the compiler).  But I guess that would be a big change.

In this case, I almost deleted the :type on r-x, because I don't think it will help the compiler much (on my ARM machine its presence doesn't seem to affect the machine code).  However, I feel the same way about the :type on lin-addr (since lin-addr is just passed to subfunctions, not operated on as a number) and I found that deleting its :type did change the machine code.  So I decided to be conservative and change as little as possible.

MattKaufmann:
Regarding:

> ... I almost want defun to do it (forward-chain
> from the guard and harvest any type information
> from the result as declares to help the
> compiler)

Since defun is built into Common Lisp, generating declares for the compiler isn't really feasible.  (I suppose it is actually, by writing them into the expansion file, but that's pretty extreme.)  However, a related idea might be feasible: generate (declaim (ftype ..)) forms from the guards.  Still, the type and :split-types mechanisms have been designed to support such assistance for the compiler, and they seem sufficient to me rather than building in such support via guards.

PR #1728, opened 2025-02-03
[RAC] Minor bug-fixes

voskh0d:
1. Compare operations between integer types are cast to the right type.
2. Fix diagnostics issue
3. Multiple value dereference type of elements during typing phase
4. Update test and add missing test files (null-statement)


voskh0d:
@mayankmanj I've squashed the commits, added the missing files (null-statements.cpp and null-statements.cpp.ref.stderr) a fix a failing test (empty was missing a line)

PR #1735, opened 2025-02-13
[X86ISA] Comment out some unneeded hyps.

ericwhitmansmith:
Perhaps the rules should also be renamed, since their names mention 'app-view' but those hyps are not needed.


acoglio (approved):
I agree they should be renamed.

acoglio (approved):
Agree on renaming.

PR #1736, opened 2025-02-19
FTY: add tables for default arguments, option for short deftagsum product names

solswords:
For each type-defining macro compatible with deftypes, this adds a table `fty::<macroname>-defaults` whose contents are used as default values for keyword arguments for that macro. E.g.
`(local (table fty::defprod-defaults :layout :list))`
would change the default product layout to list instead of alist, within your local scope.

Additionally, adds an option `:short-names` to deftagsum and defflexsum which changes the default base name for product constructors/accessors from sumname-productname to just productname (in the package of sumname).


acoglio (approved):

solswords:
I forgot this PR hadn't been merged yet and included it in a subsequent push, therefore I'm closing it. Lmk if it causes any problems but I wouldn't think so.

PR #1737, opened 2025-03-03
[Std/util] Improve `define-sk`.

acoglio:
Have guard proof failures show on the screen.

### DIFF
--- a/books/std/util/define-sk.lisp
+++ b/books/std/util/define-sk.lisp
@@ -993,9 +993,10 @@ so execution differences don't matter.</p>")
                   '(value-triple :invisible)))))
 
       ,@(and verify-guards
-             `((verify-guards ,guts.name-fn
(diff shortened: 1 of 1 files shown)


ericwhitmansmith:
LGTM.  This is analogous to this change I made for define:

https://github.com/acl2/acl2/commit/c353f1c37575d8d898b34a8b4d7777c4fa7d8222#diff-f7af72aeab99c11af8ce93e49599927e1bc2de37f03385d4e34576687856249a

Are there any other define variants that would benefit from this as well?

ericwhitmansmith (approved):

solswords (approved):

acoglio:
@ericwhitmansmith Yes, I remembered and compared your PR. I think `define-sk` is the only `define` variant.

gjurgensen:
There is also `defines`, but it looks like it is already popping the `with-output` stack for the `verify-guards` call.

PR #1739, opened 2025-03-19
[FTY] Make the auto-bound kind variable ignorable.

acoglio:
The `T-case` macro associated to a sum fixtype `T` generates a binding of `x.kind`, where `x` is the argument of the `T-case`. If the sum type consists of one summand, the `x.kind` ends up being unused, which causes an error in normal conditions. This commit makes that variable `x.kind` ignorable, thus avoiding the error for the case of a one-summand sum type.

This commit also auto-removes some trailing whitespace.


gjurgensen (approved):
LGTM.

solswords (approved):

PR #1741, opened 2025-04-04
[C$] Add `typedef` definitions to validation

gjurgensen:
This adds definitions to the `typedef` entries in the validation table. Redefined typedefs are checked for the conditions outlined in the standard, insofar as we are able to represent the conditions in our current approximation of types.

This also adds an approximate compatible type recognizer and null pointer constant recognizer.

It fixes validation for the case where an expression of unknown type is used to initialize something of target type `struct`.

It also adds a number of test cases.


books/kestrel/c/syntax/validation-information.lisp
```diff
@@ -700,6 +700,101 @@
...
+  (b* ((x (type-fix x))
+       (y (type-fix y)))
+    (or (equal x y)
+        (type-case x :unknown)
+        (type-case y :unknown)
+        (and (type-integerp x) (type-case y :enum))
+        (and (type-case x :enum) (type-integerp y))))
+  :hooks (:fix))
+
+;;;;;;;;;;;;;;;;;;;;
```
acoglio, line 754: I would use `///` and put the reflexivity and symmetry theorems under the define, so they should up in the XDOC.

books/kestrel/c/syntax/validation-information.lisp
```diff
@@ -700,6 +700,101 @@
...
+   (xdoc::p
+    "Because we currently only model an approximation of C types, our notion of
+     compatibility is also approximate.
+     Specifically, this relation overapproximates true type compatibility.
+     Compatible types should always be recognized as such,
+     but incompatible types may also be recognized.")
+   (xdoc::p
+    "In particular:"
+    (xdoc::ul
+     (xdoc::li "all composite types of a particular kind are currently
```
acoglio, line 722: Why "composite"?

books/kestrel/c/syntax/validator.lisp
```diff
@@ -2675,13 +2676,23 @@
...
+                     ((mv info? -)
+                      (valid-lookup-ord tyspec.name table))
+                     ((unless info?)
+                      (reterr (msg "The identifier ~x0 is not an in-scope ~
+                                    ordinary identifer."
+                                   (ident->unwrap tyspec.name)))))
+                  (valid-ord-info-case
+                    info?
+                    :typedef (retok (type-spec-typedef tyspec.name)
+                                    (valid-ord-info-typedef->def info?)
```
acoglio, line 2690: I think `info?.def` should suffice?

books/kestrel/c/syntax/validation-information.lisp
```diff
@@ -700,6 +700,101 @@
...
+          pointer constant [C17:6.3.2.3/3]."
+  :long
+  (xdoc::topstring
+   (xdoc::p
+    "Due to the approximate representation of types and our lack of constant
+     expression evaluation,
+     this recognizer is highly overappoximating.
+     It will recognize any pointer, integer, or unknown type."))
+  (or (type-case type :pointer)
+      (type-case type :unknown)
```
acoglio, line 780: I wonder if we need to include the unknown type here. It makes sense that the unknown type could be in fact pointer or integer, but I think that, everywhere we use `null-pointer-exprp`, we have already handled the cases of unknown type (see, for instance), the beginning code of `valid-binary`. But it doesn't hurt to include the unknown type in this predicate, maybe it's safer to do so. Just bringing this up for thought.

books/kestrel/c/syntax/validation-information.lisp
```diff
@@ -700,6 +700,101 @@
...
+  :enable type-compatiblep)
+
+(defrule type-compatiblep-symmetric
+  (equal (type-compatiblep y x)
+         (type-compatiblep x y))
+  :enable type-compatiblep)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define null-pointer-exprp ((expr exprp) (type typep))
```
acoglio, line 767: Given that the argument must be an `exprp`, I tend to use a pattern like `expr-<something>p` rather than `<something>-exprp` in cases like this. Here, I would use `expr-null-pointer-const-p`, i.e. "is the expression a null pointer constant?" vs. "is this thing a null pointer constant expression?". Also see comment for next predicate.

books/kestrel/c/syntax/validation-information.lisp
```diff
@@ -700,6 +700,101 @@
...
+     this recognizer is highly overappoximating.
+     It will recognize any pointer, integer, or unknown type."))
+  (or (type-case type :pointer)
+      (type-case type :unknown)
+      (type-integerp type))
+  :hooks (:fix))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define null-pointer-const-exprp ((const-expr const-exprp) (type typep))
```
acoglio, line 786: Related to the comment for the previous predicate, I would use `const-expr-null-pointer-const-p` here. Note that the current name could be interpreted as testing an expression (not necessarily a constant one) for being a "null pointer constant".

acoglio (changes requested):
Looks good, left a few comments.

books/kestrel/c/syntax/validation-information.lisp
gjurgensen: Perhaps the wrong word. I meant to refer to the non-atomic types: aggregate types, pointers, unions, function types. Compatibility of these types is approximate because the representation of these types is approximate.

I'll clarify that language.

acoglio: Understood. There must an official term (perhaps consisting of a few categories of types) to describe those types you were referring to (some predicates for that are defined in the validation-information.lisp file).

In the standard, 'composite type' has a specific meaning [C17:6.2.7].

gjurgensen:
Yes, its ready now.

books/kestrel/c/syntax/validation-information.lisp
gjurgensen: I see, yes, definitely the wrong word then. I update the wording to specifically mention the types which are approximated.

PR #1743, opened 2025-04-05
[FTY] Make `defset` and `defomap` more robust

gjurgensen:
Adds explicit induction hints such that `defset` and `defomap` will work with `(set-induction-depth-limit 0)`.

Also removes redundant fix-when-recognizer theorems (already covered by the returns specifiers).


acoglio (approved):
This is great!

PR #1744, opened 2025-04-05
[XDOC] Don't suppress error messages in `defsection`

gjurgensen:
This avoids the issue of `defsection` and forms which generates `defsection`s (like `define`) from failing silently on bad `:short` and `:long` forms.

For instance errors in the application of xdoc constructors in a `define` currently cause a silent failure. After this change, the failure is printed like normal to the comment window.


acoglio (approved):
Yes, very useful!

PR #1750, opened 2025-04-15
[C$] Add optional tag to struct types

gjurgensen:
The interpretation of `(type-struct (ident "foo"))` is the collective type of all struct types with tag `"foo"`. To be perfectly precise, one would need to distinguish structs types further based on the scope of their declaration (and of course information about the members).

Similarly, `(type-struct nil)` is the collective type of all tagless structs. To be more precise in this case, we'd need to also represent the location of its declaration, as each declaration of a tagless structure type denotes a new type [C17:6.7.2.3/5].

The validator has been updated to recognize different structure types and check for compatibility as appropriate. Specifically, this involved changes to validation of simple assignments, initializers, and type specifiers.

### DIFF
--- a/books/kestrel/c/syntax/validator.lisp
+++ b/books/kestrel/c/syntax/validator.lisp
@@ -1292,7 +1292,8 @@
       pointer type."))
    (xdoc::p
     "We do not perform array-to-pointer or function-to-pointer conversion
-     on the left operand, because the result would not be an lvalue.")
+     on the left operand, because the result would not be an lvalue.
+     The type of the result is the type of the left operand [C17:6.5.16/3].")
    (xdoc::p
     "The @('*=') and @('/=') operators require arithmetic operands
      [C17:6.5.16.2/2],
--- a/books/kestrel/c/syntax/validator.lisp
+++ b/books/kestrel/c/syntax/validator.lisp
@@ -1523,7 +1523,7 @@
      In our currently approximate type system,
      the other two operands must have
      both arithmetic type,
-     or both compatible structure types,
+     or both the same structure type,
      or both the union type,
      or both the void type,
      or both the pointer type,
@@ -1555,7 +1555,7 @@
         (retok (type-uaconvert type2 type3 ienv)))
        ((when (and (type-case type2 :struct)
                    (type-case type3 :struct)))
-        (if (type-compatiblep type2 type3)
+        (if (type-equiv type2 type3)
             (retok type2)
           (reterr (msg "Struct types ~x0 and ~x1 are incompatible."
                        type2
--- a/books/kestrel/c/syntax/validator.lisp
+++ b/books/kestrel/c/syntax/validator.lisp
@@ -1276,29 +1276,23 @@
      an lvalue as left operand [C17:6.5.16/2],
      but currently we do not check that.
      In our currently approximate type system,
-     the requirements in [C17:6.5.16.1/1] reduce to the two operands having
-     both arithmetic types,
-     or both the structure type,
-     or both the union type,
-     or both pointer types,
-     or one pointer type and one integer type.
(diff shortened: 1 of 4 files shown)


books/kestrel/c/syntax/validator.lisp
```diff
@@ -1276,29 +1276,23 @@
...
-     We also allow the left operand to be boolean
-     and the right operand to be a pointer;
-     although [C17:6.5.9/2] does not mention that,
-     we found it accepted by practical compiler with a strict C17 option,
-     and [C17:6.3.1.2/1] mentions a conversion
-     from scalars (which includes pointer) to booleans,
-     although it does not say that it happens
-     (it says ``When any scalar valu is converted to @('_Bool'), ...'').
-     The type of the result is the type of the left operand [C17:6.5.16/3].")
+     the requirements in [C17:6.5.16.1/1] reduce to the following four cases.")
```
acoglio, line 1279: I think it would be good to keep the sentence two lines above about the type of the result. I think it probably got deleted unintentionally.

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1529,17 +1522,14 @@
      In our currently approximate type system,
      the other two operands must have
      both arithmetic type,
-     or both the structure type,
+     or both compatible structure types,
```
acoglio, line 1525: [C17:6.5.15/3] says that the structure or union type must be the same, not just compatible.

gjurgensen: Do you understand "same type" to be a stronger judgment than "compatible"? I was interpreting them as synonyms. Do you interpret "same" to instead mean "identical"?

It is stated "two types have *compatible type* if their types are the same" [6.2.7/1]. It doesn't say "if and only if" though. And later, it adds that two types do not need to be identical to be compatible.

books/kestrel/c/syntax/validator.lisp
gjurgensen: You mean the sentence "The type of the result is the type of the left operand [C17:6.5.16/3]."? Your are right, I didn't intend to remove that; I'll add it back in.

books/kestrel/c/syntax/validator.lisp
gjurgensen: Thinking about it further, I believe you are right, "same" is stronger than "compatible". For instance, two structs declared at file scope of two different translation units may be compatible under certain criteria [6.2.7/1] but are distinct types because they are declared in different scopes [6.7.2.3/5].

gjurgensen: There is an even clearer example which I included in the new `type-compatiblep` xdoc. There may be more than one enumeration type [C17:6.2.5/16], but all enumeration types are compatible (transitively, since all enumeration types are compatible with some integer types [C17:6.7.2.2/4]).

books/kestrel/c/syntax/validator.lisp
acoglio: Yes, I meant that one. Thanks!

books/kestrel/c/syntax/validator.lisp
acoglio: Yes, the wording in [C17] is not that clear, but since sometimes it says explicitly "compatible" and other times "same", I thought it would be a stronger notion. I think "same" goes slightly beyond syntax, in the sense that, for example, `unsigned char` and `char unsigned` is the "same" type — just slight syntactic variants. But as you mention above with the example of enum types (which I didn't remember; good find!), "compatible" is definitely weaker. I think that [C17] could use better wording in its definition of "compatible" types; it says something like "... the types have the same type..." but types don't really "have" types, they "are" types already.

acoglio (approved):

PR #1764, opened 2025-05-12
[FTY] Add `deffold-map`

gjurgensen:
This is a fold for FTY types based on `deffold-reduce`. Unlike `deffold-reduce`, it does not take a `:combine` or `:default` flag, since it is a mapping fold. Instead, each case of the map is the relevant constructor applied to the recursively mapped fields.

This PR also includes some minor changes to `deffold-reduce` (a couple typos and a missing include) and to `omap`s (strengthening the `:type-prescription` rule).

### DIFF
--- a/books/kestrel/fty/deffold-map-doc.lisp
+++ b/books/kestrel/fty/deffold-map-doc.lisp
@@ -294,6 +294,19 @@
        and added to the generated ruleset described below."))
 
     (xdoc::desc
+     "Accompanying option theorems."
+     (xdoc::p
+      "For each @(tsee defoption) type specified by the @(':types') input,
+       we generate the following theorems,
+       whose exact form can be inspected with @(tsee pe) or similar command:")
+     (xdoc::ul
+      (xdoc::li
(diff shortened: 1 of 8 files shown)


books/kestrel/fty/deffold-map-doc.lisp
```diff
@@ -0,0 +1,340 @@
...
+       The name of each generated function is @('<type>-<suffix>'),
+       where @('<type>') is the type that the function operates on,
+       and @('<suffix>') is the value of this input,
+       which must be a symbol.
+       The function name is interned in the same package as @('<suffix>')."))
+
+    (xdoc::desc
+     "@(':types') &mdash; no default"
+     (xdoc::p
+      "Fixtype for which map functions must be generated.")
```
acoglio, line 74:
```suggestion
      "Fixtypes for which map functions must be generated.")
```

acoglio: I think that `:doc deffold-reduce` needs the same fix.

acoglio (approved):

PR #1766, opened 2025-05-16
Update BibTeX workshop references to cover 2018-2023 workshops

mister-walter:
Update the workshop BibTeX references to cover the 2018, 2020, 2022 and 2023 workshops.


mister-walter:
Note that I included DOI information for these entries, as many publishers require authors provide this information now. This info should be added for the previous workshop entries as well; I may take some time to do this later .

PR #1769, opened 2025-06-06
[IHS] Add two fixing theorems.

acoglio:
These are disabled because, if enabled, they cause failures elsewhere.

These are useful when `ifix` is kept disabled, which is perhaps uncommon, but avoids, like other built-ins, unnecessary case splits.

We could consider enabling these theorems and just repair the failing proofs (by disabling them, or one of them), but this is a bigger change.

### DIFF
--- a/books/ihs/basic-definitions.lisp
+++ b/books/ihs/basic-definitions.lisp
@@ -413,7 +413,13 @@ bit arbitrary but generally leads to nice lemmas.</p>"
        :exec
        (the unsigned-byte
             (logand i (the unsigned-byte
-                           (1- (the unsigned-byte (ash 1 size))))))))
+                           (1- (the unsigned-byte (ash 1 size)))))))
+
+  ///
+
+  (defthmd loghead-of-ifix
+    (equal (loghead size (ifix i))
+           (loghead size i))))
 
 (define logtail
   :short "@('(logtail pos i)') returns the high-order part of @('i'), starting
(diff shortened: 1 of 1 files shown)


ericwhitmansmith:
These rules look good to me.  However, it might be better to put them in logops-lemmas.lisp, keeping basic-definitions.lisp minimal.  Note the comment in basic-definitions.lisp that starts "This book contains only definitions, lemmas necessary to admit those definitions, and selected type lemmas..."

solswords (commented):
Do they still cause failures if they're placed enabled in logops-lemmas.lisp? I don't feel too strongly about their placement either way -- presumably they don't take any time to prove and they don't need any additional theory to do so. (And WRT the comment Eric mentioned, they could be construed as type theorems of a sort.) But I think it would be more useful to have these theorems enabled somewhere and if putting them in logops-lemmas would limit the breakage that might be better.

Btw, std/basic/arith-equivs adds an int-equiv congruence on that argument which has a similar effect (although the ifix rule can still be useful in cases where we'd need double-rewrite for congruence reasoning).

acoglio:
Putting them in `logops-lemmas.lisp`, enabled, causes no failures. I made an alternative PR with that: #1770.

I had initially put the theorems with the function definitions because, as @solswords says, they can be considered "type theorems" — we usually put those fixing theorems in a function definition, with `:hooks (:fix)`.

If the other PR is acceptable, I'll merge that one and close this one.

ericwhitmansmith:
Thanks for making the other PR.  I approved it.  BTW, I think the rationale for including "type theorems" with the definitions is just to help ensure that callers have nice :type-prescription rules automatically inferred.  These ifix rules don't help with that, as far as I can tell, so it seems good that the new PR puts them elsewhere.

acoglio:
By 'type', Sol and I meant something more general than type prescriptions: the input and output types of a function — guards, return theorems. The fixing theorems are a bit different, but somewhat related since they say how the function handles inputs of the "wrong" type.

PR #1771, opened 2025-06-09
[Std/basic] Add a return theorem for `bool->bit`.

acoglio:
Useful when `bitp` is disabled.

`bool->bit` has a bit type prescription, but this is a rewrite rule.


acoglio:
Enabled or disabled doesn't matter much to me either, so I've disabled it since @ericwhitmansmith has a preference for that.

PR #1773, opened 2025-06-13
[STD] Add acceptible declare forms to defredundant

gjurgensen:
This extends `defredundant` to accept the `irrelevant` and `optimize` declaration forms in functions. It also adds a test case for each.


books/std/util/defredundant.lisp
```diff
@@ -117,7 +117,9 @@ copying and pasting code.")
         (raise "Bad form in declare: ~x0" x))
        ((when (or (eq (car arg1) 'type)
                   (eq (car arg1) 'ignore)
-                  (eq (car arg1) 'ignorable)))
+                  (eq (car arg1) 'ignorable)
+                  (eq (car arg1) 'irrelevant)
+                  (eq (car arg1) 'optimize)))
```
solswords: minor: I think at this point I'd prefer a member call rather than more eqs in the or.

solswords (approved):
LGTM.

PR #1775, opened 2025-06-19
[r1cs] extend p1cs to take one of two primes

bendyarm:


ericwhitmansmith (changes requested):
How about making iff-consp-when-true-listp and true-listp-of-cdr local?

bendyarm:
> How about making iff-consp-when-true-listp and true-listp-of-cdr local?

Good idea, done.

PR #1776, opened 2025-06-19
[r1cs] factor out r1cs-termp

bendyarm:
Factor out so we can use `r1cs-termp` elsewhere.


ericwhitmansmith (commented):
Seems fine except I couldn't tell from the name r1cs-termp what the new notion was going to be.  How about sparse-vector-elementp?

Also, just a style thing, but please put in (declare (xargs :guard t)) instead of having the separate verify-guards.

bendyarm:
> Seems fine except I couldn't tell from the name r1cs-termp what the new notion was going to be. How about sparse-vector-elementp?
> 
> Also, just a style thing, but please put in (declare (xargs :guard t)) instead of having the separate verify-guards.

Done.

PR #1777, opened 2025-06-23
Change the `cert_param` `uses-cpp` to `uses-gcc-c17`

gjurgensen:
Similarly, changes the Makefile flag `OS_HAS_CPP` to `OS_HAS_GCC_C17`.

Updates the feature detection to search for `gcc` instead of `cpp`, and requires that `"-std=c17"` is supported.

### DIFF
--- a/books/GNUmakefile
+++ b/books/GNUmakefile
@@ -414,7 +414,7 @@ OK_CERTS := $(filter-out $(CERT_PL_USES_STP), $(OK_CERTS))
 endif
 
 ifeq ($(OS_HAS_GCC_C17), )
-$(info Excluding books that invoke GCC with \"-std=c17\": [$(CERT_PL_USES_GCC_C17)])
+$(info Excluding books that require a GCC with \"-std=c17\": [$(CERT_PL_USES_GCC_C17)])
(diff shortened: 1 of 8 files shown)


gjurgensen:
Sol, tagging you just in case you want to sanity-check the `cert_param` / Makefile / feature-testing stuff.

books/GNUmakefile
```diff
@@ -413,9 +413,9 @@ $(info Excluding books that need STP: [$(CERT_PL_USES_STP)])
 OK_CERTS := $(filter-out $(CERT_PL_USES_STP), $(OK_CERTS))
 endif
 
-ifeq ($(OS_HAS_CPP), )
-$(info Excluding books that invoke cpp: [$(CERT_PL_USES_CPP)])
-OK_CERTS := $(filter-out $(CERT_PL_USES_CPP), $(OK_CERTS))
+ifeq ($(OS_HAS_GCC_C17), )
+$(info Excluding books that invoke GCC with \"-std=c17\": [$(CERT_PL_USES_GCC_C17)])
```
solswords: This isn't incorrect, but it struck me as suggesting that some books invoke GCC without that flag. I think it would be a little bit clearer to phrase it as "Excluding books that require a GCC with \"-std=c17\""

PR #1778, opened 2025-06-24
[IHS] Add `logappn`.

acoglio:
This is an `n`-ary version of `logapp`.

### DIFF
--- a/books/ihs/basic-definitions.lisp
+++ b/books/ihs/basic-definitions.lisp
@@ -463,7 +463,7 @@ is shifted to the left by @('size') bits before the merge.</p>"
    </code>
    <p>expands to a nest of @(tsee logapp)s ending with 0:</p>
    <code>
-   (logapp size1 int1 ... (logapp sizeN intN 0)...))
+   (logapp size1 int1 ... (logapp sizeN intN 0)...)
    </code>
    <p>where the last @(tsee logapp) could be optimized
       as a @(tsee loghead) without 0,
--- a/books/ihs/basic-definitions.lisp
+++ b/books/ihs/basic-definitions.lisp
@@ -467,7 +467,8 @@ is shifted to the left by @('size') bits before the merge.</p>"
    which must be alternating bit sizes and integer values of those bits,
    the resulting nest of @(tsee logapp)s ends with 0.
    If passed an odd number of arguments,
-   the resulting nest of @(tsee logapp)s with the last argument instead.</p>"
+   the resulting nest of @(tsee logapp)s ends
+   with the last argument instead.</p>"
   (defun logappn-fn (args)
     (cond ((endp args) 0)
           ((endp (cdr args)) (car args))
--- a/books/ihs/basic-definitions.lisp
+++ b/books/ihs/basic-definitions.lisp
@@ -462,10 +462,15 @@ is shifted to the left by @('size') bits before the merge.</p>"
    (logappn 32 base
             8 padding
             24 offset)
-   </code>"
+   </code>
+   <p>If passed an even number of arguments,
+   which must be alternating bit sizes and integer values of those bits,
+   the resulting nest of @(tsee logapp)s ends with 0.
+   If passed an odd number of arguments,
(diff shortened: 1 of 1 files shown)


books/ihs/basic-definitions.lisp
```diff
@@ -454,6 +454,22 @@ is shifted to the left by @('size') bits before the merge.</p>"
...
+  :long
+  "<p>This supports readable appending of multiple bit chunks, e.g.</p>
+   <code>
+   (logappn 32 base
+            8 padding
+            24 offset)
+   </code>"
+  (defun logappn-fn (args)
+    (cond ((endp args) 0)
+          ((endp (cdr args)) 0)
```
solswords: I think rather than returning 0 on an odd number of args, I'd prefer either an error (odd number of arguments where an even number was expected) or returning that last argument as the final tail. I think that would be useful since it would let this function produce a fully general nest of logapps, not only one ending in 0.

acoglio: Good idea, I just pushed a commit.

books/ihs/basic-definitions.lisp
```diff
@@ -454,6 +454,27 @@ is shifted to the left by @('size') bits before the merge.</p>"
...
+   <code>
+   (logappn 32 base
+            8 padding
+            24 offset)
+   </code>
+   <p>If passed an even number of arguments,
+   which must be alternating bit sizes and integer values of those bits,
+   the resulting nest of @(tsee logapp)s ends with 0.
+   If passed an odd number of arguments,
+   the resulting nest of @(tsee logapp)s with the last argument instead.</p>"
```
solswords: just missing the word "ends" here

solswords (approved):
LGTM modulo one typo

books/ihs/basic-definitions.lisp
acoglio: Fixed.

ericwhitmansmith:
In the case of an even number of arguments, we get an unnecessary logapp.  For example,
:trans1 (logappn isize i ksize k jsize j)
gives
(LOGAPP ISIZE I (LOGAPP KSIZE K (LOGAPP JSIZE J '0)))
It might be better (and equivalent) to use loghead:
(LOGAPP ISIZE I (LOGAPP KSIZE K (LOGHEAD JSIZE J)))
(The case of an even number of arguments is somewhat against the spirit of logapp, which doesn't take a size for the most significant argument.  I didn't like that aspect of logapp, so I defined my own bvcat function.)

I think the documentation could also be clarified, perhaps by giving examples of the 2 cases. This macro is a bit tricky to describe because the difference between the 2 cases is that the penultimate argument (size of the last argument) is omitted.  It would also be good to indicate which part of the input becomes the most significant part of the result.

BTW, did you consider changing logapp itself (make a helper function and make logapp be this macro we are discussing), since the new behavior is backward-compatible?

acoglio:
I've expanded and clarified the documentation.

The final 0 is to keep the macro simple and uniform, and an actual nest of `logapp`s and no other `log...` function.

Generalizing `logapp` itself is an interesting idea, but perhaps it's a bigger change than we may want to do at this point?

books/ihs/basic-definitions.lisp
```diff
@@ -454,6 +454,45 @@ is shifted to the left by @('size') bits before the merge.</p>"
...
+(defsection logappn
+  :short "N-ary version of @(tsee logapp)."
+  :long
+  "<p>When passed an even number of arguments,</p>
+   <code>
+   (logappn size1 int1 ... sizeN intN)
+   </code>
+   <p>expands to a nest of @(tsee logapp)s ending with 0:</p>
+   <code>
+   (logapp size1 int1 ... (logapp sizeN intN 0)...))
```
ericwhitmansmith:
```suggestion
   (logapp size1 int1 ... (logapp sizeN intN 0)...)
```
To match how the parens are done below.

ericwhitmansmith (approved):
LGTM.  Thanks for the clear documentation.

PR #1779, opened 2025-06-25
Changed part of p1cs to logic mode

akitaki79:


books/kestrel/crypto/r1cs/sparse/p1cs.lisp
```diff
@@ -472,13 +475,20 @@
...
+           (< first-digit-pos (length base-name))  ; Added this bound check
            (let ((num (str::strval (subseq base-name first-digit-pos (length base-name)))))
              (if num
                  (list (subseq base-name 0 first-digit-pos) num)
                nil))))))
 
 (defun bit-var-range-starting-with (starting-base-and-num last-num bit-vars)
-  (declare (xargs :measure (len bit-vars)))
+  (declare (xargs :guard (and (true-listp starting-base-and-num)
+                              (>= (len starting-base-and-num) 2)
```
bendyarm:
```suggestion
                              (= (len starting-base-and-num) 2)  
```

books/kestrel/crypto/r1cs/sparse/p1cs.lisp
```diff
@@ -514,10 +556,11 @@
           (bit-var-ranges rest-bit-vars))))
 
 (defun concatenate-bit-var-ranges (ranges) ; each range is a string; can be a singleton var
+  (declare (xargs :guard (string-listp ranges)))
   (cond ((null ranges) "")
         ((null (cdr ranges)) (car ranges))
         (t (concatenate 'string (car ranges) ", " (concatenate-bit-var-ranges (cdr ranges))))))
-
+
```
bendyarm:
```suggestion

```

bendyarm: In other words, strip the spaces off the end of the line.

books/kestrel/crypto/r1cs/sparse/p1cs.lisp
```diff
@@ -472,13 +475,20 @@
            (first-digit-pos (position-if-digit chars 0)))
       (and (natp first-digit-pos)
            (> first-digit-pos 0)
+           (< first-digit-pos (length base-name))  ; Added this bound check
```
ericwhitmansmith: Can this check ever be violated?  If not, it would be good to prove that fact and remove this check.

ericwhitmansmith (commented):
LGTM pending one question about an added bounds check.

books/kestrel/crypto/r1cs/sparse/p1cs.lisp
gjurgensen: Agreed, I imagine it can be proven.

@akitaki79, I think there are two relevant facts (which need to be proven). First, that `position-if-digit` returns something smaller than the length if it doesn't return `nil` (this statement may need to be generalized to accomodate nonzero values `next-pos`). Second, that the length of `(acl2::explode base-name)` corresponds to the length of `base-name` (I imagine there is already such a rule, but good to keep in mind the reasoning).

books/kestrel/crypto/r1cs/sparse/p1cs.lisp
```diff
@@ -455,13 +455,22 @@
...
+  (declare (xargs :guard (and (character-listp char-list)
+                              (natp next-pos))))
   (if (endp char-list)
       nil
     (if (member (car char-list) *digits*)
         next-pos
       (position-if-digit (cdr char-list) (+ next-pos 1)))))
 
+
+(defthm position-if-digit-bounds
```
gjurgensen, line 467: I think this theorem would be slightly better as a linear rule:
```lisp
(defthm position-if-digit-bounds
  (implies (position-if-digit char-list next-pos)
           (<= (position-if-digit char-list next-pos)
               (+ next-pos (len char-list))))
  :rule-classes :linear)
```
The statement of this theorem is a little nicer since it doesn't bring in this `original-len` variable. But is has to be a linear rule for some technical reasons.

Briefly, the rule you have now works as a rewrite rule because it looks for instances of `(<= (position-if-digit char-list next-pos) original-len))` in the proof goal and rewrites it to `t` if it can establish the hypotheses. In the later guard proof, a term of exactly this shape appears. In contrast, my suggested rule, if it were a rewrite rule, would be trying to rewrite terms which look like: `(<= (position-if-digit char-list next-pos) (+ next-pos (len char-list)))`. But no such term shows up in the target guard proof. It works as a linear rule, because linear rules won't try to match that whole term. Instead, the linear rule will just look for `(position-if-digit char-list next-pos)` (called the "trigger term" and applies the rule from there, adding the inequality to some database behind the scenes if it can establish the hypothesis.

Like I said, this is pretty technical. You could see `:doc linear` and `:doc rewrite` for more in-depth discussions.

akitaki79: I will add this in the next PR.

bendyarm (approved):

PR #1780, opened 2025-06-26
Add oset non-emptiness predicate with reasoning support.

acoglio:


gjurgensen (approved):
LGTM.

solswords (commented):
I don't really get why this is needed -- we already have rules NEVER-IN-EMPTY and IN-HEAD which show effectively that `(emptyp x)` is equivalent to `(forall k (not (in k x)))`, with HEAD the witness for nonemptiness. Can you add some explanation for where/why the new functions are useful?

acoglio:
I've expanded the documentation to provide some motivation. I've been using `head` as a witness (as you say) for a while, but I had to use `:use` hints and disable `in-head`, which I found clumsy. With this approach, I just enable a couple of rules and rewriting does its job. It's possible that I've missed some other simple way to do this without the need for this extension.

solswords (approved):

PR #1783, opened 2025-06-28
Added guards for remaining p1cs functions

akitaki79:
I also converted nums-to-strings from a constant to a function and added some other proofs to admit the guards for `p1cs-coefficient` & `p1cs-term`.  These proofs are really inefficient right now, so I will try to get them working faster in the future.


books/kestrel/crypto/r1cs/sparse/p1cs.lisp
```diff
@@ -23,6 +23,8 @@
 ;; If you don't do this you get a program-mode-only str::pretty
 (include-book "std/strings/pretty" :dir :system)
 
+(include-book "kestrel/alists-light/top" :dir :system)
```
gjurgensen: I would make this local. I.e., `(local (include-book ...))`.

Local events are used to certify a book, but not "exported" to anyone including this book. See `:doc local`.

books/kestrel/crypto/r1cs/sparse/p1cs.lisp
```diff
@@ -158,20 +162,41 @@
...
+  (declare (xargs :guard (and (natp fp) (< 1 fp) (integerp x))))
   (if (> x (/ fp 2))
       (str::int-to-dec-string (- x fp))
     nil))
 
+(defun well-formed-term-p (term)
+  (declare (xargs :guard t))
+  (and (true-listp term)
+       (= (len term) 2)
+       (integerp (first term))))
```
gjurgensen: I think you can avoid introducing a new function here and use `sparse-vector-elementp` instead. Similarly, you can use `sparse-vectorp` instead of `all-well-formed-terms-p`.

books/kestrel/crypto/r1cs/sparse/p1cs.lisp
```diff
@@ -196,7 +240,23 @@
...
+                                      assoc-equal))))
+
 (defun p1cs-term (fp term)
+  (declare (xargs :guard (and (member fp (list (fp1) (fp2)))
+                              (well-formed-term-p term))
+                  :guard-hints (("Goal" :in-theory (disable p1cs-coefficient
+                                                            p1cs-negative
+                                                            assoc-equal
+                                                            well-formed-term-p)))
+                  :guard-debug t))
```
gjurgensen: It is best to remove `:guard-debug t` after you've succeeded in proving the guard.

gjurgensen (approved):

books/kestrel/crypto/r1cs/sparse/p1cs.lisp
ericwhitmansmith: I would recommend not including alists-light/top.  Instead, include (locally) only the books from that directory that you need.

bendyarm:
^^^ @akitaki79

gjurgensen:
Looks good. One thing -- did you try running this on any large R1CS circuits as Eric suggested? If this change from constant to function doesn't actually cause a slowdown, perhaps we could remove the `add-io-pairs` caching.

books/kestrel/crypto/r1cs/sparse/p1cs.lisp
```diff
@@ -7,7 +7,7 @@
 ;
 ; Author: Eric McCarthy (bendyarm on GitHub)
 
-(in-package "R1CS")
+(In-package "R1CS")
```
bendyarm:
```suggestion
(in-package "R1CS")
```

books/kestrel/crypto/r1cs/sparse/p1cs.lisp
```diff
@@ -195,8 +202,8 @@
            (stringp (cadr (assoc-equal key alist))))
   :hints (("Goal" :in-theory (enable assoc-equal))))
 
-;; This shows there are no collisions:
-;; (assert-equal (len *nums-to-strings*) (len (remove-duplicates (strip-cars *nums-to-strings*))))
+;;This shows there are no collisions:
+;;(assert-equal (len *nums-to-strings*) (len (remove-duplicates (strip-cars *nums-to-strings*))))
```
bendyarm: This comment should be updated to reflect the other changes.   Run the assert-equal to make sure it was updated correctly.

bendyarm:
> Looks good. One thing -- did you try running this on any large R1CS circuits as Eric suggested? If this change from constant to function doesn't actually cause a slowdown, perhaps we could remove the `add-io-pairs` caching.

The change currently does not "first calculate the constant and then define the function as exactly that constant" (quote from Zulip).  As currently structured, it does calculate each constant only once, triggered by `add-io-pairs`.  Maybe better would be to calculate the two constants outside the function and then use those constants in the function.  That would obviate the need for the add-io-pairs.

akitaki79:
@gjurgensen @bendyarm I ran a sample R1CS circuit on the different p1cs timings and got the following:

`nums-to-strings` as a constant:
; (EV-REC *RETURN-LAST-ARG3* ...) took 
; **1.52 seconds realtime**, 0.16 seconds runtime
; (9,244,848 bytes allocated).

`nums-to-strings` as a nullary function (no `add-io-pairs`):
; (EV-REC *RETURN-LAST-ARG3* ...) took 
; **336.06 seconds realtime**, 335.74 seconds runtime
; (745,435,921,088 bytes allocated).`

`Nums-to-strings` as a function (with `add-io-pairs`)
; (EV-REC *RETURN-LAST-ARG3* ...) took 
; **1.55 seconds realtime**, 0.11 seconds runtime
; (8,791,600 bytes allocated).`

Seems like the `add-io-pairs` brought the performance back to where it was.

gjurgensen (approved):

gjurgensen:
@bendyarm, just so I'm clear what you're asking, are you suggesting that the function be introduced by a `make-event` such that the function body is the result of computing this constant?

akitaki79:
@bendyarm I'm not sure if this was the approach you had in mind but what I did was calculate the constants outside the function and then use those constants in the function. It looks something like this:

```
(defconst *nums-to-strings-fp1*
   ;same code as original function except for fp1 )


(defconst *nums-to-strings-fp2*
   ;same code as original function except for fp2 )

;then the actual function
(defund nums-to-strings (prime)
  (declare (xargs :guard (member prime (list (fp1) (fp2)))))
  (cond ((equal prime (fp1)) *nums-to-strings-fp1*)
        ((equal prime (fp2)) *nums-to-strings-fp2*)
        (t (er hard 'nums-to-strings
               "Unsupported prime: ~x0. Currently supported primes are ~x1 and ~x2."
               prime (fp1) (fp2)))))
```

I timed this on the same R1CS circuit and got the following performance, which seemed pretty similar to the `add-io-pairs` approach. 

; (EV-REC *RETURN-LAST-ARG3* ...) took 
; **1.54 seconds realtime,** 0.13 seconds runtime
; (8,785,792 bytes allocated).

gjurgensen:
Disregard my previous comment. I understand now and this is simpler than what I had in mind.

ericwhitmansmith (commented):
It looks like changes have been made to indenting for else-branches of IFs and for the bodies of DEFTHMs.  That is making the diff harder to read and departs from the typical 2 spaces used for such things.

Other than that this looks good though I didn't review it in detail.

akitaki79:
> It looks like changes have been made to indenting for else-branches of IFs and for the bodies of DEFTHMs. That is making the diff harder to read and departs from the typical 2 spaces used for such things.

I'll definitely fix the indenting now that my emacs handles it properly. One thing you can do when viewing the diff for now is to hide whitespaces like shown in the photo below, which could help clean up some of the extra stuff.

image

ericwhitmansmith:
> One thing you can do when viewing the diff for now is to hide whitespaces like shown in the photo below, which could help clean up some of the extra stuff.

Cool tip!

ericwhitmansmith (approved):

PR #1784, opened 2025-06-28
[IHS] Update README.

acoglio:


MattKaufmann (approved):
Looks good to me.  I suppose you could change "cert.pl" to "../build/cert.pl", but that opens up whether you'd also say something about the --acl2 and -j arguments.  I trust your judgment on that.  Thanks for doing this update.

solswords (approved):
Seems fine. I do think if a directory's books build with the standard build system, the readme doesn't really have to say how to build them, or not to move things out of the standard directory structure, etc.

MattKaufmann:
For what it's worth, I agree with @solswords .  I think that part of the README may well have been written before there was regression testing at all (hence well before the current build system was added).

acoglio:
I agree: I just pushed a commit that removes that text from the README.

PR #1785, opened 2025-06-30
[RAC] Change array inialization

voskh0d:
std::arrays should be able to be initialized using {{}} and {} (braces elision). Before, std::arrays were only initialized using {} but for some other (older) compilers, the full syntax must be supported. Brace elision is not implemented, "{}" now reports an error


voskh0d:
@mayankmanj

PR #1787, opened 2025-07-01
[doc] Add how-to-contribute topic

gjurgensen:
Currently, the repo README "how to contribute" link points to `git-quick-start`. I think it would be nicer to have a dedicated "how to contribute" topic without the focus on git (which many new contributors will know, and which may be discussed separately for those who do not).

The contact information at the bottom of the topic was largely taken from `github-commit-code-using-push`.


ericwhitmansmith (approved):
This looks good to me.  A few suggestions:
1. Name it :doc contributing (up to you)?
2. Add a link to :doc community.
3. There are some other things to mention regarding not breaking the build (set cert_params as needed for Lisp versions, for ACL2 variants (e.g., ACL2(r)), and for any external tools needed; ensure your books don't rely on any other external tools or configurations (e.g., environment vars); avoid small timeouts in ACL2s developments; avoid absolute pathnames (e.g., paths to files in a particular user's home dir)).  Should these be discussed here or in some other topic?

gjurgensen:
@ericwhitmansmith I pushed some changes which address 2 and 3. I figure for 3, we can ask people to be thoughtful about how their book might depend on their local machine/environment, but we don't expect contributors to go beyond that (e.g. building on multiple machines before submitting a PR). 

I added the issues you outlined as examples of how a build could be affected by the machine/environment. I think it works how it is now, but if the list were to get any longer it probably should be factored out into a separate topic. My goal for this `how-to-contribute` topic is to give a concise, high-level overview of just the most important points. I figure it is best to link to other topics where any extended discussion is needed of any particular details.

For 1, I would vote in favor of keeping the `:doc how-to-contribute` name.

ericwhitmansmith:
LGTM, thanks!

PR #1791, opened 2025-07-02
[doc] Remove text about manual needing glucose.

ericwhitmansmith:
I believe it is no longer required.


solswords (commented):
I haven't checked whether it works without glucose - if it does, that's good and we can maybe remove that requirement. If not, we should be able to conditional the inclusion of books that require it (I think we already do this for some things). But I don't think CI tests the build without glucose, does it? So maybe it works, but it isn't necessarily tested regularly. Maybe it should say something to this effect?

acoglio (approved):

ericwhitmansmith:
I routinely build the manual on my laptop that doesn't have glucose.  So I think there is nothing special about glucose and the manual any more (there is a comment in books/doc/top that seems to confirm that).  It is a risk that someone might include in the manual a book that needs an external tool (glucose, z3, STP, IPASIR, etc.).  We could try to create a leeroy plan that intentionally excludes all such tools.  Is it possible to override the decisions make by features.sh?

PR #1792, opened 2025-07-03
[doc] Rewrite best-practices

gjurgensen:
The current `best-practices` is a small draft topic. This rewrites the topic from scratch (keeping the subtopics). The content is largely a style guide and aims for widely agreed upon conventions.

If anyone objects to any particular point, I'd be happy to discuss removing or qualifying it. Again, this guide is intended to apply to the community books in general and so should get broad buy-in from the community.

Likewise, I'm happy to hear any suggested additions.

### DIFF
--- a/books/doc/practices.lisp
+++ b/books/doc/practices.lisp
@@ -77,10 +77,10 @@
               (xdoc::codeblock ";This is a comment."))
     (xdoc::li "Books should begin with a copyright and licensing header (which
                may just point to the repository's top-level LICENSE file).")
-    (xdoc::li "Inline comments which are intended to be indented at the same
-               level as a form would be at a given position should begin with
-               two semicolons. This is a standard practice in Common Lisp, and
-               editors like Emacs will indent such comments appropriately.")
+    (xdoc::li "Use two semicolons for inline comments which should be indented
+               as a form would be. This is a standard practice in Common Lisp,
+               and editors like Emacs will indent such comments
+               appropriately.")
     (xdoc::li "Avoid commented code which would disrupt the s-expression
                balance if uncommented. For instance, instead of this:"
               (xdoc::codeblock
@@ -102,11 +102,15 @@
    (xdoc::h3 "Documentation")
    (xdoc::ul
     (xdoc::li (xdoc::seetopic "xdoc" "XDOC")
-              " short forms should end with a period, even if they are not a
-               full sentence."))
+              " short forms should be in ``sentence case,'' not ``title case.''
+               They should end with a period, even if they are not a full
+               sentence.")
+    (xdoc::li (xdoc::seetopic "xdoc" "XDOC")
+              " short forms should not span multiple paragraphs, or include
+               code blocks, tables, headers, etc."))
    (xdoc::h3 "Naming Conventions")
    (xdoc::ul
-    (xdoc::li "Consider ending predicates with ``p'' (or ``-p'', especially if
+    (xdoc::li "Consider ending predicates with ``p'' (or ``-p,'' especially if
                needed for clarity). This is a convention inherited from Common
                Lisp.")
     (xdoc::li "For ordinary identifiers, avoid uppercase and use hyphens to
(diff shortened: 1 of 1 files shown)


books/doc/practices.lisp
```diff
@@ -27,28 +27,154 @@
...
+              (xdoc::codeblock
+               "(if x"
+               "    y"
+               "  z)")
+              "That is, 4 spaces for the ``then'' term, and 2 spaces for the
+               ``else'' term. Emacs may indent this incorrectly by default."))
+   (xdoc::h3 "Packages")
+   (xdoc::ul
+    (xdoc::li "The use of packages is highly encouraged if you expect your
+               library to be used by others. This helps to avoid name clashes
```
acoglio: ... or if you include your library in the manual.

acoglio (approved):

books/doc/practices.lisp
gjurgensen: Perhaps this should be simplified just to "The use of packages is highly encouraged."?

acoglio: Up to you. I guess it's good to give motivation, but on the other hand other best practices are not explicitly motivated.

gjurgensen: I would still keep the "This helps to avoid name clashes" part, but drop the "if you expect your library to be used by others".

acoglio: Yes, that covers it all.

books/doc/practices.lisp
```diff
@@ -27,28 +27,154 @@
...
+   (xdoc::h3 "Comments")
+   (xdoc::ul
+    (xdoc::li "Comments should begin with a single space. E.g.:"
+              (xdoc::codeblock "; This is a comment.")
+              (xdoc::i "not")
+              (xdoc::codeblock ";This is a comment."))
+    (xdoc::li "Books should begin with a copyright and licensing header (which
+               may just point to the repository's top-level LICENSE file).")
+    (xdoc::li "Inline comments which are intended to be indented at the same
+               level as a form would be at a given position should begin with
```
ericwhitmansmith: Can you tighten the wording here?  Maybe delete "at a given position".

books/doc/practices.lisp
```diff
@@ -27,28 +27,154 @@
...
+               "  baz"
+               "  )"))
+    (xdoc::li "Consider whether comments would be more appropriate as "
+              (xdoc::seetopic "xdoc" "XDOC")
+              " documentation."))
+   (xdoc::h3 "Documentation")
+   (xdoc::ul
+    (xdoc::li (xdoc::seetopic "xdoc" "XDOC")
+              " short forms should end with a period, even if they are not a
+               full sentence."))
```
ericwhitmansmith: And should probably not capitalize each word?  And definitely should not include headings (that messes up the display of subtopics).  Also, you may want to consider what :doc defxdoc says about the :short.

books/doc/practices.lisp
```diff
@@ -27,28 +27,154 @@
...
+               the naming of rewrite rules."))
+   (xdoc::h3 "Formatting")
+   (xdoc::ul
+    (xdoc::li "Multiline @('if') forms should be formatted as follows:"
+              (xdoc::codeblock
+               "(if x"
+               "    y"
+               "  z)")
+              "That is, 4 spaces for the ``then'' term, and 2 spaces for the
+               ``else'' term. Emacs may indent this incorrectly by default."))
```
ericwhitmansmith: Tell them how to fix it?

books/doc/practices.lisp
```diff
@@ -27,28 +27,154 @@
...
+   (xdoc::h3 "Packages")
+   (xdoc::ul
+    (xdoc::li "The use of packages is highly encouraged if you expect your
+               library to be used by others. This helps to avoid name clashes
+               and polluting the default namespace. See @(see
+               working-with-packages) for further guidance.")
+    (xdoc::li "If you are defining a new package name, check to see that the
+               name is not already in use."))
```
ericwhitmansmith: Do you want to say how they should check?  Grep for defpkg in all .lsp files (and .acl2 files)?

books/doc/practices.lisp
```diff
@@ -27,28 +27,154 @@
...
+   (xdoc::ul
+    (xdoc::li "Try to avoid subgoal hints. These can make proofs brittle and
+               difficult to maintain. If it is not sufficient to apply hints to
+               the top-level goal, consider factoring out a subgoal into an
+               auxiliary lemma.")
+    (xdoc::li "Generally speaking, the less hints for a proof, the
+               better. Consider whether @(':use') or @(':expand') hints may
+               suggest good general-purpose rules (this is not always the
+               case).")
+    (xdoc::li "Try to avoid brittle @(see proof-builder) instruction hints
```
ericwhitmansmith: Maybe ":instructions" instead of "instruction hints" as :instructions are actually siblings of the :hints.

books/doc/practices.lisp
gjurgensen: "Sentence case" instead of "title case"? I agree.

books/doc/practices.lisp
gjurgensen: There is both the top level `:instructions` keyword argument to `defthm`, as well as an `:instructions` keyword argument for hints. I didn't realize that when I wrote this. I suppose the advice should apply to both.

books/doc/practices.lisp
gjurgensen: I added a `grep` command. It ended up pretty ugly, but folks can copy/paste it into their terminal. Still, if anyone thinks of a more succinct `grep` for this, I'd be all ears.

gjurgensen:
> You might want to address file naming conventions (lowercase, dash-separated) and also the convention to use .lisp only for books that certify (.lsp for other stuff).

Those rules are present in the preexisting `file-extensions` and `file-names` subtopics. Perhaps that is sufficient.

ericwhitmansmith:
> > You might want to address file naming conventions (lowercase, dash-separated) and also the convention to use .lisp only for books that certify (.lsp for other stuff).
> 
> Those rules are present in the preexisting `file-extensions` and `file-names` subtopics. Perhaps that is sufficient.

Now I see that these are subtopics, so all is good.

PR #1793, opened 2025-07-03
Doc GitHub

ericwhitmansmith:
This simplifies and clarifies the 2 sets of github instructions.  I tried to make small commits and include rationale in the log messages, but it may be easier to simplify read through the new versions of the 2 topics.


books/system/doc/acl2-doc.lisp
```diff
@@ -45256,51 +45256,45 @@ current fast alists."
...
 
  <p>The following commands will update your directory to match the latest
- contents of the github repository (on the web).</p>
+ contents of the main ACL2 repository on GitHub.</p>
 
  @({
  git fetch --all
- git merge remotes/origin/master
+ git merge -m \"Merge.\" remotes/origin/master
  })
```
gjurgensen, line 45288: Perhaps just `git pull`?

books/system/doc/acl2-doc.lisp
```diff
@@ -45311,118 +45305,102 @@ current fast alists."
...
  git fetch --all
- git merge remotes/origin/master
+ git merge -m \"Merge.\" remotes/origin/master
  })</li>
 
  <li>Build an executable.
 
  @({
- (time nice make LISP=<your_lisp>) >& make.log
+ time nice make update LISP=<your_lisp>
```
gjurgensen, line 45322: I would remove the `time nice`.

books/system/doc/acl2-doc.lisp
```diff
@@ -45311,118 +45305,102 @@ current fast alists."
...
 
- @({
- (time nice make -j 8 regression) >& make-regression-finish-1.log
- })
+ <li>Commit your changes.  The following command will commit your changes
+ locally. The @('-a') option tells git to automatically add to the commit any
+ changed or new files. The @('-m ...') option is a log message whose first line
+ should be a summary of your changes and other lines may give more details.
+ You are welcome to replace the @('-m ...') option by @('-F <filename>'), where
+ @('<filename>') is the name of a file that contains your log message.
```
gjurgensen: Is it common to use `-F`? I would think it is more common to just omit `-m` and let your terminal open an editor for you to write the commit message. (I suppose that probably doesn't work very well in an emacs shell though. Maybe `-F` makes more sense in that context.)

books/system/doc/acl2-doc.lisp
```diff
@@ -45311,118 +45305,102 @@ current fast alists."
...
- to avoid recertifying more than is necessary.  For example:
+ <li>If there were failures, then go back to Step 3 above to make appropriate
+ changes and re-test.</li>
 
- @({
- (time nice make -j 8 regression) >& make-regression-finish-1.log
- })
+ <li>Commit your changes.  The following command will commit your changes
+ locally. The @('-a') option tells git to automatically add to the commit any
+ changed or new files. The @('-m ...') option is a log message whose first line
```
gjurgensen: I feel like using `git commit -a` isn't a great practice. It is very easy to accidentally commit a file unintentionally. Perhaps we should suggest `git add`-ing the individual files. Or `git add -a` but then checking `git status` to see what is staged before committing? On the other hand, this is just a quick-start guide. We don't want to drill down too much into the details. So maybe its fine as is. Whatever you think.

books/system/doc/acl2-doc.lisp
```diff
@@ -45311,118 +45305,102 @@ current fast alists."
...
+  :long "<p>This guide is for contributors who will commit to the repository
+  rarely (e.g., a few times a year). If you find yourself committing more often,
+  see @(see github-commit-code-using-push).</p>
 
- A nice result of using pull requests is that all changes will be peer-reviewed
+ <p> If you do not plan to commit your own changes, see @(see git-quick-start)
+ instead.</p>
+
+ <p>A nice result of using pull requests is that all changes will be peer-reviewed
  before being committed.  Also, we sometimes call this method the <i>Fork and
```
gjurgensen: Perhaps it should be mentioned that even frequent contributors with push access should open PRs on occasion when they want their changes to be reviewed (like we are doing right now :smile:).

books/system/doc/acl2-doc.lisp
```diff
@@ -45311,118 +45305,102 @@ current fast alists."
  <p>After you have joined the project, you can proceed as follows when you are
  ready to contribute.</p>
 
- <h3>Change and Test</h3>
+ <h3>Make Changes and Test Them</h3>
 
  <ol>
 
- <li>Update as in (B) above:
+ <li>Before beginning your edits, update, as in (B) above:
```
acoglio, line 45312: Consider not repeating the git commands here, and just saying something like "update using the git commands in (B) above".

books/system/doc/acl2-doc.lisp
acoglio, line 45322: I would remove `time` and `nice`.

books/system/doc/acl2-doc.lisp
```diff
@@ -45311,118 +45305,102 @@ current fast alists."
...
- release-notes-books) XDOC topic in @('books/doc/relnotes.lisp').</li>
+ <li>Make book changes.  See the guidelines in the @(see how-to-contribute) topic
+  (e.g., about updating the book release notes).</li>
 
  <li>Run a regression.
 
  @({
- (time nice make -j 8 regression-fresh) >& make-regression.log
- })</li>
+ (time nice make -j 8 regression) >& make-regression.log
```
acoglio, line 45331: I would avoid `time` and `nice`. Rationale: they are not essential.

books/system/doc/acl2-doc.lisp
```diff
@@ -45311,118 +45305,102 @@ current fast alists."
...
-
- </ol>
-
- <h3>Update, and Iterate If Necessary</h3>
+ @({
+ git commit -a -m '<some message, with descriptive first line>'
+ })</li>
 
- <p>Update again as in (B) above:</p>
+ <li>Merge in remote changes, if any, by updating again as in (B) above:
```
acoglio, line 45358: Again, I would consider not repeating the git commands here.

books/system/doc/acl2-doc.lisp
```diff
@@ -45452,125 +45430,102 @@ current fast alists."
...
  <h2>(C) CONTRIBUTING</h2>
 
- <h3>Change and Test</h3>
+ <h3>Make Changes and Test Them</h3>
 
  <ol>
 
  <li>Before beginning your edits, update, as in (B) above:
 
  @({
```
acoglio, line 45449: Consider not repeating the git commands here.

books/system/doc/acl2-doc.lisp
```diff
@@ -45452,125 +45430,102 @@ current fast alists."
...
  git fetch --all
- git merge remotes/upstream/master
+ git merge -m \"Merge.\" remotes/upstream/master
  })</li>
 
  <li>Build an executable.
 
  @({
- (time nice make LISP=<your_lisp>) >& make.log
+ time nice make update LISP=<your_lisp>
```
acoglio, line 45457: Consider no `time` or `nice`.

books/system/doc/acl2-doc.lisp
```diff
@@ -45452,125 +45430,102 @@ current fast alists."
...
- Community Books' release notes &mdash; i.e., the appropriate @(see
- release-notes-books) XDOC topic in @('books/doc/relnotes.lisp').</li>
+ <li>Make book changes.  See the guidelines in the @(see how-to-contribute) topic
+  (e.g., about updating the book release notes).</li>
 
  <li>Run a regression.
 
  @({
- (time nice make -j 8 regression-fresh) >& make-regression.log
+ (time nice make -j 8 regression) >& make-regression.log
```
acoglio, line 45466: No `time` or `nice` again.

books/system/doc/acl2-doc.lisp
```diff
@@ -45452,125 +45430,102 @@ current fast alists."
...
 
- </ol>
-
- <h3>Update, and Iterate If Necessary</h3>
-
- <p>Update again as in (B) above:</p>
+ <li>Merge in remote changes, if any, by updating again as in (B) above:
 
  @({
  git fetch --all
```
acoglio, line 45496: Just refere to (B) above.

acoglio (approved):
Looks good. Left some suggestions for consideration.

books/system/doc/acl2-doc.lisp
ericwhitmansmith: Good idea.  I will remove the duplicated commands.

books/system/doc/acl2-doc.lisp
ericwhitmansmith: Would be nice, but I didn't see a way supply a log message for the merge as part of the git pull command itself.

books/system/doc/acl2-doc.lisp
ericwhitmansmith: Yeah, I think people will often be doing this inside an emacs shell, so it's probably best to explicitly supply the message.  Previously, there was verbiage here about what to do if vi pops up!

books/system/doc/acl2-doc.lisp
gjurgensen: Ok, its fine how it is.

FWIW, when I `git pull`, git opens a text editor to write a merge message when necessary. But if we expect people to want to supply commit messages in other ways (either with `-m` or `-F`) then I suppose `git pull` may not work.

books/system/doc/acl2-doc.lisp
ericwhitmansmith: Good idea.  I will add at least a sentence about that.  Do you think I should add detailed instructions for doing that (make a new branch, commit, push the branch to the acl2 remote, etc.)?  These would be different instructions than in :doc github-commit-code-using-pull-requests because those involve a user's own fork.

gjurgensen: I don't think its necessary. I personally use my own fork rather than creating a branch in the main repo anyway.

books/system/doc/acl2-doc.lisp
ericwhitmansmith: I had similar qualms about the `-a` option.  Since you brought it up too, I'll change the instructions to have an explicit `git add` step.

books/system/doc/acl2-doc.lisp
ericwhitmansmith: I'll do that, thanks.

acoglio (approved):

PR #1800, opened 2025-07-11
bibtex parser to XDOC documentation for ACL2 workshops

akitaki79:


books/workshops/references/bibtex-parser.lisp
```diff
@@ -0,0 +1,399 @@
...
+
+; necessary books
+(include-book "std/strings/top" :dir :system)
+(include-book "std/io/top" :dir :system)
+
+
+; Helper functions for string processing
+
+(defun whitespace-char-p (c)
+  "Check if character is whitespace and/or comma"
```
gjurgensen: Common Lisp style docstrings aren't used often in ACL2. I think its fine to leave them as-is. You could also consider using `define` instead of `defun`. `define` is a macro which introduces functions and also let's you specify a `:short` that your could use instead of these docstrings.

books/workshops/references/bibtex-parser.lisp
```diff
@@ -0,0 +1,399 @@
...
+  (declare (xargs :guard (characterp c)))
+  (member c '(#\Space #\Tab #\Newline #\Return #\,)))
+
+(defun skip-whitespace (str pos)
+  "Skip whitespace characters starting from position pos"
+  (declare (xargs :guard (and (stringp str)
+                              (natp pos)
+                              (<= pos (length str)))
+                  :measure (nfix (- (length str) pos))))
+  (if (or (not (natp pos)) (>= pos (length str)))
```
gjurgensen: If we assume the guard holds, this condition is false. But I assume you still need these checks for the termination proof (the guard is not assumed for the termination proof). Still, since we know that the guard will hold every time we run the function, we'd like to avoid checking this condition at execution since we know what its value will be.

This is a common pattern which we solve with `mbt`. `mbt` stands for "must be true". Under the logical interpretation, `(mbt x)` is just `x`. But during execution, it will be `t`, which is is safe because it has a guard of `x`. So you could consider rewriting your condition as `(if (not (mbt ...)) ...)`.

books/workshops/references/bibtex-parser.lisp
```diff
@@ -0,0 +1,399 @@
...
+      pos
+    (if (not (whitespace-char-p (char str pos)))
+        pos
```
gjurgensen: NIT: Since both these `then` branches lead to the same term, I'd rewrite as a single flat `if`. That is, `(if a x (if b x y))` becomes `(if (or a b) x y)`.

books/workshops/references/bibtex-parser.lisp
```diff
@@ -0,0 +1,399 @@
...
+                (characterp c)
+                (natp start-pos)
+                (<= start-pos (length str))
+                (find-char str c start-pos))
+           (<= start-pos (find-char str c start-pos))))
+
+(defthm len-of-explode-is-length
+  (implies (stringp str)
+           (equal (len (explode str)) (length str)))
+  :rule-classes (:forward-chaining))
```
gjurgensen: I don't think this should be a `:forward-chaining` rule. Would it work to change it to a normal rewrite rule? (I.e. omit the `:rule-classes` keyword argument.)

The reason I say it may not be a good `:forward-chaining` rule is because it is triggered by a very broad hypothesis and concludes something very specific. There are a lot of cases where something is a `stringp` and we don't care about `(len (explode str))`.

books/workshops/references/bibtex-parser.lisp
```diff
@@ -0,0 +1,399 @@
...
+
+(defun parse-bibtex-entries-simple (str)
+  "Simpler version that finds all @ positions first, then parses each"
+  (parse-bibtex-entries-from-positions str (find-all-at-positions str 0 nil)))
+
+(defun parse-bibtex-file (filename state)
+  "Parse a BibTeX file and return an alist from citation keys to entries"
+  (declare (xargs :guard (stringp filename)
+                  :stobjs state
+                  :verify-guards nil))
```
gjurgensen: Can this by guard-verified easily?

books/workshops/references/xdoc-generation.lisp
```diff
@@ -0,0 +1,154 @@
...
+  "Format entry type for display"
+  (cond ((equal entry-type "article") "Journal Article")
+        ((equal entry-type "inproceedings") "Conference Paper")
+        ((equal entry-type "incollection") "Book Chapter")
+        ((equal entry-type "techreport") "Technical Report")
+        (t (concatenate 'string
+                        (string-upcase (subseq entry-type 0 1))
+                        (subseq entry-type 1 nil)))))
+
+(defun generate-bibtex-entry-html (entry)
```
gjurgensen: NIT: I would call this XDOC instead of HTML. It probably produces valid HTML, but we are really targeting XDOC, which I think it technically not quite a subset of HTML.

books/workshops/references/xdoc-generation.lisp
```diff
@@ -0,0 +1,154 @@
...
+                     (concatenate 'string ", " (format-pages pages))
+                   "")
+                 (if isbn
+                     (concatenate 'string ", ISBN: " (strip-brackets isbn))
+                   "")
+                 (if doi
+                     (concatenate 'string ", DOI: " (strip-brackets doi))
+                   "")
+                 ". [" (format-entry-type entry-type) "]"
+                 "</li>")))
```
gjurgensen: Perhaps it would have been easier to use the XDOC constructors (see `:doc constructors`). But this works :smile:.

books/workshops/references/xdoc-generation.lisp
```diff
@@ -0,0 +1,154 @@
...
+                           "<p>This section contains documentation for papers and publications related to ACL2 workshops and conferences.</p>"
+                           "<p><b>Total number of entries:</b> "(if (natp entry-count)
+                                                                    (coerce (explode-atom entry-count 10) 'string)
+                                                                  "unkown")
+                           "</p>"
+                           "<ul>"
+                           entries-html
+                           "</ul>"))))
+
+(set-state-ok t)
```
gjurgensen: Instead of setting this flag, I would just provide `(declare (xargs :stobjs state))` to the following function where it is needed.

books/workshops/references/bibtex-parser.lisp
```diff
@@ -0,0 +1,399 @@
+; ACL2 BibTeX Parser
+; Parses BibTeX files and creates alist representations
+
+; Aakash Koneru
```
gjurgensen: Can you add copyright+licensing information to the header (and for the other file)? See any of the books in the `kestrel` directory for what that looks like. Copyright Kestrel and license should be the ACL2 top-level license.

gjurgensen:
Right now, these books aren't be included in the top-level books and so won't show up in the manual. Please add an `include-book` of the `xdoc-generation` book to `books/top.lisp` and `books/doc/top.lisp`.

Once you do that, you should be able to build the manual locally with `make manual` (from the `books` directory). Then, open `books/doc/manual/index.html` in a local web browser and see if you can goto your topic to ensure it is included and looks right.

books/workshops/references/bibtex-parser.lisp
akitaki79: As a normal rewrite rule, I kept running into loops and call depth errors when I tried admitting parse-bibtex-key. Changing it to a `:forward-chaining` rule seemed to solve the issue, so I changed it to that.

gjurgensen: Alright, its fine to leave it as-is for now then. Maybe something to circle back to later.

books/workshops/references/xdoc-generation.lisp
```diff
@@ -0,0 +1,159 @@
...
+  (let ((entry-count (len entries))
+        (entries (generate-bibtex-entries-XDOC entries)))
+    `(defxdoc acl2-workshops
+       :parents (pubs-papers)
+       :short "ACL2 Workshop Papers and Publications"
+       :long ,(concatenate 'string
+                           "<p>This section contains documentation for papers and publications related to ACL2 workshops and conferences.</p>"
+                           "<p><b>Total number of entries:</b> "(if (natp entry-count)
+                                                                    (coerce (explode-atom entry-count 10) 'string)
+                                                                  "unkown")
```
gjurgensen: Typo.

books/workshops/references/xdoc-generation.lisp
```diff
@@ -0,0 +1,159 @@
...
+                       (generate-bibtex-entries-XDOC (cdr entries)))
+        (generate-bibtex-entries-XDOC (cdr entries))))))
+
+
+(defun generate-workshop-documentation-xdoc (entries)
+  "Generate single XDOC documentation for all workshop entries"
+  (let ((entry-count (len entries))
+        (entries (generate-bibtex-entries-XDOC entries)))
+    `(defxdoc acl2-workshops
+       :parents (pubs-papers)
```
gjurgensen: This should actually be `pubs::pubs-papers` because the `pubs-papers` topic is in the `pubs` package. After you make this change, cert.pl will need help certifying the book because it will need to know about the `pubs` package. You'll need to create a `xdoc-generation.acl2` file, which `cert.pl` will automatically run before certifying `xdoc-generation.lisp`. The contents of the file just need to be `(ld "../../doc/publications.acl2")`. This will load the file where the `pubs` package is defined.

gjurgensen:
It is actually probably best to create a new directory for these books. Can you create a directory `books/kestrel/bibtex` and move these files there? You'll have to change the `top` and `doc/top` includes accordingly.

books/doc/top.lisp
```diff
@@ -70,6 +70,8 @@
   (include-book "practices")
   (include-book "publications")
 
+  (include-book "kestrel/bibtex/xdoc-generation" :dir :system)
```
gjurgensen: Now that these files are in `books/kestrel`, they should actually be included in `books/kestrel/top.lisp` and `books/kestrel/top-doc.lisp` instead of `books/top.lisp` and `books/doc/top.lisp`. (Those kestrel top books are already included in the other top books.)

gjurgensen (approved):

ericwhitmansmith (approved):

PR #1802, opened 2025-07-17
Continue improving installation and related topics

ericwhitmansmith:
See individual commit messages for rationale for each change.

GitHub's diff for this PR is very confused.  It tries to diff obtaining-acl2 (which was removed entirely) with obtaining-common-listp (which is the new name of installation-requirements and had to be put, alphabetically, in the same spot from which obtaining-acl2 was removed.)

External references, if any, to a few topics need to be fixed up (e.g., in the release instructions):
installation-requirements -> obtaining-common-lisp
installation-summary -> installation-instructions
obtaining-acl2 -> installation


MattKaufmann (approved):
Looks great!  When I moved the installation instructions from the web to xdoc, I avoided making many changes but I hoped that the community would then improve them.  This is exactly what I hoped for, and I don't have any changes to suggest after comparing word-for-word using compare-windows in Emacs.  Some of the changes seemed to fix issues I'd kind of noticed, but some -- such as removing :DOC obtaining-acl2 (moving the useful parts to a better place, :DOC windows-installation) -- hadn't occurred to me.  Thanks, Eric!

MattKaufmann:
@gjurgensen I think such a topic could be useful.  If it's referenced in the installation instructions, it would be good to emphasize that these are definitely optional, as they are typically relevant only for certain experienced ACL2 users and they can be obtained later when they become relevant.

books/system/doc/acl2-doc.lisp
```diff
@@ -109677,59 +109573,86 @@ arithmetic) for libraries of @(see books) for arithmetic reasoning.</p>")
...
-  :short "Alternate instructions for obtaining the ACL2 sources and community books"
-  :long "<p>This topic may be avoided by reading the installation summary: see
-  @(see installation-summary), which is intended to be self-contained.</p>
+(defxdoc obtaining-common-lisp
+  :parents (installation)
+  :short "Obtaining Common Lisp"
+  :long "<p>A requirement for installing ACL2 is to install a suitable
+ Common Lisp implementation.  ACL2 can be hosted by several Common Lisp
+ implementations, as listed alphabetically below.  The most commonly-used of
+ these are SBCL and CCL, which can be obtained without charge.</p>
```
gjurgensen, line 109582: Perhaps it should be mentioned here that 64-bit versions should be preferred over 32-bit versions of any of the following Common Lisps (for performance and to certify certain community books; see the Zulip discussion). This is mentioned in `:doc ccl-installation`, but at a glance I don't see it mentioned explicitly anywhere else.

ericwhitmansmith: Good idea.  I haven't gone over that page yet but will try to include something like that when I do.

PR #1804, opened 2025-07-17
Doc improvements

ericwhitmansmith:
See commit messages for rationale.

### DIFF
--- a/books/build/doc.lisp
+++ b/books/build/doc.lisp
@@ -1045,7 +1045,7 @@ corresponding update.</p>")
 
 
 (defxdoc using-extended-acl2-images ; Step 7
-  :parents (cert.pl ACL2::building-acl2 ACL2::books-certification)
+  :parents (cert.pl ACL2::creating-executable ACL2::books-certification)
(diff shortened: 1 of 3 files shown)


MattKaufmann (approved):
In :DOC installation-instructions, I see that

```
  make LISP=<path_to_your_lisp_executable>
```

was replaced by:

```
  make LISP=<your_command_to_run_lisp>
```

But this isn't quite right, because aliases are commands that won't work here.  For example, I tried the following and got an error, even though `make LISP=sbcl` doesn't have this problem (note that `sbcl` is on my `PATH`).

```
  alias my-lisp=sbcl
  make LISP=my-lisp
```

Perhaps just add that `<your_command_to_run_lisp>` should not be an alias.

MattKaufmann:
Sorry, I meant to add that I think these are great changes!

acoglio (approved):

ericwhitmansmith:
> Perhaps just add that `<your_command_to_run_lisp>` should not be an alias.

Done, thanks

PR #1805, opened 2025-07-20
[doc] Add note on 32-bit Lisps

gjurgensen:
This adds a note recommending 64-bit Lisps under the `obtaining-common-lisp` topic. It also removes a similar note under the `ccl-installation-extra` subtopic, which is covered by the aforementioned addition.


MattKaufmann (commented):
Looks good; here are some suggested small changes.

The change in the test for 64-bit Lisps is an improvement, since it's not tied to x86.  However, GCL 2.6.14 fails the new test:

```
>#+(or 64BIT 64-BIT 64-BIT-TARGET) 'yes #-(or 64BIT 64-BIT 64-BIT-TARGET) 'no

NO

>
```

But here's a test for 64-bit that I believe works in all Lisps.
```
(>= most-positive-fixnum (1- (expt 2 60)))
```

In fact, that's essentially the test that ACL2 actually uses; search for smallp in source file acl2-init.lisp.  That code uses environment variable ACL2_SMALL_FIXNUMS to allow 32-bit Lisps, but I don't think you need to mention anything about that.

Regarding:

```
 ... using a 32-bit Lisp will likely result in reduced
 performance as well as certification failures for certain
 @(see community-books).
```

I think the problematic books are excluded for CMUCL, and I suppose that some (many?) books might actually certify more quickly for 32-bit implementations (a wild guess, to be sure); so maybe it would be good to replace "will likely" by "may".

Finally, regarding "CMUCL does not offer a 64-bit version.": I'd add "as of this writing (July, 2025)".

MattKaufmann (approved):
I meant to click the "approve" box with my comments (I trust Grant to make suitable changes).  My bad, trying to hit "return" while making an edit.  I foolishly thought the "Submit review" button was for that....

gjurgensen:
Thanks Matt. I wasn't very satisfied with my 64-bit test (it seems there is no real standardization of `*features*`). I'm much happier with your test.

MattKaufmann (approved):
Looks great -- thanks!

PR #1806, opened 2025-07-23
Add NDSU x86 flag tests for ADD/SUB/OR/AND (macho64/pe64)

mayuf413:
Files Added
books/kestrel/axe/x86/tests/ndsu/add-all-configurations.{c,macho64,pe64}
books/kestrel/axe/x86/tests/ndsu/sub-all-configurations.{c,macho64,pe64}
books/kestrel/axe/x86/tests/ndsu/or.{c,macho64,pe64}
books/kestrel/axe/x86/tests/ndsu/and.{c,macho64,pe64}
books/kestrel/axe/x86/tests/ndsu/run-tests.lisp (updated to include the new tests)
books/kestrel/axe/x86/tests/ndsu/README.md (updated instructions)

Local Certification 
`books/kestrel/axe/x86/tests/ndsu/run-tests.lisp` certifies locally. 
Pe64 binaries caused `Unsupported executable type: :Pe64`, so I commented them out.

These tests verify processor flag behavior across carry flag (CF), zero flag (ZF), auxiliary flag (AF), overflow flag (OF), parity flag (PF), and sign flag (SF) operations.


yaso9:
FYI @ericwhitmansmith @mayuf413 I think `testgen` (the automated test generator that uses Intel's XED database I wrote for Sol's asmtest) may generate tests for most if not all of these instructions. I've found many issues in flag computations using it. I think there are also some instructions that are currently failing `testgen` generated tests that I never got a chance to look at.

It does have one limitation for checking flag computations though: While it is aware of flags that are always undefined (and it masks those off before checking the results) it is not aware of flags that are _sometimes undefined_ depending on the input values (since XED doesn't have that information). I don't think the instructions tested in this PR have that functionality, and I don't think it's particularly common behavior among x86 instructions.

The _sometimes undefined_ issue could be fairly easily fixed by having the tests call a function that takes the instruction inputs as input and computes a mask to be used to mask off the undefined bits. We could have default implementations simply return the mask currently used (which is from the information we get from XED) and allow users to overwrite that mask for instructions that have _sometimes undefined_ flags behavior.

Anyways, happy to see more testing for the x86 model. Just thought I'd mention `testgen` and that it may already be writing tests for instructions you're looking at in case you haven't tried it yet. It's helped me find some pretty difficult to track down bugs in Linux boot, and I think augmenting it with additional instruction support (vector support, for example, is either non-existent or limited; my memory is hazy) may be a better use of time than writing tests for individual instructions.

PR #1810, opened 2025-07-26
Xdoc entities

ericwhitmansmith:
This simplifies and clarifies the XML entity machinery used by xdoc, to support more easily adding additional entities, such as accented letters in names of paper authors (in future commits).

Now there are fewer things that must be kept in sync (4 things instead of 6).

I confirmed that the generated alist constants are the same (except one is reversed, like the others, because it is now generated, like the others -- and it has no duplicate keys).

I confirmed that the big string constant, *xml-entity-stuff*, is the same except for whitespace, which is now more regular since that string is now generated.

I checked the web manual topic "entities" to ensure that things still rendered correctly.  I also checked the output of :doc entities at the ACL2 prompt.

### DIFF
--- a/books/xdoc/fancy/xslt.js
+++ b/books/xdoc/fancy/xslt.js
@@ -35,9 +35,8 @@
 // Compatibility wrapper for lack of XSLT standards in browsers.
 //
 
-// Warning: Keep this in sync with *entity-strings* and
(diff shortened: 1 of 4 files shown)


solswords (approved):
I don't know much about this, tbh, but it seems good to centralize the source of truth.

PR #1811, opened 2025-07-28
Improved Jump To feature

akitaki79:
I re-indented both files which is polluting the diff and making the actual changes hard to see. To fix this, please use the hide-whitespace functionality in Github.

### DIFF
--- a/books/xdoc/fancy/xdoc.js
+++ b/books/xdoc/fancy/xdoc.js
@@ -1096,9 +1096,46 @@ function jumpInit() {
             uid: xindexObj.topicUid(key)
         });
     }
-ta_data.sort(function(a, b) {
-    return a.nicename.localeCompare(b.nicename);
-});
+    // Defer this sorting until after the page load.
+    setTimeout(() =>
+        // Sort topics. Topics from the ACL2 sources come first. After
+        // that, alphabetize (by topic name, then by value).
+        ta_data.sort(function(a, b) {
+            // Prioritize ACL2 sources
+            const sysA = xdataObj.topicFrom(a.value) === 'ACL2 Sources';
+            const sysB = xdataObj.topicFrom(b.value) === 'ACL2 Sources';
+            if (sysA && !sysB) {
+                return -1;
+            }
+            if (!sysA && sysB) {
+                return 1;
+            }
+            const compareNice = a.nicename.localeCompare(b.nicename);
+            if (compareNice !== 0) {
+                return compareNice;
+            }
+            return a.value.localeCompare(b.value);
+        }), 0);
+
+    // Take the first "count" number of elements from the "flattened"
+    // (concatenated) arrays (but don't actually concatenate them all,
+    // since they may be much larger than the count).
+    function takeFlatten(count, arrays) {
+        const result = [];
+        let taken = 0;
+
+        for (const array of arrays) {
+            for (const item of array) {
+                if (count <= taken) break;
+                result.push(item);
+                taken++;
+            }
+            if (count <= taken) break;
+        }
+
+        return result;
+    }
+
     // Custom substring matcher for infix matching
     // Ranking system:
     // - Rank 0: Exact matches (topic name matches query)
@@ -1106,43 +1143,27 @@ ta_data.sort(function(a, b) {
     // - Rank 2: Substring matches (topic name contains query)
     // Within each rank, topics are sorted by:
     // 1. ACL2 system documentation topics first
-    // 2. Alphabetical order by topic name
+    // 2. Alphabetical order (by topic name then by value)
     function substringMatcher(data) {
         return function findMatches(q, cb) {
-            var rank0and1 = [];
-            var rank2 = [];
-            var qlc = q.toLowerCase();
+            var ranks = [[], [], []];
+            const qlc = q.toLowerCase();
             for (var i = 0; i < data.length; i++) {
-                var name = data[i].nicename.toLowerCase();
-                // var rank = null;
-                var index = name.indexOf(qlc);
+                const name = data[i].nicename.toLowerCase();
+                const index = name.indexOf(qlc);
                 if (index === 0) { // prefix match
                     if (name === qlc) { // exact match
-                        rank0and1.push({ ...data[i], _rank: 0 });
+                        ranks[0].push(data[i]);
                     }
                     else { // proper prefix match
-                        rank0and1.push({ ...data[i], _rank: 1 });
+                        ranks[1].push(data[i]);
                     }
                 } else if (index !== -1) { // substring match
-                    rank2.push({ ...data[i], _rank: 2 });
+                    ranks[2].push(data[i]);
                 }
             }
-            var matches = rank0and1;
-            if (matches.length < 20) {
-                matches.push(...rank2);
-            }
-            // Sort by rank, then ACL2 Sources, then nicename
-            matches.sort(function(a, b) {
-                if (a._rank !== b._rank) return a._rank - b._rank;
-
-                //system directory priority
-                var sysA = xdataObj.topicFrom(a.value) === 'ACL2 Sources';
-                var sysB = xdataObj.topicFrom(b.value) === 'ACL2 Sources';
-                if (sysA && !sysB) return -1;
-                if (!sysA && sysB) return 1;
-                return 0;
-            });
-            cb(matches.slice(0, 20)); // Limit to 20 suggestions
+            // Limit to 20 suggestions
+            cb(takeFlatten(20, ranks));
         };
     }
 


akitaki79:
I re-indented both files which is polluting the diff and making the actual changes hard to see. To fix this, please use the hide-whitespace functionality in Github.

books/xdoc/fancy/xdoc.js
```diff
@@ -1051,184 +1051,131 @@ document.addEventListener("DOMContentLoaded", () => {
...
     }
     ret += "<br/><tt>" + key + "</tt></p>";
     return ret;
 }
 
 function jumpInit() {
-
     var ta_data = [];
-    var keys = xindexObj.allKeys();
+    var keys = Array.from(xindexObj.allKeys());
```
gjurgensen: Is the `Array.from` necessary? I think `allKeys` returns an iterator. If it works without `Arrays.from`, probably best not to change it.

books/xdoc/fancy/xdoc.js
```diff
@@ -1051,184 +1051,131 @@ document.addEventListener("DOMContentLoaded", () => {
...
-
-	var e = jQuery.Event("keydown");
-	e.keyCode = e.which = 9; // 9 == tab
-	$("#jump").trigger(e);
-
-	// We seem to never get here EXCEPT in the case where the user
-	// has typed in the entire text for one of the entries.  In
-	// that case, for whatever reason, the autocomplete feature
-	// doesn't actually trigger the submit.  So, if we get here,
-	// figure out what we're looking at and submit it manually.
```
gjurgensen: I would keep these comments.

books/xdoc/fancy/xdoc.js
```diff
@@ -1051,184 +1051,131 @@ document.addEventListener("DOMContentLoaded", () => {
...
+            // Sort by rank, then ACL2 Sources, then nicename
+            matches.sort(function(a, b) {
+                if (a._rank !== b._rank) return a._rank - b._rank;
+
+                // ACL 2 Package Priority
+                //var pkgA = xdataObj.topicBasepkg(a.value);
+                //var pkgB = xdataObj.topicBasepkg(b.value);
+                //if (pkgA === 'ACL2' && pkgB !== 'ACL2') return -1;
+                //if (pkgA !== 'ACL2' && pkgB === 'ACL2') return 1;
+
```
gjurgensen: I'd remove this commented-out code (unless there is a reason I'm not seeing to keep it in).

books/xdoc/fancy/xdoc.js
```diff
@@ -1051,184 +1051,131 @@ document.addEventListener("DOMContentLoaded", () => {
...
-			     name: "topics",
-			     displayKey: 'nicename',
-			     source: engine1.ttAdapter(),
-			     templates:
-			     {
-				 suggestion: jumpRender
-			     }
-			 });
+    // Custom substring matcher for infix matching with ranking and ACL2 Sources priority
+    function substringMatcher(data) {
```
gjurgensen, line 1147: It would be nice if there was a comment here explaining the rank/sorting system.

books/xdoc/fancy/xdoc.js
```diff
@@ -1051,184 +1051,131 @@ document.addEventListener("DOMContentLoaded", () => {
...
+        return function findMatches(q, cb) {
+            var matches = [];
+            var qlc = q.toLowerCase();
+            for (var i = 0; i < data.length; i++) {
+                var name = data[i].nicename.toLowerCase();
+                var value = data[i].value.toLowerCase();
```
gjurgensen: I don't think we want to look at `value` at all. As an example of what these fields look like, the topic with the nicename `"append$+"` has the value `"acl2____append_42_b2"`. We wouldn't want to include `"append$+"` when searching `"b2"`.

gjurgensen (changes requested):

gjurgensen:
Can you remove `index.html` from the PR? I believe the only changes are whitespace.

books/xdoc/fancy/xdoc.js
```diff
@@ -1051,142 +1051,108 @@ document.addEventListener("DOMContentLoaded", () => {
...
-	queryTokenizer: Bloodhound.tokenizers.whitespace
-    });
+        ta_data.push({
+            value: key,
+            nicename: xindexObj.topicName(key),
+            tokens: [xindexObj.topicRawname(key)],
+            nicelow: xindexObj.topicName(key).toLowerCase(),
+            uid: xindexObj.topicUid(key)
+        });
+    }
```
gjurgensen, line 1138:
```suggestion
ta_data.sort(function(a, b) {
    return a.nicename.localeCompare(b.nicename);
});
```

books/xdoc/fancy/xdoc.js
```diff
@@ -1051,142 +1051,108 @@ document.addEventListener("DOMContentLoaded", () => {
...
+
+                //system directory priority
+                var sysA = xdataObj.topicFrom(a.value) === 'ACL2 Sources';
+                var sysB = xdataObj.topicFrom(b.value) === 'ACL2 Sources';
+                if (sysA && !sysB) return -1;
+                if (!sysA && sysB) return 1;
+                return a.nicename.localeCompare(b.nicename);
+            });
+            cb(matches.slice(0, 20)); // Limit to 20 suggestions
+        };
```
gjurgensen:
```suggestion
        return function findMatches(q, cb) {
            var rank0and1 = [];
            var rank2 = [];
            var qlc = q.toLowerCase();
            for (var i = 0; i < data.length; i++) {
                var name = data[i].nicename.toLowerCase();
                // var rank = null;
                var index = name.indexOf(qlc);
                if (index === 0) { // prefix match
                    if (name === qlc) { // exact match
                        rank0and1.push({ ...data[i], _rank: 0 });
                    }
                    else { // proper prefix match
                        rank0and1.push({ ...data[i], _rank: 1 });
                    }
                } else if (index !== -1) { // substring match
                    rank2.push({ ...data[i], _rank: 2 });
                }
            }
            var matches = rank0and1;
            if (matches.length < 20) {
                matches.push(...rank2);
            }
            // Sort by rank, then ACL2 Sources, then nicename
            matches.sort(function(a, b) {
                if (a._rank !== b._rank) return a._rank - b._rank;

                //system directory priority
                var sysA = xdataObj.topicFrom(a.value) === 'ACL2 Sources';
                var sysB = xdataObj.topicFrom(b.value) === 'ACL2 Sources';
                if (sysA && !sysB) return -1;
                if (!sysA && sysB) return 1;
                return 0;
            });
            cb(matches.slice(0, 20)); // Limit to 20 suggestions
        };
```

gjurgensen (commented):
I noticed that the jump to bar pauses just a bit when you type a short string. For instance, if you just type the single character "-". I believe the issue is that the number of matches is very high, so the call to "sort" is slow. I commented a couple of suggestions which should improve this.

The first suggestion pre-sorts the topics during initialization. Then we don't bother with alphabetical sorting later on because we know items are already alphabetized and the JS's `sort` is stable.

The second tweaks `findMatches` to avoid adding rank 2 results unless we have less than 20 rank 0/1 results.

gjurgensen (approved):
LGTM. FYI, I deployed this to https://acl2.org/docs/test if anyone wants to try the change out.

For instance, you can try typing "manual". You'll see many auto-completes based on substring match, as opposed to what you'll see on the current manual, which is no auto-completes (because there are no prefix matches).

gjurgensen:
@solswords would you like to review? (If you don't want to look through the javascript, you could take a look at the preview https://acl2.org/docs/test and check that the new behavior of the "jump to" is desirable.)

books/xdoc/fancy/xdoc.js
```diff
@@ -1051,142 +1051,137 @@ document.addEventListener("DOMContentLoaded", () => {
...
+        ta_data.push({
+            value: key,
+            nicename: xindexObj.topicName(key),
+            tokens: [xindexObj.topicRawname(key)],
+            nicelow: xindexObj.topicName(key).toLowerCase(),
+            uid: xindexObj.topicUid(key)
+        });
+    }
+    // Defer this sorting until after the page load.
+    setTimeout(() =>
```
mister-walter, line 1100: Long-term, if this takes more than say 1s on an "average" machine we might want to hand this work off to a Web Worker since that will not block the main thread.

mister-walter: On my M4 Macbook Air, this takes ~40ms in Chrome.

mister-walter (approved):
Looks great! Thanks much for this work!

solswords:
I tried it and it seems like a definite improvement. Thanks!

PR #1817, opened 2025-08-02
Remove unnecessary (ttags :all) from run-tests-sub.lisp

mayuf413:
Removed ; cert_param: (ttags :all) from this test file. After further review, the book and its dependencies do not use any trust tags, so this parameter is not required.


ericwhitmansmith:
Actually, this book does depend on books that use trust tags, but this is handled by the :TTAGS :ALL in run-tests-sub.acl2:

; cert-flags: ? T :SKIP-PROOFS-OKP T :TTAGS :ALL

As far as I know, ttags is not a valid cert_param.  Note that "cert flags" and "cert params" are different.

Anyway, I will merge this now.

PR #1818, opened 2025-08-03
[oslib] Introduce `get-decoded-time$` and refactor `date`

gjurgensen:
The `oslib::date` function previously called the Common Lisp `get-decoded-time` function in its raw Lisp definition. This change adds an `oslib::get-decoded-time$` function as a wrapper around this Common Lisp function and modifies `oslib::date` to call the ACL2 `get-decoded-time$` function. This minimizes the amount of work done in raw Lisp and exposes more information about the decoded time to ACL2.

Note that the return theorems for `oslib::get-decoded-time$` are rather strong, but I believe they are justified by the spec. Relevant links:
- https://www.lispworks.com/documentation/HyperSpec/Body/f_get_un.htm#get-decoded-time
- https://www.lispworks.com/documentation/HyperSpec/Body/25_ada.htm
- https://www.lispworks.com/documentation/HyperSpec/Body/26_glo_t.htm#time_zone

The return theorems are based on the understanding that, as long as an error hasn't been signaled, the values returned form a valid decoded time, although perhaps not necessarily the correct time.

`oslib::date` was extended with a `:zone` keyword argument. By default, it is `nil` and the behavior is the same as before. If instead a non-`nil` value is provided, the returned date string will include time zone information (an abbreviation if one is known, or a UTC offset otherwise).

The combined manual `top` topic is also modified to use `:zone t` in its call to `date`.

### DIFF
--- a/books/oslib/date-logic.lisp
+++ b/books/oslib/date-logic.lisp
@@ -46,7 +46,7 @@
                (hour natp :rule-classes :type-prescription)
                (date posp :rule-classes :type-prescription)
                (month posp :rule-classes :type-prescription)
-               (year natp :rule-classes :type-prescription)
+               (year integerp :rule-classes :type-prescription)
                (day natp :rule-classes :type-prescription)
                daylight-p
                (zone rationalp :rule-classes :type-prescription)
@@ -65,7 +65,7 @@ function.  In the logic, this is modeled as a read from the ACL2 oracle.</p>
     represents the day of the month.</li>
   <li>@('month') &mdash; A positive natural number less than @('13').  January
     is @('1'), February is @('2'), etc.</li>
-  <li>@('year') &mdash; A natural number representing the year A.D.</li>
+  <li>@('year') &mdash; An integer number representing the year A.D.</li>
   <li>@('day') &mdash; A natural number less than @('7').  This represents the
     day of the week.  Monday is @('0'), Tuesday is @('1'), etc.</li>
   <li>@('daylight-p') &mdash; A generalized boolean indicating whether daylight
@@ -91,7 +91,7 @@ the current time.</p>"
              (integer-range-p 0 24 (third val))
              (integer-range-p 1 32 (fourth val))
              (integer-range-p 1 13 (fifth val))
-             (natp (sixth val))
+             (integerp (sixth val))
              (integer-range-p 0 7 (seventh val))
              (rationalp (ninth val))
              (<= -24 (ninth val))
@@ -293,7 +293,7 @@ the current time.</p>"
     (let ((month (nth (- month 1) '("January" "February" "March" "April" "May"
                                     "June" "July" "August" "September" "October"
                                     "November" "December")))
-          (year   (str::natstr year))
+          (year   (str::natstr (if (<= 0 year) year 0)))
           (date   (str::natstr date))
           (hour   (if (< hour 10)
                       (str::cat "0" (str::natstr hour))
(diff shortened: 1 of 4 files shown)


books/oslib/date-logic.lisp
```diff
@@ -27,31 +27,125 @@
...
+(include-book "std/strings/cat" :dir :system)
+(include-book "std/strings/decimal" :dir :system)
 (include-book "std/util/define" :dir :system)
+(local (include-book "std/typed-lists/string-listp" :dir :system))
 (local (xdoc::set-default-parents oslib))
 
-(define date (&optional (state 'state))
-  :returns (mv (datestamp stringp :rule-classes :type-prescription)
+;; These return theorems are justified by the specification for
+;; - get-decode-time: https://www.lispworks.com/documentation/HyperSpec/Body/f_get_un.htm#get-decoded-time
```
ericwhitmansmith:
```suggestion
;; - get-decoded-time: https://www.lispworks.com/documentation/HyperSpec/Body/f_get_un.htm#get-decoded-time
```

books/oslib/date-logic.lisp
```diff
@@ -27,31 +27,125 @@
...
-Common Lisp system.</p>
+  :long "<p>This is a wrapper to the standard @('get-decoded-time') Common Lisp
+function.  In the logic, this is modeled as a read from the ACL2 oracle.</p>
+
+<p>There are nine return values in addition to @('state'):</p>
+<ul>
+  <li>@('second') &mdash; A natural number less than @('60').</li>
+  <li>@('minute') &mdash; A natural number less than @('60').</li>
+  <li>@('hour') &mdash; A natural number less than @('24').</li>
+  <li>@('date') &mdash; A positive natural number less than @('32').  This
```
ericwhitmansmith:
```suggestion
  <li>@('date') &mdash; A positive integer less than @('32').  This
```

books/oslib/date-logic.lisp
```diff
@@ -27,31 +27,125 @@
...
+    represents the day of the month.</li>
+  <li>@('month') &mdash; A positive natural number less than @('13').  January
```
ericwhitmansmith:
```suggestion
  <li>@('month') &mdash; A positive integer less than @('13').  January
```

books/oslib/date-logic.lisp
```diff
@@ -27,31 +27,125 @@
...
+            (sixth val)
+            (seventh val)
+            (eighth val)
+            (ninth val)
+            state)
+      (mv 0 0 0 1 1 0 0 nil 0 state)))
+  ///
+
+  (defret get-decoded-time$.second-linear
+    (< second 60)
```
ericwhitmansmith: You might use (<= second 59) here (and similarly elsewhere) for a slightly stronger linear rule.

solswords (approved):
Looks good; I agree that it's nice to do less in raw Lisp.

I haven't checked that the return value range/type assumptions are justified by the CLHS spec but assuming they're right I think it's good to have strong "theorems" about them.

A minor complaint that doesn't have a lot to do with this PR in particular: I feel like including std/strings/cat and std/strings/decimal is overkill just to have convenient string concatenation and natural number to string functions (functionality which, after all, is included in base ACL2). I see this happening a lot and find myself doing it too just because it's convenient. Including these includes 76 total books (for perspective, std/util/define includes 19) -- not that these take long to certify, but it means that any change in any of these books will then require a lot of books to be recertified. I think what we're missing are barebones books for these two purposes that just wrap the built-in ACL2 functions and prove any necessary type theorems. But (to be clear) we don't need to hold up this PR for that.

PR #1819, opened 2025-08-05
Search feature improvements

akitaki79:
This only searches titles/shorts right now. Long searching (with FTS5) may be added to the PR soon.


gjurgensen (approved):

mister-walter:
I gave this a quick look at it looks good! Thanks @akitaki79!

PR #1822, opened 2025-08-07
Add x86 tests for PAND, POR, PXOR

mayuf413:
This pull request adds test cases for the following x86 SIMD instructions:

PAN/D

POR

PXOR


ragerdl:
Hah this looked familiar for a moment.

ericwhitmansmith:
Yep, but these are x86 instructions rather than ACL2(p) constructs.  :-)

PR #1823, opened 2025-08-08
Fix broken links in Publications

akitaki79:


akitaki79:
> Wonderful, and thanks! But it would be good to merge this into testing instead of master.

Oh sorry, I completely forgot to do that. Thanks.

gjurgensen (approved):

PR #1827, opened 2025-08-14
[C$] Extend lexer to handle preprocessing directives

gjurgensen:
This adds preprocessing directives as a new lexeme. It is lexed very similarly to line comments, except that when we reach a `#` character, we also look backward to check that any preceding characters on the line are whitespace [C17:6.10/2]. This condition is checked by the new `only-whitespace-backward-through-line` function. This function does not "unread" characters, it simply looks back through the (read portion of the) chars array.

### DIFF
--- a/books/kestrel/c/syntax/parser.lisp
+++ b/books/kestrel/c/syntax/parser.lisp
@@ -4804,7 +4804,11 @@
    (xdoc::p
     "We begin with the character immediately before the last read character,
      and check that every character is whitespace until we reach either
-     a new-line or the start of the file."))
+     a new-line or the start of the file.")
+   (xdoc::p
+    "Since @(tsee read-char) converts all recognized new-line sequences
+     into a single line feed character,
+     we detect new-lines by simply checking for a line feed."))
   (b* ((chars-read (parstate->chars-read parstate))
        ((when (= chars-read 1))
         t)
@@ -4825,8 +4829,7 @@
      :returns (all-whitespace booleanp)
      (b* ((char+pos (parstate->char i parstate))
           (char (char+position->char char+pos)))
-       (cond ((or (= char 10)  ; LF new-line
-                  (= char 13)) ; CR new-line
+       (cond ((= char 10) ; new-line
               t)
              ((or (= char 32) ; SP
                   (and (<= 9 char) (<= char 12))) ; HT VT FF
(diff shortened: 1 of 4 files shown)


books/kestrel/c/syntax/parser.lisp
```diff
@@ -4712,6 +4713,136 @@
...
+                  (new-parstate parstatep :hyp (parstatep parstate)))
+     :parents nil
+     (b* (((reterr) (irr-position) parstate)
+          ((erp char pos parstate) (read-char parstate)))
+       (cond
+        ((not char) ; EOF
+         (reterr-msg :where (position-to-msg pos)
+                     :expected "a character"
+                     :found (char-to-msg char)
+                     ;; TODO GJ: update
```
acoglio: Is this TODO done?

books/kestrel/c/syntax/parser.lisp
```diff
@@ -4712,6 +4713,136 @@
...
+  ((define only-whitespace-backward-through-line-loop ((i natp)
+                                                       (parstate parstatep))
+     :parents nil
+     :guard (< i (parstate->chars-length parstate))
+     :returns (all-whitespace booleanp)
+     (b* ((char+pos (parstate->char i parstate))
+          (char (char+position->char char+pos)))
+       ;; TODO: need to support other newlines?
+       (cond ((or (= char 10)  ; LF new-line
+                  (= char 13)) ; CR new-line
```
acoglio: I think that `read-char` never produces a CR (13), so we could skip this check if that's the case.

books/kestrel/c/syntax/parser.lisp
```diff
@@ -4712,6 +4713,136 @@
...
+  :prepwork
+
+  ((define only-whitespace-backward-through-line-loop ((i natp)
+                                                       (parstate parstatep))
+     :parents nil
+     :guard (< i (parstate->chars-length parstate))
+     :returns (all-whitespace booleanp)
+     (b* ((char+pos (parstate->char i parstate))
+          (char (char+position->char char+pos)))
+       ;; TODO: need to support other newlines?
```
acoglio: We only recognize CR and/or LF as new-line currently, so I'd say that we don't need to worry about this until/if we see some code with some other kind of new-line (Unicode).

acoglio (approved):

books/kestrel/c/syntax/parser.lisp
gjurgensen: Oh yes, I meant to remove those "TODO" comments earlier. Fixed now.

books/kestrel/c/syntax/parser.lisp
gjurgensen: Good catch. That is indeed the case. I dropped the CR check and added a note to that function's xdoc long.

PR #1831, opened 2025-08-26
Implement server-side seaching for the web manual

gjurgensen:
This PR improves the web manual search by using sqlite FTS5 to quickly search topics/shorts/longs on the server.

Note that the `desktop.css` and `mobile.css` files are generated from the SCSS files. It looks like there are more changes than there actually are because the SCSS->CSS generator I used didn't match the style of the originals.

The sqlite database has been modified to include more data (topic names and shorts). It also builds an FTS table to allow for FTS5 searches. This increases the size of the database from 310M to 430M.

Search results are annotated with their "score", which is used to rank them. The calculation is fairly ad-hoc. It starts with the FTS BM25 score, but then we adjust it to overweight topics with exact results in the topic names and such.

You can preview these changes at: https://acl2.org/docs/test

### DIFF
--- a/books/xdoc/fancy/xdoc.js
+++ b/books/xdoc/fancy/xdoc.js
@@ -947,8 +947,16 @@ function searchAddHit(matches, hits, key) {
 }
 
 function searchGoServer(query_str) {
-    const url = XDATAGET + "?search=" + encodeURIComponent(query_str);
+    // Should these first couple of things be factored out? Same between Server and Local
+    $("#searching_message").hide();
+    if (query_str.length === 0) {
+        $("#data").append("<h3>No results (empty search)</h3>");
+        return;
+    }
 
+    $("#data").append("<h1><u>" + htmlEncode(query_str) + "</u></h1>");
+
+    const url = XDATAGET + "?search=" + encodeURIComponent(query_str);
     fetch(url, {
         method: 'GET',
     }).then(res => res.json()).then(obj => {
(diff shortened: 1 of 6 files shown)


gjurgensen:
I'll double check that nothing weird is happening with colons. I'm guessing that FTS5 is just parsing `:expand` as two tokens.

You can also try putting phrases in double quotes (e.g. searching `":expand" hint`) to search part of the query literally.

books/xdoc/fancy/xdoc.js
```diff
@@ -913,58 +928,125 @@ function searchGo(str) {
...
+        return results;
+    }
+}
+
+async function serverSupportedSearch(query_str) {
+    const max_display = 100;
+    const max_results = 1000;
+
+    const url = XDATAGET + "?search=" + encodeURIComponent(query_str);
+    const [results, [client_results, _, query_tokenized]] = await Promise.all([
```
mister-walter, line 999: This was a little unclear to me the first time I read through it so it might benefit from a comment saying what's going on: as I understand it, both a server-side search as well as a local search are done simultaneously. Once both are complete, the results of the local search are used to modify the weighting (e.g. the order in which the results are shown) of the server-side search results.

A bit of discussion on the way that the weights are updated might also be nice, but those details are less critical than the overall picture of what's going on.

books/xdoc/fancy/xdoc.js
```diff
@@ -1008,73 +1105,103 @@ function searchGoMain(query_str) {
...
+        $("#data").append("<h3>No results</h3>");
+        return;
+    }
+
+    // Sort results by rank, then ACL2 Sources priority, then frequency, then alphabetical
+    results_array = results_array.sort(function(a, b) {
+        if (a.rank !== b.rank) return a.rank - b.rank;
+
+        // ACL2 Sources priority
+        // Note: on the server-supported flavor of the manual, topicFrom may
```
mister-walter: Is this comment relevant given this code is only run in `searchGoLocal`? Or maybe I'm misunderstanding what it's saying?

mister-walter (approved):
Overall looks great, just a few nits regarding comments.

books/xdoc/fancy/xdoc.js
gjurgensen: Good catch, that comment is indeed no longer relevant. I removed it.

PR #1834, opened 2025-08-29
[C$] Annotate ASTs with UIDs

gjurgensen:
This adds annotation fields and types to `initdeclor`s and `param-declors`, and extends the types of variables and `fundef`s to store a UID.


books/kestrel/c/syntax/validation-information.lisp
```diff
@@ -78,6 +78,55 @@
...
+     to be unique to a given variable, function, type name, etc.
+     E.g., there may be many variables throughout a program
+     with the name @('x'), but all such distinct variables
+     will have distinct unique identifiers.")
+   (xdoc::p
+    "Unique identifiers are assigned during validation
+     to aid subsequent analysis.
+     By annotating identifiers with their unique alias,
+     disambiguation of variables becomes trivial."))
+  ((ident nat))
```
acoglio: Maybe I would call this component `uid` instead of `ident`, since it's not a C identifiers as such.

(It's often awkward to find a good name for a one-component product type. Sometimes I've used `unwrap` but I've been moving away from that.)

acoglio (approved):

PR #1835, opened 2025-08-30
[FTY] Add some list types for FTY aggregates.

acoglio:
The form of these definitions is the one generated by `deflist`.


gjurgensen (approved):
Can these definitions be disabled by default? Otherwise, LGTM.

acoglio:
Yes, they should of course. I just pushed a commit to do that.

PR #1839, opened 2025-09-06
[C$] Add type parameter to pointer types

gjurgensen:
Many theorems of the name `<function>-when-type-kind-syntaxp` are added to `types.lisp`. These are added to functions which are decided solely by the `type-kind` value of the argument. These theorems just expand the function definition when there is a hypothesis `(equal (type-kind type) kind)` where `kind` is a constant.

I introduce `type-composite` to explicitly construct composite types. Currently, this function is not doing very much interesting, just biasing the "merging" toward more specific (i.e. non-unknown) types. The interesting behavior for composite types has to do with array size and function parameter type lists [C17:6.2.7/3], which we are not modelling in the approximate type system.

### DIFF
--- a/books/kestrel/c/syntax/types.lisp
+++ b/books/kestrel/c/syntax/types.lisp
@@ -24,6 +24,8 @@
 (local (acl2::disable-builtin-rewrite-rules-for-defaults))
 (set-induction-depth-limit 0)
 
+(local (include-book "kestrel/utilities/ordinals" :dir :system))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (defxdoc+ types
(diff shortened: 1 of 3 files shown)


books/kestrel/c/syntax/types.lisp
```diff
@@ -39,111 +41,111 @@
...
-  (:struct ((tag? ident-optionp)))
-  (:union ())
-  (:enum ())
-  (:array ())
-  (:pointer ())
-  (:function ())
-  (:unknown ())
-  :pred typep)
+(encapsulate ()
+  (set-induction-depth-limit 1)
```
acoglio: I would use a `:prepwork ((set-induction-depth-limit 1))` inside, at the end of, the `fty::deftagsum`, which seems more readable.

books/kestrel/c/syntax/types.lisp
```diff
@@ -39,111 +41,111 @@
...
+      (xdoc::li
+       "A collective type for all enumeration types [C17:6.2.5/20].
+        This is an approximation,
+        because there are different enumeration types.")
+      (xdoc::li
+       "A collective type for all array types [C17:6.2.5/20].
+        This is an approximation,
+        because there are different array types.")
+      (xdoc::li
+       "A parameterized pointer type [C17:6.2.5/20].
```
acoglio: I would rephrase this slightly to say something like "a pointer type that includes information about the pointed-to type", since "parameterized type" has often a more specific meaning in type theory (although C does not have those kind of parameterized types).

books/kestrel/c/syntax/types.lisp
```diff
@@ -39,111 +41,111 @@
...
+        but if the expression is used elsewhere,
+        we need to accept it, because it could have the right type.
+        We use this unknown type for this purpose:
+        the expression @('s.m') has unknown type,
+        and unknown types are always acceptable."))
+     (xdoc::p
+      "Besides the approximations noted above,
+       currently we do not capture atomic types [C17:6.2.5/20],
+       which we approximate as the underlying (argument) type.
+       Furthermore, we do not capture qualified types [C17:6.2.5/26]."))
```
acoglio: I see that the mention of `typedef` names has been removed. Do we expand them? If so, I think it would be good to mention that here, i.e. that we do not capture them in this fixtype, but we handle them by expanding them.

books/kestrel/c/syntax/types.lisp
```diff
@@ -234,7 +236,17 @@
   :short "Check if a type is a standard signed integer type [C17:6.2.5/4]."
   (and (member-eq (type-kind type) '(:schar :sshort :sint :slong :sllong))
        t)
-  :hooks (:fix))
+  :hooks (:fix)
+
+  ///
+
+  (defrule type-standard-signed-integerp-when-type-kind-syntaxp
```
acoglio, line 243: I wonder if we should disable these by default, given that the LHS is fairly generic? (They could be put in a rule set, to be enabled for the proofs that need this kind of reasoning.)

books/kestrel/c/syntax/types.lisp
```diff
@@ -771,24 +908,71 @@
...
+
+(define type-composite ((x typep) (y typep))
+  :guard (type-compatiblep x y)
+  :returns (composite typep)
+  :short "Construct a composite @(see type) [C17:6.2.7/3]."
+  :long
+  (xdoc::topstring
+   (xdoc::p
+    "In our current approximate type system, the composite type is
+     the type of @('x') if the type of @('y') is unknown,
```
acoglio: Why "the type of x" vs. just "x", since x is a type? (Same for y.)

books/kestrel/c/syntax/types.lisp
```diff
@@ -771,24 +908,71 @@
...
+    :induct t)
 
   (defrule type-compatiblep-symmetric
     (equal (type-compatiblep y x)
            (type-compatiblep x y))
+    :induct t))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define type-composite ((x typep) (y typep))
```
acoglio, line 938: Should the composite of unknown and not-unknown be unknown instead? Perhaps it makes no difference in our current type system. In general, are there situations in which, in correct code, unknown stands for an actual type T that, when composited with a not-unknown T', should give something slightly different from T'? Maybe it needs a more refined type system for this to happen?

books/kestrel/c/syntax/types.lisp
gjurgensen: I seem to have forgot that `x` and `y` were types when I wrote that :smile:. I'll replace "type of x/y" with "x/y".

books/kestrel/c/syntax/types.lisp
gjurgensen: Yes I think our type system would need to be further refined for such a case to arise (specifically, we'd need array sizes or full function prototypes). As of now, compositing the types will never change either type other than refining the unknowns. It can effect which of the two types is returned though. For instance, `(type-composite (type-enum) (type-sint))` would currently return `(type-enum)` (in arbitrary choice between the two). But if we abstract the enum, we get `(type-composite (type-unknown) (type-sint))` which evaluates to `(type-sint)`. That is fine though. Both answers are acceptable composite types for the two original types (the only constraint in this case is that the composite is compatible with both input types).

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1151,22 +1152,24 @@
...
-       ((unless (or (and (type-case type1 :pointer)
-                         (type-integerp type2))
-                    (and (type-integerp type1)
-                         (type-case type2 :pointer))))
+       ((when (and (type-case type1 :pointer)
+                   (type-integerp type2)))
+        (retok (type-pointer->to type1)))
+       ((unless (and (type-integerp type1)
+                     (type-case type2 :pointer)))
         (retmsg$ "In the array subscripting expression ~x0, ~
```
acoglio, line 1160: This should be allowed according to [C17:6.5.2.1]: A[B] is the same as B[A].

books/kestrel/c/syntax/types.lisp
gjurgensen: Are you thinking the rules may be undesirable in some contexts, or that the rules may be expensive?

If the former, I think the rules are universally desirable because they will only apply when the LHS can be rewritten to a ground term (which will be boolean).

If the latter, I assumed the rules wouldn't be too expensive. There is no backchaining, just binding a free variable and then checking the `syntaxp` constraint. Unless the free variable binding is more expensive than I imagine?

Regardless, if you think we should disable these rules by default, I can do so and add them to a ruleset like you suggest.

books/kestrel/c/syntax/validator.lisp
gjurgensen: I believe it is still allowed. The condition of the `unless` is the allowed behavior. (I can change it to two `when`s if the asymmetry is confusing.)

books/kestrel/c/syntax/validator.lisp
```diff
@@ -2192,6 +2251,29 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(define make-pointers-to ((pointers typequal/attribspec-list-listp)
```
acoglio: Perhaps we could move this to abstract-syntax-operations.lisp?

acoglio (approved):

books/kestrel/c/syntax/types.lisp
acoglio: I think they are always desirable, because if (:e member-equal) is enabled (as usually is), then the LHS rewrites to t or nil.

I was thinking of the performance aspect, but we already have lots of enabled rules that shouldn't be enabled, so I wouldn't do anything for now.

books/kestrel/c/syntax/types.lisp
acoglio: But if instead we have an enum and and int, but we have abstracted the int to unknown, then would be calculate enum as the composite?

books/kestrel/c/syntax/validator.lisp
acoglio: Yes, I misread, sorry.

books/kestrel/c/syntax/types.lisp
gjurgensen: Yes, `(type-composite (type-enum) (type-unknown))` would yield `(type-enum)`. But that seems fine, no? We consider enums compatible with integers, and the only requirement of composite types outside of arrays and functions is that the constructed type is compatible with both of the two input types.

books/kestrel/c/syntax/validator.lisp
gjurgensen: That would require including `types.lisp` in `abstract-syntax-operations.lisp`, which would create a circular dependency (apparently `types.lisp` indirectly depends on `abstract-syntax-operations.lisp`).

acoglio: Right. Perhaps we might move it to types.lisp then? It just seems a more general operation.

books/kestrel/c/syntax/types.lisp
acoglio: I guess so. Looking at [C17:6.2.7], it's not clear to me that it defines precisely what the composite type of two types is. I think it's fine to leave the code as you have it there.

PR #1840, opened 2025-09-07
[C$] Add element type field to array types

gjurgensen:
The extends array types with an element type, but array types remain an approximation because we do not store the array size.

This also weakens the null pointer check now that we have more descriptive pointer types. Only void/unknown pointer types are allowed, not arbitrary pointer types.

It also extends `type-formalp` and `ldm-type` to apply to pointer types. Arrays are still excluded because they do not store the array size.

### DIFF
--- a/books/kestrel/c/syntax/types.lisp
+++ b/books/kestrel/c/syntax/types.lisp
@@ -86,7 +86,7 @@
       This is an approximation,
       because there are different enumeration types.")
     (xdoc::li
-     "An array types [C17:6.2.5/20],
+     "An array type [C17:6.2.5/20],
       derived from the ``element type.''
       This is an approximation,
       because we do not track the size of the array type.")
--- a/books/kestrel/c/syntax/validator.lisp
+++ b/books/kestrel/c/syntax/validator.lisp
@@ -3617,7 +3617,7 @@
              (initer-case initer :list))
         (b* (((erp new-elems types table)
               (valid-desiniter-list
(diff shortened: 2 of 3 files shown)


books/kestrel/c/syntax/types.lisp
```diff
@@ -86,12 +86,13 @@
       This is an approximation,
       because there are different enumeration types.")
     (xdoc::li
-     "A collective type for all array types [C17:6.2.5/20].
+     "An array types [C17:6.2.5/20],
```
```suggestion
     "An array type [C17:6.2.5/20],
```

books/kestrel/c/syntax/validator.lisp
```diff
@@ -3763,7 +3789,7 @@
                            (designor-fix designor)
                            (type-fix target-type))))
               (retok (make-designor-sub :index new-index :range? new-range?)
-                     (type-unknown)
+                     (type-fix target-type)
```
acoglio: Should this be `(type-array->of target-type)`? Since the output target type is the one resulting from applying the designator to the input target type.

gjurgensen: Ah yes, you are right. I originally misinterpreted  the meaning of that return type.

This addition required some further changes surrounding the treatment of initializers. One part of that was adding a check to `valid-designor-list` that considers whether we have a type which may be an aggregate or union with an element/member type which is also an aggregate or union. This gets quite hairy, so for now, I fall back to the unknown type in such situations.

PR #1841, opened 2025-09-09
Allow arbitrary STOBJ-LET bindings in STOBJ-GET

copperteal:
The old bindings guard in `STOBJ-GET` accepts lists of arbitrary doublets but rejects `STOBJ-LET` bindings with custom updaters.  This patch sets the bindings guard in `STOBJ-GET` to match the description given in "SECTION: Precise documentation for stobj-let" of the nested-stobjs documentation.

### DIFF
--- a/books/std/stobjs/nested-stobjs.lisp
+++ b/books/std/stobjs/nested-stobjs.lisp
@@ -35,7 +35,6 @@
 
 (include-book "std/util/bstar" :dir :system)
 
-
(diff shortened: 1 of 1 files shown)


copperteal:
Sounds good.  I've reduced the patch to a one-liner protecting the guard of `strip-cars` and of the `assoc` in `stobj-get-stobj-return-vars`.

solswords (approved):
Looks good. I can go ahead and merge it unless you're waiting on anyone else?

PR #1842, opened 2025-09-10
[C$] Add return type to function types

gjurgensen:


books/kestrel/c/syntax/validator.lisp
```diff
@@ -1210,10 +1210,9 @@
      the first sub-expression must have pointer type [C17:6.5.2.2/1];
      Furthermore, it must be a pointer to a function type.
      Since we currently have just one function type,
```
acoglio: It looks like the line just above should be deleted.

acoglio (approved):

PR #1843, opened 2025-09-11
[C$] Optimize proofs

gjurgensen:
This locally disables tau in a number of books. Useless runes had to be regenerated as a result of the proof changes.

### DIFF
--- a/books/kestrel/c/syntax/disambiguator.lisp
+++ b/books/kestrel/c/syntax/disambiguator.lisp
@@ -201,12 +201,6 @@
 
   ///
 
-  (defrule alistp-when-dimb-scopep-cheap
-    (implies (dimb-scopep scope)
-             (alistp scope))
-    :rule-classes ((:rewrite :backchain-limit-lst (0)))
-    :induct t)
-
   (defrule dimb-kindp-of-cdr-of-assoc-equal-when-dimb-scopep
     (implies (dimb-scopep scope)
              (iff (dimb-kindp (cdr (assoc-equal ident scope)))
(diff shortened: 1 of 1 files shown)


books/kestrel/c/syntax/disambiguator.lisp
```diff
@@ -199,12 +201,18 @@
 
   ///
 
+  (defrule alistp-when-dimb-scopep-cheap
```
acoglio: I would prefer to disable this and to remove the backchain limit (and the `-cheap` from the name). Rationale:
- Despite the backchain limit, it seems better to even avoid looking for hyps that match the rule hyp (even without backchaining).
- If a proof needs the fact that a dimb-scopep is also an alistp, then let's enable it "like we mean it". Maybe it's only 1 or 2 backchain steps away.

Does defalist not already generate a rewrite rule like this?

books/kestrel/c/syntax/disambiguator.lisp
```diff
@@ -233,7 +241,16 @@
   :true-listp t
   :elementp-of-nil t
   :pred dimb-tablep
-  :prepwork ((local (in-theory (enable nfix)))))
+  :prepwork ((local (in-theory (enable nfix))))
+
+  ///
+
+  (defrule alistp-of-car-when-dimb-tablep-cheap
```
acoglio: Same comment as for the other theorem.

books/kestrel/c/syntax/disambiguator.lisp
gjurgensen: Yes, `defalist` generates such a theorem, just not a "cheap" rule. I removed the cheap rule and so now just enable the `defalist`-generated theorem where necessary (only a couple guard theorems).

PR #1846, opened 2025-09-16
Clarify x86isa model build instructions in context of Linux boot

yaso9:
These doc changes are in response to a discussion with Warren about the build instructions being confusing, especially the description of the `X86ISA_EXEC` option, now that we can run a kernel (and thus, handle syscalls) in the model. These changes clarify the build instructions to hopefully reduce ambiguity about the `X86ISA_EXEC` option. The primary change is the addition of a paragraph at the top of the build instructions topic which states that if you're running a kernel in the model, like Linux, the value of `X86ISA_EXEC` is irrelevant along with a link to the doc topic on `app-view`.

The references to full execution and execution of syscall instructions have been clarified to mention that this only applies in app-view with links to the app-view topic.

Additionally, the paths where one should run `make` or `cert.pl` have also been clarified.

### DIFF
--- a/books/projects/x86isa/doc.lisp
+++ b/books/projects/x86isa/doc.lisp
@@ -231,14 +231,25 @@
   :short "Building books related to the x86 ISA and the machine-code
   analysis framework."
 
-  :long "<p>Some ways of building the @('x86isa') books are:</p>
+  :long "<p><b>Note:</b> If you intend on running a kernel in the x86 model,
+  like Linux, then you will NOT use the @(see app-view), so the value of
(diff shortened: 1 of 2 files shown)


books/projects/x86isa/doc.lisp
```diff
@@ -253,8 +264,9 @@
       X86ISA_EXEC=t \\
       ACL2=/Users/shilpi/acl2/saved_acl2
```
acoglio: Maybe we should generalize this directory to something like `/path/to/saved_acl2`?

yaso9: Changed it to `/path/to/saved_acl2`

PR #1848, opened 2025-09-19
Mention CONST property in DEFCONST docs

copperteal:
Small patch to add mention of the `const` property to the documentation for `defconst`.


MattKaufmann:
Looks good; thanks.  Tiny picky point; please replace @(see getprop) with @(tsee getprop).

copperteal:
You got it.

PR #1849, opened 2025-09-19
[interface] Make elisp function argument optional

gjurgensen:
This addresses an issue observed when attempting to do `M-x customize-variable`. This seemed to be calling indent-sexp with zero arguments, while it is defined in `acl2-indent.el` with 1 argument. Making the argument optional fixes the issue.


westfold:
That's good. Thanks!

PR #1858, opened 2025-09-28
Remove vestigial include from cert

copperteal:
`cowles/portcullis` consists of package definitions for commutative algebraic structures.  It is not used in the osets library and the community books rebuild fine without it.

### DIFF
--- a/books/std/osets/cert.acl2
+++ b/books/std/osets/cert.acl2
@@ -33,5 +33,4 @@
 #!ACL2
 (in-package "ACL2")
 (include-book "std/portcullis" :dir :system)
-(include-book "cowles/portcullis" :dir :system)
 ; cert-flags: ? t :ttags :all


ericwhitmansmith:
This seems like a good change, but I wonder why that include was there in the first place (Chesterton's Fence).  It seems to be from this commit by @solswords :
https://github.com/acl2/acl2/commit/fff7c7af3f49e9e9a3e88c73973e377022599afa

solswords:
If you look at books/build/Makefile-deps at the dependencies of these books
(std/strings/pretty-program, std/strings/defs-program, and xdoc/str are the
ones I know of), there's a pretty extensive list of .lisp files that they
include uncertified in order to build up a set of program-mode definitions.
The idea was that if all you want to do is be able to use the function
upcase-string (and in program mode), but you have to certify a whole bunch
of std/lists books first (because they're included in
std/strings/arithmetic, which is included in std/strings/case-conversion,
which is where upcase-string is defined), that's a bit of a bummer.
Granted, on a fast machine it may only take minutes CPU / a fraction of a
minute with parallelism to certify all those books nowadays. So maybe it's
not worth bothering -- someone could experiment with replacing includes of
std/strings/defs-program with std/strings/defs, similarly for
pretty-program, and replace xdoc/str itself with something that just
included the relevant books, and see if there's any noticeable difference
in anything's critical path.

PR #1862, opened 2025-10-17
[X86ISA] Explain how to measure performance of Linux.

acoglio:
The information comes from Yahya Sohail. I just changed the function name.

### DIFF
--- a/books/projects/x86isa/linux/doc.lisp
+++ b/books/projects/x86isa/linux/doc.lisp
@@ -151,33 +151,53 @@
            model. You can mount it like any other drive on a Linux system.</p>
 
            <p>To measure the performance of booting Linux on the @('x86isa') model
-           (running on your specific machine), add a function of the form</p>
+           (running on your specific machine), add code the form</p>
 
            <code>
            (defun run-model-count-steps (n x86)
-            (declare (xargs :mode :program
-                            :stobjs (x86)))
-            (b* (((when (equal (cpl x86) 3)) (mv n x86))
-                 (x86 (x86-fetch-decode-execute x86)))
-              (run-model-count-steps (1+ n) x86)))
+             (declare (xargs :mode :program
+                             :stobjs (x86)))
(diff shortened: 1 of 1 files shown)


books/projects/x86isa/linux/doc.lisp
```diff
@@ -148,7 +148,31 @@
            @('blkx86isa'). This device is essentially a view into the gigabyte of
            physical memory at address @('0x100000000'). This could be useful for
            transfering files into and out of the Linux system running in the
-           model. You can mount it like any other drive on a Linux system.</p>"))
+           model. You can mount it like any other drive on a Linux system.</p>
+
+           <p>To measure the performance of running Linux on the @('x86isa') model
```
yaso9: Minor nitpick, but I'd say "booting Linux" instead of "running Linux".

books/projects/x86isa/linux/doc.lisp
```diff
@@ -148,7 +148,31 @@
...
+           (running on your specific machine), add a function of the form</p>
+
+           <code>
+           (defun run-model-count-steps (n x86)
```
yaso9, line 157: I'd include a brief statement of what this function does somewhere around here. It runs the model until we enter privilege level 3, which corresponds to user mode. Thus, this function will execute Linux until it drops into user mode to run `init`.

yaso9:
Sorry, didn't see this until I got the notification for Eric's comment above. LGTM other than the couple of comments I left.

books/projects/x86isa/linux/doc.lisp
ericwhitmansmith: > I'd include a brief statement of what this function does somewhere around here. It runs the model until we enter privilege level 3, which corresponds to user mode. Thus, this function will execute Linux until it drops into user mode to run `init`.

That seems like a great idea.  I initially didn't know what the test `(equal (cpl x86) 3))` was doing in the code.

PR #1863, opened 2025-10-23
[C$] Add parameter information to function types

gjurgensen:
This distinguishes between three types of function parameters: (1) function prototypes (the typical, modern syntax), (2) "old style" function identifier lists (excluding the following special case) and (3) the special empty identifier list case occurring outside of definitions. These three kinds of function parameters are distinguished because they are treated differently in terms of type compatibility and composites, and in the type-checking of function call expressions.

This PR extends the validator to assign functions the proper type, and also reject ill-typed function call expressions. The validator currently approximates the types of "old style" function definitions, assigning the unknown type to each parameter. (Refining the typing for this case will require a bit of refactoring, since we can't know the type of the parameters until later, after validating the declarations which following the function declarator.

The `valid-dirdeclor` function was tweaked a bit in order to validate the parameters (in the `:function-params` and `:function-names` cases) before validating the inner declarator. This is necessary to get the types of the parameters to form the proper function type which must be passed to the recursive validation of the inner validator.

As a result of this change, we no longer have the refined `fundef-params-p` in time for the parameter validation. So, we instead peek into the inner declarator (using the new `dirdeclor-has-paramsp`) to check whether the function definition parameters are in the inner declarator. Validation functions no longer return a refined `fundef-params-p` value, since it is no longer used.

This also replaces some events at the tops of books with a call to the new `controlled-configuration`.

### DIFF
--- a/books/kestrel/c/syntax/validator.lisp
+++ b/books/kestrel/c/syntax/validator.lisp
@@ -1189,7 +1189,7 @@
    (xdoc::ol
     (xdoc::li
      "The number of arguments must be the same as the number of parameter types,
-      or greater than equal when an ellipsis is present.")
+      or greater than or equal when an ellipsis is present.")
     (xdoc::li
      "For each parameter/argument pair,
       the same restrictions apply as in the case of simple assignment.
@@ -4300,7 +4300,7 @@
        as the parameters of a function being defined [C17:6.7.6.3/3]
        Thus, we raise an error when the list is nonempty
        and @('fundef-params-p') is @('nil')
-       (i.e. we are validating the parameters of a defined function).
+       (i.e. we are not validating the parameters of a defined function).
        Otherwise, we ensure that the names have no duplicates,
        and we push a new scope for the parameters and the function body,
        but we do not add the parameters to the new scope,


books/kestrel/c/syntax/types.lisp
```diff
@@ -850,172 +906,471 @@
...
-                                 ((:e tau-system)))))
-  :hooks (:fix))
+                                 ((:e tau-system))))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-(define type-compatiblep ((x typep) (y typep))
-  :returns (yes/no booleanp)
-  :short "Check that two @(see type)s are compatible [C17:6.2.7]."
+(define type-da-promote ((type typep) (ienv ienvp))
```
acoglio: `da` seems a bit short to me. What do you think of `defarg`, i.e. `type-defarg-promote`?

gjurgensen: I'm fine with a longer name. How about `type-default-arg-promote`? (I figure someone might read `def` and assume it abbreviates "definition" instead of "default".)

acoglio: Yes, even better/clearer.

books/kestrel/c/syntax/types.lisp
```diff
@@ -850,172 +906,471 @@
...
+                            (type-list-da-promote x.params ienv)
+                            ienv)))
+      :old-style
+      (type-params-case
+        y
+        :prototype (and (not y.ellipsis)
+                        (type-list-compatiblep
+                          (type-list-da-promote x.params ienv)
+                          y.params
+                          ienv))
```
acoglio, line 1139: It looks like there is no requirement between `:old-style` and `:old-style`. This looks consistent with [C17], which simply says nothing about that situation, as opposed to giving requirements for the other situations. It may be worth mentioning this explicitly in the documentation of this function (either as another `xdoc::li` or as text after the `xdoc::ol`), because it took me a bit to realize this, and I expect others may experience the same.

books/kestrel/c/syntax/types.lisp
```diff
@@ -850,172 +906,471 @@
...
+             (type-params-compatiblep x y ienv))
+      :flag type-params-compatiblep)
+    (defthm type-list-compatiblep-symmetric
+      (equal (type-list-compatiblep y x ienv)
+             (type-list-compatiblep x y ienv))
+      :flag type-list-compatiblep)))
+
+(encapsulate ()
+  (local
+    (defun double-list-induct (x y)
```
acoglio: Maybe you could use https://acl2.org/doc/?topic=ACL2____CDR-CDR-INDUCT. It works to include it just locally.

books/kestrel/c/syntax/types.lisp
```diff
@@ -19,12 +20,10 @@
 
 (include-book "std/util/defirrelevant" :dir :system)
 
-(local (include-book "kestrel/built-ins/disable" :dir :system))
-(local (acl2::disable-most-builtin-logic-defuns))
-(local (acl2::disable-builtin-rewrite-rules-for-defaults))
-(local (in-theory (disable (:e tau-system))))
-(set-induction-depth-limit 0)
+(local (include-book "std/basic/controlled-configuration" :dir :system))
```
acoglio: I usually include this non-locally because it is tiny and, if non-local, is only needed in the "leaf" files: most other files can just call the macro (1 line vs. 2 lines).

books/kestrel/c/syntax/types.lisp
```diff
@@ -19,12 +20,10 @@
...
+(local (acl2::controlled-configuration))
```
acoglio: The macro `acl2::controlled-configuration` already generates all-local events, so it doesn't have to be itself local (if you include its book non-locally as suggested in the previous comment).

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1201,14 +1168,90 @@
...
+          arguments given the argument types."
+  :long
+  (xdoc::topstring
+   (xdoc::p
+    "Given a function type which includes a prototype,
+     two conditions must hold [C17:6.5.2.2/2]:")
+   (xdoc::ol
+    (xdoc::li
+     "The number of arguments must be the same as the number of parameter types,
+      or greater than equal when an ellipsis is present.")
```
acoglio:
```suggestion
      or greater than or equal when an ellipsis is present.")
```

acoglio: I'm not readily finding where, in [C17], it says that more arguments than parameters are allowed when the function type has ellipsis, even though it seems right.

books/kestrel/c/syntax/validator.lisp
```diff
@@ -4214,50 +4294,46 @@
...
-       Note that the value of @('fundef-params-p') is the one
-       after validating the inner direct declarator.
-       If we are not validating the declarator of a function definition
-       (i.e. if @('fundef-params-p') is @('nil')),
-       in which case as just mentioned the list must be empty,
-       there is nothing left to do, and we return;
-       note that there is no function prototype scope here.
+       Thus, we raise an error when the list is nonempty
+       and @('fundef-params-p') is @('nil')
+       (i.e. we are validating the parameters of a defined function).
```
acoglio:
```suggestion
       (i.e. we are not validating the parameters of a defined function).
```

books/kestrel/c/syntax/validator.lisp
gjurgensen: Yes, surprisingly I can't find a clear-cut statement of this, but it seems to be implied. [C17:6.5.5.2/7] describes conversion of "trailing arguments" after the last parameter when a function prototype uses the ellipsis notation. [C17:6.7.6.3/9] suggests that there may be more parameters when an ellipsis is supplied and a footnote describes how to access the corresponding arguments.

I think the only concerning statement is [C17:6.5.2.2/2], which says the number of arguments "must agree", with the number of parameters. However, when an ellipsis is present, I think [C17:6.7.6.3/9] suggests that the "number of parameters" is not the number of explicitly listed parameters, but something greater than or equal to that due to the unknown number of additional parameters represented by the ellipsis.

I can also confirm that GCC is happy to accept code like the following under `-std=c17 -pedantic`.
```c
int foo(int x, ...) {
  return x;
}

int bar(void) {
  return foo(0, 1, 2, 3, 4);
}
```

PR #1866, opened 2025-10-31
Improve :doc how-to-contribute.

ericwhitmansmith:
Clarify which things are required.  Order topics for convenience of the contributor.  Suggest adding xdoc.  Suggest projects/ as a good location for new developments.  Make pushing to testing a more prominent requirement (and explain why).


books/system/doc/acl2-doc.lisp
```diff
@@ -52277,36 +52316,31 @@ tables in the current Hons Space."
...
+ <h3>Required: Change the Base Branch of your PR to @('testing')</h3>
+
+ <p>When opening your PR to merge changes into the main ACL2 repository, set
+ the base branch to the @('testing') branch (or similar) (<i>not</i> the
+ @('master') branch).  Once your PR is accepted into the @('testing') branch,
+ it will be tested and merged automatically into the @('master') branch if all
+ tests pass.</p>
 
- <h3>Best Code Practices</h3>
+ <h3>Optional: Follow Best Practices for ACL2 Code</h3>
```
gjurgensen: Maybe something stronger than "Optional", like "Suggested" or "Recommended". Something to indicate that the step is encouraged, just not policed like the others.

Same for the other uses of "Optional" in headers.

books/system/doc/acl2-doc.lisp
```diff
@@ -52277,36 +52316,31 @@ tables in the current Hons Space."
...
 
  <p>See @(see best-practices) for recommended code practices.</p>
 
- <h3>Update the Release Notes</h3>
+ <h3>Optional: Document Your New Books</h3>
+
+ <p>Consider creating documentation using the @(tsee xdoc) system.  Then, to
+ ensure your new topics appear in the manual, ensure they are included by
+ @('books/doc/top.lisp'), perhaps via another book that it includes, such as
+ @('books/projects/top-doc.lisp').</p>
```
gjurgensen: Perhaps this is better suited for `:doc best-practices`? There is a "Documentation" section there.

Other info in this topic also overlap with `best-practices`. Namely, the suggestion to put new developments under the `projects` directory and the warning against fragile timeouts.

Its probably fine to duplicate information between these two topics, especially to highlight common issues in PRs. If this topic ever gets too bloated though, we might want to refactor some info into `best-practices` and more strongly advise new contributors to see that topic.

gjurgensen (approved):

PR #1870, opened 2025-11-13
build/features.sh: Improve SMTLink decision logic.

ericwhitmansmith:
This avoids an issue where 'python' is not available so SMTLink books are attempted but fail.


ericwhitmansmith:
Sol, I think your suggestion implies that SMTLink doesn't actually need to call z3 from the command line, only via Python.  If SMTLink never needs that, your suggestion seems fine to me.

PR #1871, opened 2025-11-15
[C$] Drop default preprocessing arguments and infer `-std=`

gjurgensen:
The external preprocessor no longer assumes any preprocessing argument (outside of the necessary `-E`) when no explicit arguments are provided. `input-files` now chooses a `-std=` argument for the preprocessor based on the `:std` and `:gcc` options.

### DIFF
--- a/books/kestrel/c/syntax/input-files-doc.lisp
+++ b/books/kestrel/c/syntax/input-files-doc.lisp
@@ -148,7 +148,7 @@
                 (xdoc::td "@('-std=c23')")
                 (xdoc::td "@('-std=gnu23')")))
      (xdoc::p
-      "The @(':preprocess-args') specifies additional arguments
+      "The @(':preprocess-args') input specifies additional arguments
        to pass to the preprocessor beyond the two outlined above.
        If @(':preprocess-args') is a string list,
        this list of arguments is passed to the preprocessor for each file,
--- a/books/kestrel/c/syntax/input-files.lisp
+++ b/books/kestrel/c/syntax/input-files.lisp
@@ -201,7 +201,7 @@
        ((when (and (not preprocessor)
                    preprocess-args))
         (reterr (msg "Since the :PREPROCESS input is NIL, ~
-                      the :PREPROCESS-ARGS input must also be @('nil'), ~
+                      the :PREPROCESS-ARGS input must also be NIL, ~
                       but it is ~x0 instead."
                      preprocess-args)))
        ((unless (or (string-listp preprocess-args)
@@ -852,7 +852,7 @@
     "(input-files-prog :files             ...  ; required"
     "                  :path              ...  ; default \".\""
     "                  :preprocess        ...  ; default nil"
-    "                  :preprocess-args   ...  ; no default"
+    "                  :preprocess-args   ...  ; default nil"
     "                  :process           ...  ; default :validate"
     "                  :keep-going        ...  ; default nil"
     "                  :std               ...  ; default 17"


books/kestrel/c/syntax/input-files-doc.lisp
```diff
@@ -121,7 +121,7 @@
        is performed via the @(tsee preprocess-file) tool."))
 
     (xdoc::desc
-     "@(':preprocess-args') &mdash; no default"
+     "@(':preprocess-args') &mdash; default @('nil')"
```
acoglio, line 124: We need to update this in the 'General Form' section too.

books/kestrel/c/syntax/input-files-doc.lisp
```diff
@@ -131,35 +131,41 @@
       "If @(':preprocess') is @('nil'),
        the @(':preprocess-args') input must be absent.")
```
acoglio:
```suggestion
       the @(':preprocess-args') input must be @('nil').")
```

acoglio: Because now that there is a default, its absence is equivalence to its presence with the default (`nil`).

books/kestrel/c/syntax/input-files-doc.lisp
```diff
@@ -131,35 +131,41 @@
...
+                (xdoc::th "@(':gcc nil')")
+                (xdoc::th "@(':gcc t')"))
+      (xdoc::tr (xdoc::th "@(':std 17')")
+                (xdoc::td "@('-std=c17')")
+                (xdoc::td "@('-std=gnu17')"))
+      (xdoc::tr (xdoc::th "@(':std 23')")
+                (xdoc::td "@('-std=c23')")
+                (xdoc::td "@('-std=gnu23')")))
+     (xdoc::p
+      "When present, the @(':preprocess-args') specifies additional arguments
```
acoglio:
```suggestion
      "The @(':preprocess-args') specifies additional arguments
```

acoglio: Now that there is a default, its absence is equivalent to its presence with the default.

books/kestrel/c/syntax/external-preprocessing.lisp
```diff
@@ -108,11 +108,8 @@
     '"gcc")
```
acoglio, line 108: The documentation string for `preprocessor` mentions gcc as default, and then again gcc under "other" reasonable values. Probably the second one should be cpp.

(This is not related to this PR, but something I noticed.)

books/kestrel/c/syntax/input-files-doc.lisp
```diff
@@ -121,45 +121,51 @@
...
+                (xdoc::th "@(':gcc nil')")
+                (xdoc::th "@(':gcc t')"))
+      (xdoc::tr (xdoc::th "@(':std 17')")
+                (xdoc::td "@('-std=c17')")
+                (xdoc::td "@('-std=gnu17')"))
+      (xdoc::tr (xdoc::th "@(':std 23')")
+                (xdoc::td "@('-std=c23')")
+                (xdoc::td "@('-std=gnu23')")))
+     (xdoc::p
+      "The @(':preprocess-args') specifies additional arguments
```
acoglio:
```suggestion
      "The @(':preprocess-args') input specifies additional arguments
```

books/kestrel/c/syntax/input-files.lisp
```diff
@@ -190,19 +188,20 @@
...
-  (b* (((reterr) nil nil)
-       ((when (not preprocess-args-presentp))
-        (retok nil nil))
-       ((when (not preprocessor))
+  (b* (((reterr) nil)
+       ((when (and (not preprocessor)
+                   preprocess-args))
         (reterr (msg "Since the :PREPROCESS input is NIL, ~
-                      the :PREPROCESS-ARGS input must be absent, ~
+                      the :PREPROCESS-ARGS input must also be @('nil'), ~
```
acoglio:
```suggestion
                      the :PREPROCESS-ARGS input must also be NIL, ~
```

acoglio: (Since it's not XDOC.)

books/kestrel/c/syntax/input-files.lisp
```diff
@@ -845,7 +897,6 @@
                              files
```
acoglio, line 897: We also need to update :doc input-files-prog in this file.

PR #1874, opened 2025-12-17
Update movddup.c and elf64

mayuf413:
The 5 status flags (CF, ZF, AF, PF, and SF) pass all tests, but the overflow flag (OF) fails.


mayuf413:
> I am not seeing the failing OF test (and all the tests I see pass). So I am assuming you will add the OF test after I merge this.

I have added the OF test

PR #1876, opened 2025-12-26
[OMAP] Extend omaps library

gjurgensen:
The biggest additions are adding extensionality (+ double containment, pick a point), and fleshing out the `submap` and `compatiblep` theories.

Some of the existing rules in `std/omaps/core` were modified, requiring changes in some community books (mostly `projects/aleo/bft`). In particular:
- We now avoid terms of shape `(consp (assoc key map))`, rewriting to `(and (assoc key map) t)`. This let's us reason about `assoc` under `iff`.
- ~~We rewrite `(set::in key (keys map))` to `(and (assoc key map) t)` by default. This keeps us reasoning about maps directly. Occasionally this needs to be disabled in proofs and the rule rewriting in the opposite direction enabled.~~

Edit: After discussion on Zulip, the change described in the second bullet point was reversed; both directions are now disabled by default and may be enabled when needed.

### DIFF
--- a/books/std/omaps/extensionality.lisp
+++ b/books/std/omaps/extensionality.lisp
@@ -134,7 +134,7 @@
    (xdoc::p
     "This rule rewrites an equality of omaps into an equality of
      @(tsee assoc) on an arbitrary element. The right-hand side is the
-     expansion of @('(ext-equal x y'))."))
+     expansion of @('(ext-equal x y)')."))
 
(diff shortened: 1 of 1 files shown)


acoglio (approved):

gjurgensen:
I'll give it a couple days before merging, in case anyone wants to chime in.

PR #1879, opened 2025-12-31
[defaggregate] Fix tagless :tree case.

ericwhitmansmith:
Previously, two theorems tried to show that aggregates are conses, but tagless aggregates using the :tree layout will be nil if all of their fields are nil.

### DIFF
--- a/books/std/util/defaggregate.lisp
+++ b/books/std/util/defaggregate.lisp
@@ -1315,7 +1315,7 @@ would have had to call @('(student->fullname x)').  For instance:</p>
                                  (str::cat "TYPE-OF-" (symbol-name make-foo))
                                  name)
                         (or (consp (,make-foo ,@field-names))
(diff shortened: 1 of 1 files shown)


gjurgensen:
> I think it seems fine. Could also consider just not producing a type-prescription rule for the constructor in the tagless tree case, but I don't feel strongly either way.

Perhaps an unconditional `type-prescription` rule would be best. Either `true-listp`, or if that isn't true, a disjunction of `consp` and `(equal ... nil)`.

ericwhitmansmith:
I modified this to take Grant's suggestion of an unconditional :type-prescription rule.  A relevant use case might be a value that can either be an aggregate or :none, where the new rule should support proving that the aggregate is never :none (it must always be nil or a cons).  I'll merge this in a few days if nobody objects.

PR #1882, opened 2026-01-02
First pass at ACL2 Bridge for SBCL

### LINKED ISSUE #1880: Port ACL2 Bridge to SBCL
Currently the ACL2 Bridge (`centaur/bridge`) only works on CCL.  Other Lisps are in need and for me that is SBCL.  

I've also got a vibed implementation by Copilot which appears to work or at least run.

jimwhite:
Vibed with Copilot and I can't speak to the quality of this code but it does run the Python demo successfully.
Fixes #1880

### DIFF
--- a/books/centaur/bridge/bridge-sbcl-raw.lsp
+++ b/books/centaur/bridge/bridge-sbcl-raw.lsp
@@ -166,13 +166,7 @@
 
 ; Reading Commands ------------------------------------------------------------
 
-#+sbcl
-(defconstant eof
-  ;; Special constant to denote end of file.
-  ;; Guard against SBCL's strict constant redefinition check on reload.
-  (if (boundp 'eof) (symbol-value 'eof) (cons :eof :eof)))
-#-sbcl
-(defconstant eof
+(defparameter eof
   ;; Special constant to denote end of file
   (cons :eof :eof))
 


ericwhitmansmith:
I believe @westfold is reviewing this.

books/centaur/bridge/bridge-sbcl-raw.lsp
```diff
@@ -0,0 +1,831 @@
...
+(defvar *worker-stack-size*)
+(defvar *worker-vstack-size*)
+(defvar *worker-tstack-size*)
+
+;; Track socket path for cleanup. Unix domain socket files must be explicitly
+;; deleted - socket-close does NOT remove them (standard POSIX behavior).
+(defvar *bridge-socket-path* nil)
+
+;; Shutdown flag for graceful listener termination.
+;; SBCL's destroy-thread is unreliable - use this flag instead.
```
jimwhite, line 492: @westfold 
I don't think this comment it very accurate.  Pretty sure the "unreliable" it refers to is the way we have to poll for threads to terminate which is unlike how CCL threads work where the destroy (seems to) happen atomically.

I could change it to something like:

```
BT:destroy-thread requires us to poll for that to get done so we use this flag.
```

jimwhite:
> I looked at and it generally looks good. 2 Issues:
> 
> 1. There is extra code to get around an sbcl bug in destroy-thread (helpfully mentioned in a comment). However, sbcl's destroy-thread is deprecated in favor of terminate-thread. In fact, the code calls bordeaux-threads:destroy-thread which, for sbcl, calls terminate-thread, so sbcl's destroy-thread is not even called. Thus it seems that the workaround code is unnecessary. Can you persuade Copilot to not generate the workaround code?

The bt:destroy-thread looks correct to me.  It uses BT functions throughout and doesn't do SBCL specific stuff.  
Perhaps you're referring to the "unreliable" comment about SBCL destroy-thread?  I've commented inline about that.

So I don't see workaround code but if I'm missing it please add a comment on the relevant line.

> 2. Most of the code uses quicklisp code which is not sbcl specific. I tried changing the existing bridge-raw.lsp to use the quicklisp functions your vibe code discovered, rather than calling ccl functions directly in most cases. This seems to work and I think it is a better solution. I would like to see thorough testing.I have been trying to test it with an old system of ours that relied on the bridge code, but so far have been foiled by seemingly unrelated issues.

It would be nice to use the same BT solution for both SBCL and CCL, but I think doing just SBCL first should be preferred so that I don't break something that works.  After we get sufficient confidence in this this impl then that would be a separate PR.

I am thinking about adding some additional tests but it isn't obvious to me what they should do.  Are there any libraries or tools that I could look at or run as a stress test?

books/centaur/bridge/bridge-sbcl-raw.lsp
```diff
@@ -0,0 +1,831 @@
...
+  (finish-output stream))
+
+||#
+
+
+
+; Reading Commands ------------------------------------------------------------
+
+#+sbcl
+(defconstant eof
```
jimwhite: @westfold Do you have any idea why this is being done?  Definitely looks suss to me.
I just asked Copilot and it says SBCL is stricter (causing an error rather than warning) on redef from reload than CCL.  It recommended using defvar as a cleaner solution:

```
(defvar eof
  ;; Special sentinel to denote end of file
  (cons :eof :eof))
```

westfold: Yes, the boundp check is nonsense. Redefinition would be an error, so defparameter is appropriate, although defvar would work.

westfold: 1. The bt stuff calls the appropriate implementation-specific functions. It seems that bt::destroy-thread used to call sb::destroy-thread but now calls sb::terminate-thread. Did you add the comment, or did Copilot? The workaround code I am referring to is *bridge-shutdown* and all the code dependent on it. I doubt that this is necessary, although I am not certain.
2. Yes, I agree it makes sense to first use a separate file for sbcl before doing a combined version. More tests would be good, but probably not essential for this merge request.

jimwhite: > 1. The bt stuff calls the appropriate implementation-specific functions. It seems that bt::destroy-thread used to call sb::destroy-thread but now calls sb::terminate-thread. Did you add the comment, or did Copilot? The workaround code I am referring to is _bridge-shutdown_ and all the code dependent on it. I doubt that this is necessary, although I am not certain.

I'm not certain either but had Copilot initial try just using the CCL logic directly but it didn't work because with BT (and SBCL) the termination works differently.  This polling logic using BT then came from how hunchentoot does it.  The verbiage is Copilot's but it isn't always the best/clearest because we went through a bunch of code evolution and there is some detritus from all that.

westfold (approved):
I've looked at this in some detail and tested it with our old syntheto system: it looks good and works well in my tests.

westfold:
I forgot to say that I needed to change defconstant to defparameter for the eof:
(defparameter eof
  ;; Special constant to denote end of file
  (cons :eof :eof))
This could be a misfeature in the way sbcl does the compilation.

PR #1887, opened 2026-01-05
improve error message for a declaration symbol that is the right name…

bendyarm:
… but in the wrong package

Please do not merge unless Matt approves.


MattKaufmann (approved):
Looks good.  @bendyarm and Claude Code did an excellent initial job and I see that my suggestions (sent by email) have been incorporated.

PR #1888, opened 2026-01-05
[books/GNUmakefile] Improve basic target.

ericwhitmansmith:
It was certifying some slow number theory books and various books about primes, just because of std/util/add-io-pairs-tests.lisp.  I excluded that file, and doing so reduced the wall-clock time for make basic by about half, on my big machine.


MattKaufmann (approved):
Looks good to me.  I'll leave it to @ericwhitmansmith to merge when ready.

PR #1892, opened 2026-01-08
Add option in Jenkins scripts to pass make options for ACL2 itself

ericwhitmansmith:
The existing MAKEOPTS argument only applies to making the books.

My initial use for this is to pass ACL2_SAFETY=3.


MattKaufmann (approved):
I'm not very familiar with jenkins, but as far as I can tell this all looks good.  I suppose there could be some surprises if ACL2_PAR, ACL2_REAL, or LISP were supplied in MAKEACL2OPTS, but presumably it would be documented somewhere not to do that.

ericwhitmansmith:
Thanks, Matt.  I add some comments on usage of the new var.  I'll merge this now so I can test it on leeroy with master.

PR #1893, opened 2026-01-11
Make ACL2 Bridge Python and Ruby clients consistent wrt sending (STOP)

jimwhite:
- Don't send (BRIDGE::STOP) command on EOF by default; add --stop flag
- Add documentation about stopping the server to Python README
- Add Ruby installation instructions to Ruby README

### DIFF
--- a/books/centaur/bridge/python/README
+++ b/books/centaur/bridge/python/README
@@ -39,8 +39,8 @@ displayed without disrupting the REPL.
 2. Start ACL2 in a shell
 3. Run ~(ld "demo.lsp")~ in ACL2
 4. Run ~python demo.py~ (or ~python async_demo.py~) in another shell
-5. Use ~^D~ to exit
-6. The Bridge server will continue for other clients.  Stop with ~^C~ or run ~python demo.py --stop~ (or ~python async_demo.py --stop~) to have the client send a (BRIDGE::STOP) command to stop the server.
+5. Use ~^D~ to exit, the Bridge server will continue for other clients.
+6. If demo.py or async_demo.py is ran with the flag --send-stop-on-eof, then the respective Python demo will send a (BRIDGE::STOP) message to the Bridge server when it exits.
 
 ** Manual Stop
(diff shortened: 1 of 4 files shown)


jimwhite:
Ping.

copperteal:
It's fine to change the demos to not send a `(bridge::stop)` by default.  Can you change the flag to `--send-stop-on-eof`?

With respect to the Python README:
---

Step 1 of the updated README does not ensure the ACL2 repo will be able to properly run the bridge server on a fresh clone.  I tested this. A larger build of the community books is required.

However, ACL2 build instructions are out of scope for a document concerned with telling someone how to run a demonstration of the Python Bridge bindings.  So I'd like to drop item 1 entirely.  

Likewise, step 2 is unnecessary and should be dropped. (If the user decides to run ACL2 outside of a shell, that is harmless, and if they don't know that's possible, then they'll assume they need to run it in a shell, and if they don't know how to start ACL2, this is the wrong document).

"Manual Stop" -- alternative means of sending SIGINT to the bridge server are most definitely out of scope for a document introducing the Python bindings.  Maybe they should be in an XDOC page or a README for the Bridge itself?

Item 6 is misleading.  It suggests that running `python demo.py --stop` will kill the bridge server.  It does not.  Can you truncate it to """
6. The Bridge server will continue for other clients.  It can be stopped interactively with \~^C\~ in the ACL2 session.
"""?

However, it is desirable to mention the flag with something like, "If `demo.py` or `async_demo.py` is ran with the flag `--send-stop-on-eof`, then the respective Python demo will send a `(bridge::stop)` message to the Bridge server on shutdown."

With respect to the Ruby README:
---

Rather than assuming the user doesn't know how to obtain Ruby, could we tell them which gems are required and which Ruby versions are compatible with the bindings code?

Just as with the Python code, the bridge will not run properly if only the bridge is certified.

The previous version of the Ruby README had a consistent format of preceding each section header with two empty lines.  That was broken when line 48 (old file, just before "Questions" header) was removed.  Will you restore that line?

PR #1899, opened 2026-01-28
[acl2s] Fix a type error in cgen's callback

mister-walter:
cgen's callback did not properly account for the fact that the ctx argument may be a string and thus under certain circumstances it would try to take the car of a string, causing a type error when ACL2 is built with `(safety 3)`.

This was fixed by adding a `consp` check before the offending `car` form.


mister-walter:
@pmanolios can you run your ACL2s-specific tests on this to ensure I didn't screw anything up? I have an ACL2 regression running now.

Issue #1901, opened 2026-01-28
Highlighted :doc text is not readable on terminal with dark mode


yaitskov:
doc.lisp explicitly sets the white color as a background (control sequence `esc[47m`) for highlighted snippets.
In case of a terminal with a dark color scheme these fragments are barely readable or disappear.
I suggest to either use inversion control sequence (ie 7 instead of 47) or hardcode black foreground together with white background (ESC[30m)

Gnome Terminal (kgx):
<img width="708" height="154" alt="Image" src="https://github.com/user-attachments/assets/188d3bc8-2106-45ae-b49b-569b800191a6" />

kitty:
<img width="708" height="154" alt="Image" src="https://github.com/user-attachments/assets/3a5e0f0e-a1c8-42a1-8de6-d36a70565a9e" />

PR #1902, opened 2026-01-28
ACL2-1901 use color inverse sequence instead of white background

yaitskov:
https://github.com/acl2/acl2/issues/1901

### DIFF
--- a/books/xdoc/display.lisp
+++ b/books/xdoc/display.lisp
@@ -120,9 +120,13 @@
    (list* "b" (sgr-prefix "31;1") *sgr-suffix*)
    (list* "i" (sgr-prefix "3") *sgr-suffix*)
    (list* "u" (sgr-prefix "4") *sgr-suffix*)
-   (list* "tt" (sgr-prefix "47") *sgr-suffix*)
-   (list* "v" (sgr-prefix "47") *sgr-suffix*)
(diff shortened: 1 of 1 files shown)


yaitskov:
> Also: How will this change affect those who use an ordinary white background? Perhaps an environment variable could be used to control this behavior.

Screenshot From 2026-01-29 19-31-47

If the inversion mode distracts attention too much in the day mode, then switching just to bold should look more neutral in both cases. 
ACL2 is shipped with the pre-generated doc file, and it is even under the version control.
Btw, I haved found info how to regenerate the doc.lisp file. I guess I have to do `acl2 < doc.lisp`, but I haven't tried.

MattKaufmann:
@gjurgensen  I've been assuming that this is about the output of :DOC at the terminal (either before or after including xdoc, e.g., via (include-book "xdoc/init" :dir :system)), which I think would apply in the same way to using the emacs-based acl2-doc browser (https://acl2.org/doc/?topic=ACL2____ACL2-DOC).

yaitskov:
> you make this change, I'll accept the PR;

Thanks, @MattKaufmann. I like the proposed style for dark mode. I squashed commits with latest version and rebased the branch.

@gjurgensen, thanks for the workaround with the Emacs trick. I didn't know about it.

PR #1903, opened 2026-02-03
[doc] Add new doc topic on Emacs workflow.

ericwhitmansmith:
It seems that new ACL2 users, especially ones not trained by an experienced user, do not always understand the expected Emacs-based workflow for using ACL2.  This new xdoc topic attempts to remedy that.  Feedback welcome!

Thanks to Matt Kaufmann for helpful feedback on a draft of this.


books/doc/practices.lisp
```diff
@@ -642,3 +642,102 @@ whose effect may be more easy to guess.</p>")
...
+
+    (xdoc::p "Typically, you want to develop a @('.lisp') file containing ACL2
+    @(see events) (definitions, theorems, etc.).  In ACL2 terminology, we call
+    this @('.lisp') file a <i>book</i>.  Your goal is to get ACL2 to accept
+    everything in the book.")
+
+    (xdoc::p "In Emacs, you open the book for editing and you also start a shell
+    buffer (within Emacs, via a command like @('M-x shell')).  You start ACL2 in
+    the shell buffer, which starts up its REPL (Read-Eval-Print Loop).  Now you
+    can submit commands to ACL2 REPL via the shell buffer.")
```
gjurgensen:
```suggestion
    can submit commands to the ACL2 REPL via the shell buffer.")
```

books/doc/practices.lisp
```diff
@@ -642,3 +642,102 @@ whose effect may be more easy to guess.</p>")
...
+    the book.")
+
+    (xdoc::p "Various shortcuts are available:")
+
+    (xdoc::ol
+
+      (xdoc::li "ACL2 provides custom emacs commands to easily submit events
+      and sequences of events to the shell buffer.  See @(see emacs), which
+      describes how to make those commands available.")
+
```
gjurgensen: Perhaps it should also be mentioned that the best way to find out which commands are provided and their default key bindings is to look at the distributed emacs file itself. I see that `:doc emacs`, linked here, says the following:

> The file begins with considerable comments describing what it offers.

Maybe that is enough to direct people to look at the file source.

books/system/doc/acl2-doc.lisp
```diff
@@ -32310,26 +32310,34 @@ ld) and @(tsee include-book)"
...
- (load \"DIR/books/emacs/emacs-acl2.el\") ; for recent Emacs versions
+ (load \"DIR/books/emacs/emacs-acl2.el\")
  })
 
- <p><b>-OR-</b></p>
+ <p>where @('DIR') should be changed to the path to your main ACL2 directory.
+ </p>
+
+ <p>An easy way to arrange this is to put this @('load') command into your
+ @('.emacs') file in your home directory.</p>
```
gjurgensen: Many users will have a different Emacs init file name/path. In particular, `.emacs.d/init.el` is common. (This source actually refers to this as typical, and `.emacs` as "archaic": https://wikemacs.org/wiki/User's_Initialization_File. That is much stronger than the actual GNU Emacs manual, which says `.emacs` is the "traditional" location: https://www.gnu.org/software/emacs/manual/html_node/emacs/Init-File.html.)

```suggestion
 Emacs init file (e.g. @('~/.emacs.d/init.el'), @('~/.emacs')).</p>
```

PR #1904, opened 2026-02-05
[C$] Extend struct and union types

gjurgensen:
This extends the information we track for struct and union types. Both types now include a UID, an optional translation unit name, and the `tag/members`, which has two cases: `:tagged`, in which case we have a name, and `:untagged`, in which case we have the members. The UID is necessary to distinguish distincts types in the same translation unit with the same name. The translation unit name is necessary for type compatibility. If there is no translation unit name, this means the type was created as the type composite of two struct/union types from different translation units.

A "type-completions" map is added to associate struct/union UIDs to their members. This is a `fast-alist`. It was originally an `omap`, but the map can get quite large. The linear-time look-ups significantly deteriorated the performance.

### DIFF
--- a/books/kestrel/c/syntax/types.lisp
+++ b/books/kestrel/c/syntax/types.lisp
@@ -1830,7 +1830,7 @@
      alongside our representation of types,
      in order to reflect true type compatibility.")
    (xdoc::p
-    "Finally, we note the perhaps surprising fact
+    "Finally, we note the perhaps surprising fact that
      type compatibility is intransitive and therefore
      not an equivalence relation.
      For instance, consider the following functions.")
--- a/books/kestrel/c/syntax/validation-information.lisp
+++ b/books/kestrel/c/syntax/validation-information.lisp
@@ -437,6 +437,10 @@
      This information accumulates
      as we validate each translation unit in the ensemble.")
    (xdoc::p
+    "The @('completions') field is a map of @(see type-completions).
+     This maps @(see UID)s corresponding to struct and union types
+     to their list of named members.")
+   (xdoc::p
     "The @('next-uid') field stores the next unused "
     (xdoc::seetopic "uid" "unique identifier")
     "."))
--- a/books/kestrel/c/syntax/validator.lisp
+++ b/books/kestrel/c/syntax/validator.lisp
@@ -1777,9 +1777,9 @@
       and the other is a pointer to the @('void') type.
       As a GCC/Clang extension,
       we also allow one operand to be a pointer to the @('void') type,
-      and the other is a pointer to <i>any</i> type
+      and the other to be a pointer to <i>any</i> type
       (either object or function).
-      We do not aware of explicit GCC documentation of this feature,
+      We are not aware of explicit GCC documentation of this feature,
       but a related feature is listed by the standard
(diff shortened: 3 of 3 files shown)


books/kestrel/c/syntax/types.lisp
```diff
@@ -1682,7 +2579,9 @@
    :ulong (type-ulong)
    :sllong (type-sllong)
    :ullong (type-ullong)
-   :struct (type-struct (ident (c::ident->name ctype.tag)))
+   ;; TODO: we can't really create a struct, unless we wanted to invent a UID
+   ;; and tunit. Then, we could perhaps create an incomplete struct type.
+   :struct (irr-type)
```
gjurgensen, line 2584: It wasn't clear to me what to do with `ildm-type` here. The representation of the struct type here has more information than the version in the `c` package.

books/kestrel/c/syntax/uid.lisp
```diff
@@ -0,0 +1,104 @@
+; C Library
+;
+; Copyright (C) 2025 Kestrel Institute (http://www.kestrel.edu)
```
acoglio:
```suggestion
; Copyright (C) 2026 Kestrel Institute (http://www.kestrel.edu)
```

books/kestrel/c/syntax/types.lisp
```diff
@@ -1161,118 +1633,503 @@
...
+      and their @(tsee type-params) are compatible
+      (see @(tsee type-params-compatible-p-aux)).")
+    (xdoc::li
+     "For any other case, the types are compatible only if they are equal."))
+   (xdoc::p
+    "Eventually, we shall refine the notion of compatibility,
+     alongside our representation of types,
+     in order to reflect true type compatibility.")
+   (xdoc::p
+    "Finally, we note the perhaps surprising fact
```
```suggestion
    "Finally, we note the perhaps surprising fact that
```

books/kestrel/c/syntax/types.lisp
acoglio: Right, we don't have a lot of choices at this time.

books/kestrel/c/syntax/validation-information.lisp
```diff
@@ -427,6 +443,7 @@
   ((filepath filepath)
    (scopes valid-scope-list)
    (externals valid-externals)
+   (completions type-completions)
```
acoglio, line 450: Maybe we should add a sentence to this doc mentioning the `completions` component?

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1658,7 +1774,14 @@
      "Both operands have compatible pointer types.")
     (xdoc::li
      "One operand is a pointer to an object type
-      and the other is a pointer to the @('void') type.")
+      and the other is a pointer to the @('void') type.
+      As a GCC/Clang extension,
+      we also allow one operand to be a pointer to the @('void') type,
+      and the other is a pointer to <i>any</i> type
+      (either object or function).
+      We do not aware of explicit GCC documentation of this feature,
```
acoglio:
```suggestion
      We are not aware of explicit GCC documentation of this feature,
```

PR #1928, opened 2026-03-31
x86isa testgen fix Makefile deps

yaso9:
- **Fix testgen Makefile deps**


acoglio (approved):
I'm not familiar with the details, but it looks reasonable.

PR #1931, opened 2026-04-14
[C$] Add support for basic constant evaluation.

gjurgensen:
The symbol `'value` is no longer exported from the ACL2 package. This PR therefore features a number of incidental changes, where the `acl2::value` function had to be explicitly prefixed by the acl2 package.

### DIFF
--- a/books/kestrel/c/syntax/evaluation.lisp
+++ b/books/kestrel/c/syntax/evaluation.lisp
@@ -1,0 +1,1248 @@
+; C Library
+;
+; Copyright (C) 2026 Kestrel Institute (http://www.kestrel.edu)
+;
+; License: A 3-clause BSD license. See the LICENSE file distributed with ACL2.
+;
+; Author: Grant Jurgensen (grant@kestrel.edu)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(in-package "C$")
+
+(include-book "abstract-syntax-operations")
+(include-book "implementation-environments")
+
+(include-book "xdoc/defxdoc-plus" :dir :system)
+
+(include-book "centaur/fty/deftypes" :dir :system)
+
+(include-book "kestrel/utilities/arith-fix-and-equiv-defs" :dir :system)
+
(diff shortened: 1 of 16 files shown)


books/kestrel/c/syntax/evaluation.lisp
```diff
@@ -0,0 +1,1248 @@
...
+(define value-signed-integerp ((val valuep))
+  :returns (yes/no booleanp)
+  :parents (value)
+  :short "Recognize values with signed integer type [C17:6.2.5/4]."
+  :long
+  (xdoc::topstring-p
+   "This predicate is overapproximate in the sense that
+    it also recognizes unknown values &mdash;
+    i.e. values which <emph>might</emph> have signed integer type.")
+  (and (value-case val '(:schar :sshort :sint :slong :sllong))
```
acoglio:
```suggestion
  (and (value-case val '(:unknown :schar :sshort :sint :slong :sllong))
```

books/kestrel/c/syntax/evaluation.lisp
```diff
@@ -0,0 +1,1248 @@
...
+  ///
+  (fty::deffixequiv-mutual well-formed-value/value-list-p))
+
+;;;;;;;;;;;;;;;;;;;;
+
+(defrule value-ushort->get-when-well-formed-value-p-linear
+  (implies (well-formed-value-p val ienv)
+           (<= (value-ushort->get val)
+               (+ -1 (expt 2 (* 8 (ienv->short-bytes ienv))))))
+  :rule-classes :linear
```
acoglio: It may be worth explicating the trigger term for the linear rule, i.e. (value-ushort->get val). I'm not sure if that's what ACL2 would pick by default. Similar remark for similar rules.

books/kestrel/c/syntax/evaluation.lisp
```diff
@@ -0,0 +1,1248 @@
...
+  :inline t)
+
+;;;;;;;;;;;;;;;;;;;;
+
+(defrule value-kind-of-value-uint-mod
+  (equal (value-kind (value-uint-mod get ienv))
+         :uint)
+  :enable value-uint-mod)
+
+(defrule value-uint->get-of-value-uint-mod
```
acoglio: Do we want this always enabled? Not sure.

acoglio (approved):

PR #1937, opened 2026-04-21
[quicklisp cffi] remove code that was causing osicat bug

bendyarm:
The code removed is used for bundle operations,
like building a static lib for ECL, which we don't do in ACL2.

Also, the code removed should have been done differently,
like to add methods to bundle operations so they pick up
output files of `process-op`.

This change will do for now.  We should file an issue with ASDF/CFFI
When ASDF/CFFI are updated we should get the new version of CFFI.

Thanks to Qiantan Hong for diagnosing and fixing this bug.


solswords (commented):
No idea of the specifics, but it seems fine. However, could you add a patch note to the others at the bottom of update-libs.sh? It seems like that's how we've been keeping track of manual changes to bundled sources.

gjurgensen:
Perhaps `:doc quicklisp` should also be edited to instruct users to look at their stdout output after running `update-libs.sh` and to follow the patching instructions. Just reading the docs, I would have assumed running the script is sufficient, and might not pay attention to the output if the script succeeded (i.e. error code 0).

westfold:
As far as I can tell, it is fine.

bendyarm:
Fixes #1938.

PR #1939, opened 2026-04-24
[C$] Extend validation of initializers

gjurgensen:
This adds the notion of initializer subobjects, subobjects stacks, and initializer contexts, defined in `validation-information.lisp`, to represent the necessary information for the validator to check iniitializers.

The validator termination is modified to take a step counter for the `valid-initer`. The reason is described in the new documentation.

Several test cases were added of initialization and compound literals.

### DIFF
--- a/books/kestrel/c/syntax/validation-information.lisp
+++ b/books/kestrel/c/syntax/validation-information.lisp
@@ -13,6 +13,7 @@
 (include-book "types")
 (include-book "uid")
 (include-book "unambiguity")
+(include-book "evaluation")
 
 (include-book "kestrel/fty/deffold-reduce" :dir :system)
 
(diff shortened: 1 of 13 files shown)


books/kestrel/c/syntax/validation-information.lisp
```diff
@@ -46,6 +51,573 @@
...
+     "// Assume this is in translation unit named \"foo.c\""
+     "// and `struct foo_s` has UID 0."
+     "struct foo_s {"
+     "  int x;"
+     "  int y;"
+     "};"
+     ""
+     "struct foo_s foo[10] = {"
+     "  [5].x = 0,"
+     "  1,"
```
acoglio, line 170: This is equivalent to `[5].y = 1,`, right? If so, you could consider adding that as a comment, to help the reader further.

acoglio (approved):

PR #1940, opened 2026-04-27
Extend whitelist for `disable-most-builtin-logic-defuns`

gjurgensen:
This adds many symbols to the whitelist. divided into three categories: synonyms, identities, and constants.


gjurgensen:
I would also advocate for adding `atom`, `endp`, and `null` to the whitelist. But I expect those may be more controversial, so I left them out of this PR.

Of those, `endp` is likely the least controversial as a simple abbreviation of `atom`. `null` is essentially just `not`. `atom` might be comparable to `/=`, which we enable despite not being a true synonym.

acoglio (approved):

PR #1942, opened 2026-05-27
[C$] Refine our model of array types.

acoglio:
Add an optional positive integer as size. Extend notions of type compatibility and composite types accordingtly.

Since this matches the notion of array types in the language formalization, the formalp predicates and the mapping to the language formalization have been extended to include array types.

There are intentional TODOs in the validator to try and set the size if possible.  It is currently always set to `nil` (i.e. unknown size), so the behavior is essentially like before this change. But this change paves the way for further changes.


gjurgensen (approved):
LGTM.

PR #1943, opened 2026-05-28
[OMAP] Adding functions: compose, identityp, injectivep, inverse, restrict-values, and closedp

ltmquan:
Added 7 files:
 - compose.lisp: map composition
 - identityp.lisp: identifier for identity maps
 - injectivep.lisp: identifier for injective maps
 - inverse.lisp: map inverse
 - restrict-values.lisp: restrict a map to a set of values
 - closedp.lisp: identifier for closed maps (value set is a subset of key set)
 - restrict.lisp: theorem relating compatiblep and restrict
 
 @gjurgensen


books/std/omaps/assoc.lisp
```diff
@@ -80,3 +80,22 @@
                      val)))
   :induct t
   :enable rlookup)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defrule cons-of-key-and-cdr-assoc
```
gjurgensen: I think I would disable this by default and enable it only when needed.

books/std/omaps/closedp.lisp
```diff
@@ -0,0 +1,34 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define closedp ((x mapp))
+  :returns (yes/no booleanp)
+  :short "Check if an omap is closed under its domain."
+  :long
+  (xdoc::topstring-p
+   "An omap is closed if it is empty or
+    its set of values is a subset of its set of keys.")
+  (set::subset (values x) (keys x))
```
gjurgensen: Is this `set::` necessary?

books/std/omaps/compose.lisp
```diff
@@ -0,0 +1,54 @@
...
+   "Maps key @('k') to @('w') when @('y') maps @('k') to @('v')
+    and @('x') maps @('v') to @('w').
+    Keys of @('y') with no matching entry in @('x') are dropped.")
+  (if (emptyp y)
+      nil
+      (mv-let (k v) (head y)
+        (let ((pair (assoc v x)))
+          (if pair
+              (update k (cdr pair) (compose x (tail y)))
+              (compose x (tail y))))))
```
gjurgensen: Please format the `if` else branches with 2 spaces instead of 4. See the "formatting" section of `:doc best-practices`.

books/std/omaps/identityp.lisp
```diff
@@ -0,0 +1,153 @@
...
+  ///
+
+  (defcong mequiv equal (identityp x) 1)
+
+  (defrule values-is-keys-when-identityp
+    (implies (identityp x)
+             (equal (values x)
+                    (keys x)))
+  :enable (values keys)
+  :rule-classes (:rewrite :forward-chaining))
```
gjurgensen: I think just `:rewrite` is fine here. No need for `:forward-chaining`.

books/std/omaps/identityp.lisp
```diff
@@ -0,0 +1,153 @@
...
+
+  (defrule assoc-of-identityp
```
gjurgensen: I'd call this `assoc-when-identityp`.

books/std/omaps/injectivep.lisp
```diff
@@ -0,0 +1,142 @@
...
+
+  ///
+
+  (defcong mequiv equal (injectivep x) 1)
+
+  (defrule injectivep-implies-injectivep-tail
+      (implies (and (injectivep x)
+                    (not (emptyp x)))
+               (injectivep (tail x)))
+    :rule-classes :type-prescription)
```
gjurgensen: I would remove this rule if possible, or change it to a `:rewrite` rule.

books/std/omaps/injectivep.lisp
```diff
@@ -0,0 +1,142 @@
...
+     This is more efficient for execution than the recursive membership checks
+     in the @(':logic') version."))
+  (mbe :logic
+       (or (emptyp x)
+           (b* (((mv & v) (head x)))
+             (and (not (set::in v (values (tail x))))
+                  (injectivep (tail x)))))
+       :exec (equal (cardinality (keys x))
+                    (cardinality (values x))))
+  :verify-guards nil
```
gjurgensen: I like to add a comment like `;; Verified below` to indicate that the function is *eventually* guard-verified.

books/std/omaps/inverse.lisp
```diff
@@ -0,0 +1,124 @@
...
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define inverse ((x mapp))
+  :parents (omaps)
+  :returns (map mapp)
+  :short "Compute the inverse of an omap."
+  :long
+  (xdoc::topstring-p
+   "Maps each value @('v') of @('x') back to its corresponding key @('k').
```
gjurgensen, line 29: I think it would be best to be more explicit about what the behavior is for non-injective maps. (I.e., it maps each value back to an arbitrary choice of one of its keys.)

books/std/omaps/inverse.lisp
```diff
@@ -0,0 +1,124 @@
...
+               (equal (values (inverse x))
+                      (keys x)))
+    :enable assoc-to-in-of-keys)
+
+  (defrule inverse-implies-injectivep
+      (implies (injectivep x)
+               (injectivep (inverse x)))
+    :enable (injectivep
+             in-of-keys-to-assoc)
+    :rule-classes :type-prescription)
```
gjurgensen: I would again avoid `:type-prescription` and do a regular rewrite rule instead.

books/std/omaps/inverse.lisp
```diff
@@ -0,0 +1,124 @@
...
+                      (assoc k x)))
+    :enable (injectivep
+             set::expensive-rules
+             rlookup-to-in-of-values
+             in-of-keys-to-assoc))
+
+  (defrule inverse-inverse-when-injectivep
+      (implies (and (injectivep x)
+                    (mapp x))
+               (equal (inverse (inverse x)) x))
```
gjurgensen: I'd suggest dropping the `(mapp x)` hypothesis and replacing the RHS of the equality with `(mfix x)`.

books/std/omaps/restrict-values.lisp
```diff
@@ -0,0 +1,80 @@
...
+   "Retains only the key-value pairs of @('x')
+    whose associated value belongs to @('values').
+    All other pairs are dropped.")
+  (cond ((emptyp x) nil)
+        (t (mv-let (k v)
+               (head x)
+             (if (set::in v values)
+                 (update k v
+                         (restrict-values values (tail x)))
+               (restrict-values values (tail x))))))
```
gjurgensen: I would rewrite this `cond` to an `if` since it has only the two branches.

books/std/omaps/restrict.lisp
```diff
@@ -0,0 +1,95 @@
...
+
+(defrule equal-of-update-is-equal-of-delete
+  (equal (equal (update k v x) y)
+         (and (mapp y)
+              (assoc k y)
+              (equal (cdr (assoc k y)) v)
+              (equal (delete k x)
+                     (delete k y))))
+  :use equal-of-delete-implies-equal-of-update)
+
```
gjurgensen: I would disable this be default. Also move it to `update.lisp` or `delete.lisp` (I'd prefer the former).

gjurgensen (changes requested):

gjurgensen:
You should also include these new books in `top.lisp`.

ltmquan:
@gjurgensen I just made the suggested changes.

PR #1945, opened 2026-05-31
[C$] Improve the precision of the validator.

acoglio:
Add and use a version of the unknown type that restricts it to be scalar.

This was Grant Jurgensen's idea.


books/kestrel/c/syntax/validator.lisp
```diff
@@ -1596,46 +1609,52 @@
...
-                     (retok (type-unknown)))
+      (:bitnot (b* (((when (or (type-case type-arg :unknown)
+                               (type-case type-arg :unknown-scalar)))
+                     (retok (type-unknown-scalar)))
                     ((unless (type-integerp type-arg))
                      (reterr msg)))
                  (retok (type-integer-promote type-arg ienv))))
-      (:lognot (b* (((when (type-case type-arg :unknown))
-                     (retok (type-unknown)))
+      (:lognot (b* (((when (and (type-case type-arg :unknown)
```
gjurgensen: Should be `or`, not `and`.

gjurgensen:
I think `valid-arrsub` and `valid-cond` would also need to be updated.

gjurgensen:
The `:indir` case of `valid-unary` should check if `unknown-scalar` in addition to `unknown`.

gjurgensen (approved):
LGTM. I tested against nginx, which successfully validated.

PR #1946, opened 2026-06-01
[OMAP] Adding functions: compose, identityp, injectivep, inverse, restrict-values, and closedp

ltmquan:
Added 7 files:

compose.lisp: map composition
identityp.lisp: identifier for identity maps
injectivep.lisp: identifier for injective maps
inverse.lisp: map inverse
restrict-values.lisp: restrict a map to a set of values
closedp.lisp: identifier for closed maps (value set is a subset of key set)
restrict.lisp: theorem relating compatiblep and restrict

@gjurgensen


books/std/omaps/closedp.lisp
```diff
@@ -0,0 +1,35 @@
...
+  :long
+  (xdoc::topstring-p
+   "An omap is closed if it is empty or
+    its set of values is a subset of its set of keys.")
+  (subset (values x) (keys x))
+  ///
+
+  (defcong mequiv equal (closedp x) 1)
+
+  (defrule identityp-implies-closedp
```
acoglio: Not sure whether this should be enabled by default. How often, when `(clodesp <map>)` comes up, the `<map>` is an identity map?

books/std/omaps/compose.lisp
```diff
@@ -0,0 +1,54 @@
...
+          (compose x (tail y))))))
+  :verify-guards :after-returns
+
+  ///
+
+  (defcong mequiv equal (compose x y) 1)
+
+  (defcong mequiv equal (compose x y) 2)
+
+  (defrule assoc-of-compose
```
acoglio, line 44: Not sure I would enable this by default: the RHS is a more complex than the LHS, and I don't see a clear reason why it is "simpler" — but if there is a reason that I'm missing, it would be good to document.

books/std/omaps/identityp.lisp
```diff
@@ -0,0 +1,160 @@
...
+  (or (emptyp x)
+      (mv-let (k v) (head x)
+        (and (equal k v)
+             (identityp (tail x)))))
+  
+  ///
+
+  (defcong mequiv equal (identityp x) 1)
+
+  (defrule values-is-keys-when-identityp
```
acoglio, line 42: Similar comment to another rule: how often, when `(values <map>)` comes up, the `<map>` is an identity?

books/std/omaps/identityp.lisp
```diff
@@ -0,0 +1,160 @@
...
+    (implies (identityp x)
+             (equal (values x)
+                    (keys x)))
+  :enable (values keys))
+
+  (defrule assoc-when-identityp
```
acoglio, line 48: Analogous comment as above.

books/std/omaps/identityp.lisp
```diff
@@ -0,0 +1,160 @@
...
+    :enable (assoc assoc-of-restrict in-of-keys-to-assoc))
+
+  (defruled compose-is-restrict-when-Y-identityp
+      (implies (identityp y)
+               (equal (compose x y)
+                      (restrict (keys y) x)))
+    :enable (compose-is-restrict-when-Y-identityp-helper
+             extensionality))
+
+  (defrule self-compose-is-self-when-identityp
```
acoglio: Does it work to eliminate the `(mapp x)` hyp and change the RHS to `(mfix x)`?

Also, same comment about enablement as others.

books/std/omaps/identityp.lisp
```diff
@@ -0,0 +1,160 @@
...
+      (implies (and (identityp x)
+                    (mapp x))
+               (equal (compose x x) x))
+    :enable extensionality))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define-sk identityp-sk ((x mapp))
```
acoglio, line 106: Can we have this function fix `x` to be a map, like `identityp` implicitly does? That way, the two definitions are unconditionally equivalent, and the `(mappx )` hyp can be removed from identityp-when-identityp-sk.

books/std/omaps/identityp.lisp
```diff
@@ -0,0 +1,160 @@
...
+
+(defruled identityp-when-identityp-sk
+    (implies (and (identityp-sk x)
+                  (mapp x))
+             (identityp x))
+  :hints ('(:use (:instance identityp-sk-necc
+                            (k (mv-nth 0 (head x))))))
+  :enable (identityp
+           identityp-sk-of-tail-when-identityp-sk))
+
```
acoglio, line 140: It could be also useful to have a disabled rule that rewrites `identityp-sk` to `identityp`, with a theory invariant. This way we can move in either direction in different proofs as needed.

books/std/omaps/injectivep.lisp
```diff
@@ -0,0 +1,141 @@
...
+                    (cardinality (values x))))
+  :verify-guards nil ;; Verified below
+
+  ///
+
+  (defcong mequiv equal (injectivep x) 1)
+
+  (defrule injectivep-implies-injectivep-tail
+      (implies (and (injectivep x)
+                    (not (emptyp x)))
```
acoglio: Do we need the non-emptiness hyp? I think that tail of empty is nil.

books/std/omaps/injectivep.lisp
```diff
@@ -0,0 +1,141 @@
...
+               (injectivep (tail x))))
+
+  (defrule injectivep-implies-unique-head-value
```
acoglio: I would disable this and the next. They backchain to injective from something fairly generic that doesn't necessarily relate to injectivity.

books/std/omaps/injectivep.lisp
```diff
@@ -0,0 +1,141 @@
...
+             (injectivep x)))
+
+  (theory-invariant (incompatible (:rewrite injectivep-to-cardinality-of-keys-and-values)
+                                  (:rewrite cardinality-of-keys-and-values-to-injectivep)))
+
+  (verify-guards injectivep
+      :hints (("Goal" :in-theory
+                      (enable cardinality-of-keys-and-values-to-injectivep))))
+
+  (defrule identityp-implies-injectivep
```
acoglio: I would probably disable it, since many injective maps are not the identity map.

Perhaps a middle ground is to disable this and others to be conservative, but to also provide a ruleset with this and other rules enabled, a bit like the `set::expensive-rules` ruleset in osets.

books/std/omaps/injectivep.lisp
```diff
@@ -0,0 +1,141 @@
...
+      (implies (identityp x)
+               (injectivep x))
+    :enable injectivep-to-cardinality-of-keys-and-values
+    :use values-is-keys-when-identityp
+    :rule-classes :forward-chaining)
+
+  (defrule rlookup-of-cdr-assoc-when-injectivep
```
acoglio: I would disable.

books/std/omaps/inverse.lisp
```diff
@@ -0,0 +1,126 @@
...
+  (defrulel assoc-of-inverse-when-assoc
+      (implies (and (injectivep x)
+                    (assoc v x)
+                    (equal (cdr (assoc v x)) k))
+               (equal (assoc k (inverse x))
+                      (cons k v)))
+    :enable values
+    :use equal-val-implies-equal-key-when-injectivep)
+
+  (defrule injectivep-implies-not-rlookup-head-val-tail
```
acoglio, line 80: Maybe disable.

books/std/omaps/inverse.lisp
```diff
@@ -0,0 +1,126 @@
...
+    :enable rlookup-to-in-of-values)
+
+  (defrule assoc-of-inverse
+      (implies (injectivep x)
+               (equal (assoc k (inverse x))
+                      (and (set::in k (values x))
+                           (cons k (set::head (rlookup k x))))))
+    :enable (rlookup values set::expensive-rules))
+
+  (defrule rlookup-of-update-when-not-assoc
```
acoglio: This does not involve inverse, so we could move it to a more general place.

books/std/omaps/restrict-values.lisp
```diff
@@ -0,0 +1,81 @@
...
+    :enable (restrict-values set::emptyp)
+    :rule-classes (:rewrite :type-prescription))
+
+  (defrule restrict-values-when-right-emptyp
+      (implies (emptyp x)
+               (equal (restrict-values vs x) nil))
+    :enable restrict-values
+    :rule-classes (:rewrite :type-prescription))
+
+  (defrule assoc-of-restrict-values
```
acoglio, line 58: Maybe disable, since the RHS is not obviously "simpler".

acoglio (changes requested):

books/std/omaps/identityp.lisp
gjurgensen: I'm not sure fixing `x` explicitly matters, since we are only using it in the context of `assoc`, which should respect `mequiv`. I imagine the `(mapp x)` hypothesis can be removed from `identityp-when-identityp-sk` rule with the current definition (and we should).

books/std/omaps/injectivep.lisp
ltmquan: Do you think we should create an `omap::expensive-rules` for the whole book or should I just create a ruleset for these newly added files for now?

acoglio: It could make sense to have a ruleset of "expensive rules" that contains both some of the new rules and some of the old (i.e. existing) rules. But perhaps we could do that in a separate PR, just to get this one finalized.

What does @gjurgensen think?

gjurgensen: I'd suggest sticking to a small change for now. Either disabling it, or creating a ruleset (`omap::expensive-rules`) containing just some subset of these new rules from the new books.

As @acoglio says, we can always go back later in a separate PR and add some existing rules to said ruleset.

ltmquan:
I changed some `defrule`s to be `defruled`s and disabled others at the end of each file (for rules that were useful when enabled in that file)

acoglio (approved):

gjurgensen (approved):

PR #1947, opened 2026-06-01
[unicode] add some utility theorems to facilitate round-trip theorems…

bendyarm:
… of users of unicode such as parsers.

### DIFF
--- a/books/unicode/uchar.lisp
+++ b/books/unicode/uchar.lisp
@@ -456,8 +456,7 @@
 (defthm nat-listp-when-ustring?
   (implies (ustring? x)
            (nat-listp x))
-  :rule-classes ((:rewrite)
-                 (:forward-chaining)))
+  :rule-classes ((:forward-chaining)))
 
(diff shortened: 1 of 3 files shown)


books/unicode/uchar.lisp
```diff
@@ -453,7 +453,16 @@
   (implies (ustring? x)
            (ustring? (cdr x))))
 
+(defthm nat-listp-when-ustring?
```
solswords, line 456: I'd suggest disabling the rewrite rule generated by this theorem externally, just in case unicode books are incidentally included somewhere when reasoning about nat-listp in a context that doesn't have to do with unicode.

solswords (approved):

books/unicode/uchar.lisp
bendyarm: Good point.  Actually the rewrite rule is not needed for anything; I decided just to remove it.

acoglio (approved):

PR #1949, opened 2026-06-03
[C$] Improve precision of validator.

acoglio:
Add and use an 'unknown arithmetic type'.


books/kestrel/c/syntax/validator.lisp
```diff
@@ -1202,25 +1202,33 @@
...
+                       (type-case type2 :unknown-scalar)
+                       (type-case type2 :unknown-arithmetic))))
         (retok (type-pointer->to type1)))
        ((when (and (type-case type2 :pointer)
                    (or (type-integerp type1)
+                       ;; type1 could be integer:
                        (type-case type1 :unknown)
-                       (type-case type1 :unknown-scalar))))
+                       (type-case type1 :unknown-scalar)
+                       (type-case type2 :unknown-arithmetic))))
```
gjurgensen: I think this should be `(type-case type1 :unknown-arithmetic)`. Right now, the condition can't hold since we know `type2` is a `:pointer`.

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1202,25 +1202,33 @@
...
-       ((when (and (or (type-case type2 :unknown-scalar)
-                       (type-case type2 :unknown))
+       ((when (and (or (type-case type2 :unknown-scalar) ; type2 could be
+                       (type-case type2 :unknown))       ; pointer
                    (or (type-integerp type1)
+                       ;; type1 could be integer:
                        (type-case type1 :unknown)
-                       (type-case type1 :unknown-scalar))))
+                       (type-case type1 :unknown-scalar)
+                       (type-case type2 :unknown-arithmetic))))
```
gjurgensen: Same comment as above. I think this should be `type1`, not `type2`.

books/kestrel/c/syntax/validator.lisp
```diff
@@ -2002,26 +2124,43 @@
...
+                       (type-case type-arg1 :unknown-scalar)
+                       (type-case type-arg1 :unknown-arithmetic)
+                       (type-case type-arg2 :unknown)
+                       (type-case type-arg2 :unknown-scalar)
+                       (type-case type-arg2 :unknown-arithmetic)))
+             (retok (type-unknown-arithmetic)))
+            ((unless (and (type-arithmeticp type-arg1)
                           (type-arithmeticp type-arg2)))
              (reterr msg)))
          (retok (type-fix type-arg1))))
```
gjurgensen, line 2054: Previously `:asm-rem` required `type-integerp` of both arguments, not `type-arithmeticp`. Is this change intended?

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1894,56 +1908,138 @@
...
-       ((when (or (type-case type-arg1 :unknown)
-                  (type-case type-arg1 :unknown-scalar)
-                  (type-case type-arg2 :unknown)
-                  (type-case type-arg2 :unknown-scalar)))
-        (retok (type-unknown-scalar)))
+       ;; ((when (or (type-case type-arg1 :unknown)
+       ;;            (type-case type-arg1 :unknown-scalar)
+       ;;            (type-case type-arg2 :unknown)
+       ;;            (type-case type-arg2 :unknown-scalar)))
+       ;;  (retok (type-unknown-scalar)))
```
gjurgensen: Remove commented-out code.

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1894,56 +1908,138 @@
...
-                     (retok (type-integer-promote type-arg1 ienv))))
+      ((:shl :shr) (b* (((when (or (type-case type-arg1 :unknown)
+                                   (type-case type-arg1 :unknown-scalar)
+                                   (type-case type-arg1 :unknown-arithmetic)))
+                         (if (or (type-case type-arg1 :unknown)
+                                 (type-case type-arg1 :unknown-scalar)
+                                 (type-case type-arg1 :unknown-arithmetic)
+                                 (type-integerp type-arg2))
+                             (retok (type-unknown-arithmetic))
+                           (reterr msg)))
```
gjurgensen: This `if` condition is always true.

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1894,56 +1908,138 @@
...
-               ((and (type-case type1 :pointer)
-                     (type-integerp type2))
-                (retok type1))
-               ((and (type-case type1 :pointer)
-                     (type-case type2 :pointer))
-                (retok (type-unknown-scalar)))
+               ((or (type-case type1 :unknown)         ; could be pointer
+                    (type-case type1 :unknown-scalar)) ; or arithmetic
+                (cond ((or (type-case type2 :unknown)
+                           ;; (type-case type2 :unknow-scalar)
```
gjurgensen: I think this commented-out disjunct should be uncommented.

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1894,56 +1908,138 @@
...
+                       (retok type2))
+                      ((type-arithmeticp type2)
+                       (retok (type-uaconvert type1 type2 ienv)))
+                      (t (reterr msg))))
+               ((type-arithmeticp type1) ; not integer, checked above
+                (cond ((type-arithmeticp type2)
+                       (retok (type-uaconvert type1 type2 ienv)))
+                      ((or (type-case type2 :unknown)
+                           (type-case type2 :unknown-scalar))
+                       ;; type2 cannot ba pointer, in valid code:
```
gjurgensen: Typo: ba -> be

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1894,56 +1908,138 @@
...
+                       (retok (type-unknown-arithmetic)))
+                      (t (reterr msg))))
+               ((type-case type1 :pointer)
+                (cond ((or (type-case type2 :unknown)
+                           (type-case type2 :unknown-scalar)
+                           (type-integerp type2))
```
gjurgensen: Should this allow `(type-case type2 :unknown-arithmetic)` as well? I would think so, since we allow `type2` to be a known integer type.

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1894,56 +1908,138 @@
...
+      ((:mul :div :rem)
+       (b* (((when (and (or (type-case type-arg1 :unknown)
+                            (type-case type-arg1 :unknown-scalar))
+                        (or (type-case type-arg2 :unknown)
+                            (type-case type-arg2 :unknown-scalar))))
+             (retok (type-unknown-arithmetic)))
+            ((unless (and (type-arithmeticp type-arg1)
+                          (type-arithmeticp type-arg2)))
+             (reterr msg)))
+         (retok (type-uaconvert type-arg1 type-arg2 ienv))))
```
gjurgensen: This doesn't seem to allow one type to be unknown/unknown-scalar and the other arithmetic.

books/kestrel/c/syntax/types.lisp
```diff
@@ -1279,9 +1287,11 @@
         (cond
          ((or (type-case type1 :unknown)
               (type-case type1 :unknown-scalar)
+              (type-case type1 :unknown-arithmetic)
               (type-case type2 :unknown)
-              (type-case type2 :unknown-scalar))
-          (type-unknown-scalar))
+              (type-case type2 :unknown-scalar)
+              (type-case type2 :unknown-arithmetic))
+          (type-unknown-arithmetic))
```
gjurgensen: I think these checks for unknowns need to be moved to the top of the function, before considering the float types. For instance, if we have one type that is `double` and the other that is `:unknown` (or any of the other unknown variants), the current function would give `double` as the result, but the correct answer is `unknown-arithmetic` (if the `:unknown` type represented a `float`, the conversion would be a `double`; if it represented a `long double`, the conversion be a `long double`).

books/kestrel/c/syntax/types.lisp
gjurgensen: I think this can be tightened to `type-unknown-arithmetic`.

books/kestrel/c/syntax/validator.lisp
```diff
@@ -1985,13 +2088,32 @@
...
+             (if (or (type-case type-arg2 :unknown)
+                     (type-case type-arg2 :unknown-scalar)
+                     (type-case type-arg2 :unknown-arithmetic)
+                     (type-integerp type-arg2))
+                 (retok (type-unknown-arithmetic))
+               (reterr msg)))
+            ((when (or (type-case type-arg2 :unknown)
+                       (type-case type-arg2 :unknown-scalar)
+                       (type-case type-arg2 :unknown-arithmetic)
+                       (type-integerp type-arg2)))
```
gjurgensen: I don't think we want this `type-integerp` disjunct here. Otherwise, this case will always fire, and the `retok` at the end will be dead.

gjurgensen (changes requested):

acoglio:
As discussed on Zulip, I made an overall simplification to `valid-binary` that should take care of the remaining issues.

gjurgensen:
I think there is still that issue I commented on earlier in type-uaconvert.

acoglio:
Missed that. Should be in now.

gjurgensen (approved):

PR #1951, opened 2026-06-05
[fty] add :ctor-style to defmake-self

bendyarm:


books/kestrel/fty/defmake-self-doc.lisp
```diff
@@ -63,6 +64,22 @@
...
        necessary type dependencies."))
 
+    (xdoc::desc
+     "@(':ctor-style')"
+     (xdoc::p
+      "Controls the constructor form emitted for product and tagged-sum types.
+       On @(':positional') (the default), the form uses the by-position
+       constructor function, e.g. @('(foo arg1 arg2)').  On @(':maker'), it uses
+       the keyword constructor macro, e.g. @('(make-foo :field1 arg1 :field2
+       arg2)'), which is more legible when inspecting values interactively.")
```
gjurgensen: Nitpick: the claim that the maker forms are more legible strikes me as a bit too opinionated for the documentation. I might just leave that last clause out, or soften it too "_may be_ more legible".

gjurgensen:
It would be good to add a test of `:ctor-style :maker` at some point to `defmake-self-tests.lisp`.

gjurgensen:
One potential issue. `defflexsum` has a `:no-ctor-macros` option, which is exposed to `defprod` and `deftagsum` and which prevents the constructor macros from being generated. Unfortunately this option doesn't seem to be documented, but you can see it is allow by the following example:
```lisp
(in-package "FTY")

(include-book "centaur/fty/top" :dir :system)
(include-book "std/testing/must-fail" :dir :system)

(defprod myprod
  :no-ctor-macros t
  ((field1 natp)
   (field2 stringp)))

; The positional constructor function exists.
(assert-event (myprod-p (myprod 3 "hello")))

; The make-myprod constructor macro was NOT generated.
(acl2::must-fail
 (make-event '(make-myprod :field1 3 :field2 "hello")))
```

I figure that `defmake-self` should either test that the constructors exist, or we should just document that this is a gap in the current implementation. (The latter seems reasonable to me since almost no one seems to use this option in practice.)

gjurgensen (approved):

gjurgensen:
At some point in the future, we could also think about whether we want to do something special for keyword arguments with default values. I.e., if the argument is the default value, we could choose to omit that argument. Not at all necessary for this PR, just something to consider.

bendyarm:
Resolved all the pending comments and added `:universal` argument.  Re-requesting review.

books/kestrel/fty/defmake-self.lisp
```diff
@@ -716,17 +772,344 @@
...
+(define defmake-self-universal-node-p (flex)
+  :returns (yes booleanp)
+  :short "Recognize an AST ``node'' type that the @(':universal') dispatcher
+          can serialize directly."
+  (and (flexsum-p flex)
+       (let ((typemacro (flexsum->typemacro flex)))
+         (if (or (eq typemacro 'defprod)
+                 (eq typemacro 'deftagsum))
+             t
+           nil))))
```
gjurgensen: Nitpick: this `if` could be replaced with the condition.

books/kestrel/fty/defmake-self.lisp
```diff
@@ -716,17 +772,344 @@
...
+  :short "The constructor macro for the first member type
+          declared with @(':no-ctor-macros')."
+  (b* (((when (endp members)) nil)
+       (flex (car members))
+       ;; Only products and tagged sums have constructor macros (the same
+       ;; types for which :maker emits them).
+       ((unless (defmake-self-universal-node-p flex))
+        (defmake-self-members-first-no-ctor-macro (cdr members)))
+       ((unless (flexsum-p flex))
+        (defmake-self-members-first-no-ctor-macro (cdr members)))
```
gjurgensen: `defmake-self-universal-node-p` implies `flexsum-p` . Perhaps we should only check the latter?

gjurgensen (approved):
LGTM.

books/kestrel/fty/defmake-self.lisp
bendyarm: Thanks for the observation.  Yes, that was defensive, 
since the return type of OR is not as clear as T or NIL.  
But the content of the OR is simple and the book
looks better and certifies with the change.

books/kestrel/fty/defmake-self.lisp
bendyarm: I think you mean we should only check the former, which is more restrictive.  Agreed.  This was something to simplify the guard proof.  I added a guard hint and removed the extra `b*` clause.

PR #1952, opened 2026-06-07
[unicode] add some doc

bendyarm:


solswords (approved):
I'm not that familiar with this library but this seems accurate as far as I'm aware.

PR #1953, opened 2026-06-09
Cleanup quicklisp

westfold:
Make usocket package available directly (was loaded by hunchentoot)
Add copyright notices to zippy and dexador and add them to top.lisp.


gjurgensen:
This is showing unrelated changes by AC to `kestrel/c/syntax`, but I can't figure out why.

Edit: seems good now after merging in the most recent `testing-kestrel`.

books/quicklisp/usocket.lisp
```diff
@@ -0,0 +1,29 @@
...
+;   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+;   DEALINGS IN THE SOFTWARE.
+;
+
+(in-package "ACL2")
+(include-book "base")
+
+(defttag :quicklisp)
+
+(include-raw "usocket-raw.lsp" :host-readtable t)
```
gjurgensen: We need a `depends-on` comment here:
```lisp
; (depends-on "usocket-raw.lsp")
```
https://acl2.org/doc/?topic=BUILD____RAW-LISP-AND-OTHER-DEPENDENCIES

westfold: I think the Coglio commits occur here because I used an ACL2 clone that started from master instead of testing-kestrel, and then did a merge request to testing-kestrel. Not sure how to fix this without starting again, or if it actually matters if AC's commits have already been merged into master.

gjurgensen: Alright I merged in `testing-kestrel` and now the `diff` only shows the expected files.

gjurgensen (approved):
LGTM as long as the build succeeds. Perhaps @bendyarm could review as well?

bendyarm:
Looks fine.  Minor nit: the copyright notices are rather verbose.  I suggest rebase and merge.

ericwhitmansmith:
It would be good to use a BSD license with a non-verbose header.  See example at the top of this file:

https://github.com/acl2/acl2/blob/master/books/kestrel/arithmetic-light/abs.lisp

Also, please include authorship info if not already present.

PR #1954, opened 2026-06-15
JSONRPC - An interface for handling JSON-RPC 2.0 requests via sockets

ltmquan:
Two main functionalities:
 - Process a JSON-RPC 2.0 request or batch of requests from a file and writes the result to an output file
 - Start a JSON-RPC 2.0 server over a specified port and listen to a stream of requests or batches of requests, and returns corresponding results
 
 To be reviewed by @gjurgensen.


books/kestrel/jsonrpc/package.lsp
```diff
@@ -0,0 +1,63 @@
...
+                            acl2::int
+                            acl2::bool
+                            acl2::nat-to-string
+                            acl2::parse-string-as-json
+                            acl2::string-upcase
+                            acl2::trans-eval-error-triple
+                            acl2::read-file-into-character-list-safe
+                            acl2::write-strings-to-file
+                            acl2::defttag
+                            acl2::include-raw)))
```
gjurgensen: Nitpick: These symbols don't need the `acl2::` prefix since the package is declared in the ACL2 package (see the `in-package` at the top).

books/kestrel/jsonrpc/socket-example.lisp
```diff
@@ -0,0 +1,72 @@
...
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; looks up a member by name in a member-list, returns nil if not found
+(define find-member-value ((name stringp) (members member-listp))
+  :returns (val valuep)
+  (cond ((endp members) (value-null))
+        ((equal (member->name (car members)) name)
+         (member->value (car members)))
+        (t (find-member-value name (cdr members)))))
```
gjurgensen: Is this functionality available via one of the functions in the `json/operations.lisp` book?

books/kestrel/jsonrpc/socket.lisp
```diff
@@ -0,0 +1,63 @@
...
+  <p>Usage:</p>
+
+  @({
+    (run-jsonrpc-server 7070 state)
+  })
+
+  <p>Messages must be delimited by a newline character.</p>
+
+  <p>This function does not return under normal operation.  It can be
+  interrupted with a SIGINT (Ctrl-C).</p>")
```
gjurgensen: You should probably mention in the documentation that the port is bound to `0.0.0.0` and therefore accepts connections on every network interface (as opposed to binding to to `127.0.0.1`/`localhost` and therefore only accepting connections originating from the same machine). Eventually we might consider making this configurable, but for now it would suffice just to document the behavior.

books/kestrel/jsonrpc/parse-rpc.lisp
```diff
@@ -0,0 +1,156 @@
...
+; returns:
+;  - has-id: true if the "id" field is present
+;  - is-valid: true if either the "id" field is not present or it is present and
+; of the correct type
+;  - id: the id
+(define parse-rpc-id ((obj valuep))
+  :short "Extract the @('id') field from a JSON-RPC request object."
+  :guard (equal (value-kind obj) :object)
+  :returns (mv (has-id booleanp) (is-valid booleanp) (id idp))
+  (b* ((id-val? (object-member-value? "id" obj))
```
gjurgensen: The `object-member-value?` function raises a hard error when there are multiple members of the same name in the object. Perhaps you should use `object-member-value` to get all values, and then return a structured parse error if there is more than one value for the member.

The same comment applies to other uses of `object-member-value?`.

books/kestrel/jsonrpc/json-to-string.lisp
```diff
@@ -0,0 +1,115 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defxdoc+ json-to-string
+  :parents (jsonrpc)
+  :short "Serializing JSON values to strings."
+  :long "<p>These functions serialize @(see valuep) objects to JSON strings
+  suitable for writing to files or sending over a transport. The main function
+  is @(see value-to-json-string). String values are properly escaped per
+  RFC 4627. Rational numbers that are integers are printed without a decimal
+  point; non-integer rationals are printed as @('numerator/denominator').</p>"
```
gjurgensen: Fractions are not valid JSON numbers. They would need to be printed as decimals. For now, we can throw a hard error and document this limitation. But we shouldn't silently emit invalid JSON.

books/kestrel/jsonrpc/json-to-string.lisp
```diff
@@ -0,0 +1,115 @@
...
+
+(define json-escape-char ((ch characterp))
+  :returns (s stringp)
+  (case ch
+    (#\" "\\\"")
+    (#\\ "\\\\")
+    (#\Newline "\\n")
+    (#\Return "\\r")
+    (#\Tab "\\t")
+    (otherwise (string ch))))
```
gjurgensen, line 43: I don't think this escapes everything that would actually need to be escaped. Maybe just add a comment to this effect for now.

books/kestrel/jsonrpc/socket-example.lisp
```diff
@@ -0,0 +1,72 @@
...
+;
+; Author: Quan Luu (quan.luu@kestrel.edu)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Socket server test.
+;
+; Load this file, then start the server:
+;
+;   (include-book "kestrel/jsonrpc/socket-tests" :dir :system)
```
gjurgensen: This book doesn't seem to exist.

books/kestrel/jsonrpc/socket.lisp
```diff
@@ -0,0 +1,63 @@
...
+
+(include-book "process-rpc")
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defxdoc run-jsonrpc-server
+  :parents (jsonrpc)
+  :short "Start a TCP socket server that handles JSON-RPC 2.0 requests."
+  :long "<p>@('run-jsonrpc-server') opens a TCP server socket on the given
+  port and enters an accept loop.  For each incoming connection it reads
```
gjurgensen: I think this is slightly misleading. It only accepts *one* connection, at which point it enters a loop handling messages from that single connection.

books/kestrel/jsonrpc/top.lisp
```diff
@@ -0,0 +1,162 @@
...
+
+  @({
+    (process-json-rpc-file \"request.json\" \"response.json\" state)
+  })
+
+  <p><b>Socket transport.</b> The entry point is @(see run-jsonrpc-server).
+  It opens a TCP server socket on the given port and loops accepting
+  connections.  Each connection may carry multiple request/response exchanges
+  (keep-alive) and is closed when the client disconnects.  Messages are
+  framed by JSON structure — pretty-printed multi-line JSON is accepted.</p>
```
gjurgensen: I don't think this is true. Multi-line JSON messages aren't accepted because we use newlines as the message delimiter. We should be clear about that in the documentation.

gjurgensen (commented):
It seems `socket-example.lisp` and `subtract-example.lisp` are basically duplicates. I imagine one of them should be removed?

gjurgensen:
Nitpick: some files have "trailing whitespace" and some do not end with a newline. See the "Whitespace and Special Characters" section of `:doc best-practices`. https://acl2.org/doc/?topic=ACL2____BEST-PRACTICES

gjurgensen:
Also, I'm thinking that it is actually too permissive to allow calling anything in the `JSONRPC` package. It would be better for the dispatch functions to take an `allowed-methods` list of all the callable methods. As a special case, we could allow something like `:any` to handle the unrestricted case, but I'd rather that not be the default.

gjurgensen:
I think this library also needs to be added to `books/kestrel/top.lisp` and `books/kestrel/top-doc.lisp`. (The latter is especially important; without it I don't think the XDOC will be available in the manual.)

books/kestrel/jsonrpc/socket.lisp
gjurgensen: Thinking about this more, I think we should implement this option now. It would be good to treat the local-only (i.e. `127.0.0.1`/`localhost`) option as the default, with the ability to opt-in to cross-machine connections.

books/kestrel/jsonrpc/socket-example.lisp
ltmquan: `json/operations.lisp` has `object-member-value` and its variants but it takes a `valuep` as the second argument instead of `member-listp`

gjurgensen: OK, that's fine to keep it then. One thing to think about is whether you need to check for multiple fields of the same name.

books/kestrel/jsonrpc/socket.lisp
ltmquan: I had previously allowed multiple connections before, since it's easier for testing. I changed it to be one connection only but didn't update the doc. I'm actually a bit unsure about what we should allow, one or multiple connections. What do you think?

gjurgensen: What would "multiple connections" look like? Would we go back to listening for new connections after one connection has finished, or would it be possible to somehow support multiple connections in parallel?

books/kestrel/jsonrpc/socket.lisp
ltmquan: I've updated the `run-jsonrpc-server` function to now take an additional parameter that specifies the network interface. If the parameter is `nil`, then it defaults to 127.0.0.1.

books/kestrel/jsonrpc/socket.lisp
ltmquan: It would just be sequential -- so we'd go back to listening for new connections after one connection has finished. I was thinking if we wanted to open a port, we could keep it open for several clients to connect to, so whenever one disconnects, we don't have to start a new server and can listen for the next client.

ltmquan:
- removed `socket-example.lisp`.
 - fixed the whitespace and newline issue.
 - added a `allowed-methods` to the `run-jsonrpc-server`
 - added `jsonrpc` to `kestrel/top.lisp` and `kestrel/top-doc.lisp`

books/kestrel/jsonrpc/socket.lisp
gjurgensen: That sounds good.

gjurgensen:
Is there any reason that `json-response.txt` needs to be committed? I'm thinking perhaps it shouldn't be, and that it should be added to a local `.gitignore`.

books/kestrel/jsonrpc/parse-rpc.lisp
```diff
@@ -0,0 +1,171 @@
...
+        (mv (id-null)
+            (request+error-error
+             (make-invalid-request-error
+              "Missing or invalid \"jsonrpc\" field; must be \"2.0\""))))
+       (method-val-list (object-member-values "method" val))
+       ((unless (equal (len method-val-list) 1))
+        (mv (id-null)
+            (request+error-error
+             (make-invalid-request-error
+              "Duplicate \"member\" field"))))
```
gjurgensen: Presumably this should be `"method"`, not `"member"`.

ltmquan:
I removed `json-request.txt` and `json-response.txt`

gjurgensen:
I think `json-request.txt` was fine (although you might rename it to something like `example-request.json`). It is just `json-response.txt` that I think shouldn't be included, since it is not necessary to run the test and generated automatically. And I think that it should be added to `.gitignore` to prevent people from accidentally committing it after running a test.

ltmquan:
I just added back `example-request.json`.

gjurgensen (approved):

PR #1957, opened 2026-06-20
x86isa: Fix 5 real model bugs in asmtest (CMPPS/PD/SS/SD, shifts, PSLLDQ/PSRLDQ, LZCNT, VPOR)

yaso9:
# x86isa asmtest: Fix real model bugs (1, 3, 5, 6, 8)

## Summary

This PR fixes **five** real model bugs in the x86isa formal model, where
the ACL2 model disagreed with the Intel SDM behavior described in the
[Felix Cloutier mirror of the Intel 64 and IA-32 Architectures Software
Developer's Manual, December 2023][sdm]. Each fix is in its own commit
with a focused message.

[sdm]: https://github.com/fay59/cloutier

The asmtest framework runs each instruction snippet against both the
ACL2 model and the host x86 implementation and reports any
mismatches. This PR was driven by `BUG_SOURCE_REPORT.md` (in the
`tools/execution/asmtest/` tree), which enumerated 10 issues found in
asmtest runs. Six of these were classified by the report as **real model
bugs** (where the model disagrees with the Intel SDM); this PR fixes
five of them. The remaining issues are intentionally out of scope per
the task: implementation mismatches in spec-undefined regions (Bug 2
— SQRT QNaN sign, Bug 4 — CF for count ≥ operand size, Bug 9 — rcl/rcr
initial CF) and a test framework defect (Bug 10 — CMOV snippets don't
initialize R8). Bug 7 (CVTPD2PS / CVTSD2SS over-saturation) was
investigated but requires deeper RTL audit; see "Unresolved" section.

### Bug 1 — CMPPS / CMPPD / CMPSS / CMPSD predicate imm8 mask

**Symptom:** For predicates 4–7 (NEQ, NLT, NLE, ORD), the model
produced the inverse of the correct result. Predicates 0–3 (EQ, LT,
LE, UNORD) worked correctly.

**Justification (SDM deviation):** The Intel SDM (`cmppd:18`)
specifies: *"Compare packed double precision floating-point values in
xmm2/m128 and xmm1 using **bits 2:0 of imm8 as a comparison
predicate**."* (`cmpss:73` adds: *"Bits 3 through 7 of the immediate
are reserved."*) Bit 2 of the predicate encodes the "Not-*" half of
predicates 4–7 (NEQ, NLT, NLE, ORD). The model was reading only the
low 2 bits via `(n02 imm)`, which collapsed `0b1xx → 0b0xx` and
treated NEQ as EQ, NLT as LT, NLE as LE, and ORD as UNORD — the
logical complements of the correct comparison results.

**Fix:** Replaced `(n02 imm)` with `(n03 imm)` at the four predicate
read sites in `books/projects/x86isa/machine/instructions/fp/logical.lisp`
(one in CMPPS, one in CMPPD, two in CMPSS/CMPSD — one per operand).

**File:** `books/projects/x86isa/machine/instructions/fp/logical.lisp`
**Commit:** `28e7d33072`

**Tests fixed:**
- `testgen_1332` (CMPPS) — 1994 mismatches → 0
- `testgen_1345` (CMPSS) — 0
- `testgen_1347` (CMPPD) — 0
- `testgen_1351` (CMPSD) — 0

### Bug 3 — shl/sar/shr/shld/shrd and rcl/rcr/rol/ror flags recomputed when count = 0

**Symptom:** When the effective count was 0, the model computed PF
(and sometimes ZF) from the unchanged result instead of preserving
RFLAGS. ROL and ROR additionally marked OF as undefined for a
zero-bit rotate.

**Justification (SDM deviation):** The Intel SDM
(`sal:sar:shl:shr:551`, "Flags Affected") states: *"The SF, ZF, and PF
flags are set according to the result. **If the count is 0, the flags
are not affected.** For a non-zero count, the AF flag is undefined."*
SHLD and SHRD have identical wording
(`shld:119`/`shrd:119`). The rotate SDM
(`rcl:rcr:rol:ror:584`) states: *"For RCL and RCR instructions, a
zero-bit rotate does nothing, i.e., affects no flags. For ROL and
ROR instructions, if the masked count is 0, the flags are not
affected."*

**Fix:** Two changes, all per the SDM:
  1. Fixed the `src = 0` branch in `rol-spec-gen` and `ror-spec-gen`
     to return `(mv input-rflags 0)` (no flags undefined). The prior
     `(mv input-rflags (!rflagsBits->of 1 0))` was marking OF as
     undefined, which contradicts the SDM's "the flags are not
     affected" for ROL/ROR with count 0.
  2. Removed the `count = 0` short-circuit at the dispatch sites in
     `x86-sal/sar/shl/shr/rcl/rcr/rol/ror` and `x86-shld/shrd`. The
     spec functions in `shifts-spec.lisp` (and now `rol-spec-gen` /
     `ror-spec-gen`) already return `(mv dst input-rflags 0)` for
     `count = 0`. With the ROL/ROR spec now correct, the
     dispatch-site short-circuit was redundant (and incidentally
     masked the ROL/ROR spec bug from being visible at the dispatch
     site). Per the review feedback, the flag-preservation logic
     belongs in the spec functions, not at the dispatch site.

**Files:**
- `books/projects/x86isa/machine/instructions/rotate-and-shift.lisp`
- `books/projects/x86isa/machine/instructions/rotates-spec.lisp`

**Commit:** `be3dfe0b36`

**Tests affected (10 total, plus ROL/ROR count=0):**
- `testgen_429` (SHL r8b), `testgen_431` (SHR r8b),
  `testgen_433` (SHL r64), `testgen_435` (SHL r32),
  `testgen_445` (SHR r64), `testgen_447` (SHR r32),
  `testgen_457` (SAR r8b), `testgen_459` (SAR r64),
  `testgen_1642` (SHRD), `testgen_1646` (SHLD)

The fix is correct per the SDM. Remaining mismatches in these tests are
cascade effects: prior instructions in the asmtest loop set PF = 1 in
flags, and the fix correctly preserves that PF = 1 per SDM, while
some prior instructions in the model compute PF differently from
the host. Those upstream mismatches are out of scope for this PR.

ROL and ROR do not have a dedicated "count = 0" snippet in the
asmtest corpus, but the full-suite ROL/ROR snippets now report
zero mismatches after the spec fix:
- `testgen_365` (ROL r8b, imm8), `testgen_367` (ROL r64, imm8),
  `testgen_377` (ROR r8b, imm8), `testgen_378` (ROR r64, imm8),
  `testgen_397` (ROR r8b, CL), `testgen_399` (ROR r64, CL) — all 0.

### Bug 5 — PSLLDQ / PSRLDQ count ≥ 16 returns unshifted operand

**Symptom:** When the shift count was ≥ 16, the model returned the
unshifted XMM operand; the host returned all zeros.

**Justification (SDM deviation):** The Intel SDM (`pslldq:84`,
"Description") states: *"If the value specified by the count operand
is greater than 15, the destination operand is set to all 0s."* The
SDM pseudocode explicitly sets `TEMP := 16` in this case so that the
subsequent `DEST[127:0] := SRC[127:0] << (TEMP * 8)` (or `>>`) produces
zero by virtue of a 128-bit shift. The model instead clamped count to
0, making the instruction a no-op rather than a zero-store.

**Fix:** Changed the clamp in `pshift.lisp` from `0` to `16`, matching
the SDM pseudocode.

**File:** `books/projects/x86isa/machine/instructions/pshift.lisp`
**Commit:** `1b905154b1`

**Tests fixed:**
- `testgen_666` (PSRLDQ) — 1870 mismatches → 0
- `testgen_667` (PSLLDQ) — 0

### Bug 6 — LZCNT helper returns operand-size-bits for every input

**Symptom:** For every input, the model returned 0x40 = 64 (the
operand size in bits), regardless of the actual leading-zero count.

**Justification (SDM deviation):** The Intel SDM (`lzcnt:49`,
"Description") states: *"Counts the number of leading most significant
zero bits in a source operand (second operand) returning the result
into a destination (first operand)."* `lzcnt:51` explicitly contrasts
with BSR: *"LZCNT differs from BSR. **For example, LZCNT will produce
the operand size when the input operand is zero.**"* The model's
recursive `lzcnt` helper had a base case `(when (equal i+1 0)) bits)`
that fired when `i+1 = 0`, returning `bits` (the operand-size in
bits). The call site passed an initial `i+1` of `0`, so the base case
fired immediately and the helper never examined any bits of the
operand.

**Fix:** Changed the call site in `bit.lisp` from
`(lzcnt (ash operand-size 3) 0 source)` to
`(lzcnt (ash operand-size 3) (ash operand-size 3) source)`. The
initial `i+1` now equals `bits` so the first iteration examines bit
`(bits − 1)`, the MSB, and decrements toward 0 over subsequent
recursive calls. The base case then fires correctly only after the
helper has walked all bits without finding a 1.

**File:** `books/projects/x86isa/machine/instructions/bit.lisp`
**Commit:** `b1d82e368c`

**Tests fixed:**
- `testgen_2025` (LZCNT) — 1000 mismatches → 0

### Bug 8 — VPOR dispatch uses *OP-XOR* instead of *OP-OR*

**Symptom:** The model computed XMM9 ^ XMM10 (XOR) rather than
XMM9 | XMM10 (OR) for every VEX-encoded VPOR invocation. Every
mismatch showed the XOR pattern.

**Justification (SDM deviation):** The Intel SDM (`por:95`,
"Description") states: *"Performs a bitwise logical OR operation on
the source operand (second operand) and the destination operand (first
operand) and stores the result in the destination operand. **Each
bit of the result is set to 1 if either or both of the corresponding
bits of the first and second operands are 1; otherwise, it is set to
0.**"* The `por:30` opcode entry confirms `VPOR xmm1, xmm2,
xmm3/m128` is *"Bitwise OR of xmm2/m128 and xmm3"*. The shared
logical body dispatches via `case operation` on constants from
`arith-and-logic-spec.lisp` where `*OP-OR* = 1` and `*OP-XOR* = 5`.
The two VEX-encoded VPOR dispatch entries (128-bit and 256-bit
variants) were passing `5` (XOR) instead of `1` (OR).

**Fix:** Changed `(operation . #x5)` to `(operation . #x1)` in both
entries at lines 12320–12321 and 12330–12331 of `inst-listing.lisp`.

**File:** `books/projects/x86isa/machine/inst-listing.lisp`
**Commit:** `325c2c5d23`

**Tests fixed:**
- `testgen_2555` (VPOR) — 1000 mismatches → 0

## Test verification

Each fix was certified by running:

```
~/acl22/books/build/cert.pl --acl2 ~/acl22/saved_acl2 -j 8 \
    ~/acl22/books/projects/x86isa/tools/execution/top \
    ~/acl22/books/projects/x86isa/tools/execution/asmtest/asmtest
```

in `books/projects/x86isa/tools/execution/asmtest/` after `make clean`,
followed by rebuilding `check-test.lx86cl64` and running the targeted
failing test with `./check-test -- <testname>`. Confirmed results
after all five fixes:

| Bug | Tests fixed | Mismatches after fix |
|-----|-------------|----------------------:|
| 1   | `testgen_1332`, `testgen_1345`, `testgen_1347`, `testgen_1351` | 0 / 0 / 0 / 0 |
| 3   | `testgen_429`, `431`, `433`, `435`, `445`, `447`, `457`, `459`, `1642`, `1646`; plus ROL/ROR `testgen_365`, `367`, `377`, `378`, `397`, `399` | shifts/rotates: count=0 mismatches eliminated; cascade mismatches remain. ROL/ROR: 0 |
| 5   | `testgen_666`, `testgen_667` | 0 / 0 |
| 6   | `testgen_2025` | 0 |
| 8   | `testgen_2555` | 0 |

## Out of scope (per task definition)

Per the task description, the following were intentionally not addressed:

- **Bug 2** — SQRTPS/PD/SS/SD QNaN sign preservation (implementation
  mismatch; SDM leaves the QNaN sign for SQRT(negative) undefined and
  classifies this as a real implementation mismatch, not a real model
  bug).
- **Bug 4** — CF for shl/shr count = operand size (implementation
  mismatch; SDM marks CF as undefined for count ≥ operand size).
- **Bug 9** — rcl/rcr initial CF (implementation mismatch; CF is part
  of the rotation per SDM but the host starts with CF = 0).
- **Bug 10** — CMOV snippets don't initialize R8 (test framework defect).

## Unresolved

**Bug 7** — CVTPD2PS / CVTSD2SS over-saturation. The model returns
`0xFF800000` (-inf) for in-range finite doubles where the expected
result is a specific SP value (e.g., `0xF55C1794` for input
`0xc6ab82f277096367` which decodes to -2.79e32). Direct invocation
of `rtl::sse-post-comp` on the failing input returns the correct SP
value; manual REPL testing of the individual RTL functions shows
correct behavior. The bug appears in the composed call chain via
`ec-call` through `rat-round` and `rat-to-fp` in execution mode.
A defensive fix in `sse-cvt-fp1-to-fp2` was attempted (calling
`nencode` directly when `overflowp` is false and `nrepp` is true)
but did not change the asmtest result. This likely involves ACL2's
`:exec` semantics for `ec-call`-marked functions and guard violations
in the compiled form of `rat-to-fp` or its callers. Deeper RTL audit
is required.

## Checklist

- [x] Each fix is in its own commit
- [x] Commit messages are descriptive and reference the SDM
- [x] Each fix was tested before commit
- [x] All commits compile cleanly via cert.pl

Reviewers: @ericwhitmansmith, @acoglio

### DIFF
--- a/books/projects/x86isa/machine/inst-listing.lisp
+++ b/books/projects/x86isa/machine/inst-listing.lisp
@@ -12318,7 +12318,7 @@
                :OP2 '(H X)
                :OP3 '(W X))
           '(x86-vandp?/vandnp?/vorp?/vxorp?/vpand/vpandn/vpor/vpxor-vex
-            (operation . #x5))
+            (operation . #.*OP-OR*))
           '((:EX (CHK-EXC :TYPE-4 (:AVX)))))
     (INST "VPOR"
           (OP :OP #xFEB
@@ -12328,7 +12328,7 @@
                :OP2 '(H X)
                :OP3 '(W X))
           '(x86-vandp?/vandnp?/vorp?/vxorp?/vpand/vpandn/vpor/vpxor-vex
-            (operation . #x5))
+            (operation . #.*OP-OR*))
           '((:EX (CHK-EXC :TYPE-4 (:AVX2)))))
     (INST "VPORD"
           (OP :OP #xFEB
--- a/books/projects/x86isa/machine/instructions/rotate-and-shift.lisp
+++ b/books/projects/x86isa/machine/instructions/rotate-and-shift.lisp
@@ -207,35 +207,35 @@
        ;; Computing the flags and the result:
        (input-rflags (the (unsigned-byte 32) (rflags x86)))
 
-       ((mv result
-            (the (unsigned-byte 32) output-rflags)
-            (the (unsigned-byte 32) undefined-flags))
-        (case reg
-          (0
-           ;; ROL
-           (rol-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
-          (1
-           ;; ROR
-           (ror-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
-          (2
-           ;; RCL
-           (rcl-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
-          (3
-           ;; RCR
-           (rcr-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
-          (4
-           ;; SAL/SHL
-           (sal/shl-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
-          (5
-           ;; SHR
-           (shr-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
-          (7
-           ;; SAR
-           (sar-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
-          ;; The guard for this function will ensure that we don't
-          ;; reach here.
-          (otherwise
-           (mv 0 0 0))))
+      ((mv result
+           (the (unsigned-byte 32) output-rflags)
+           (the (unsigned-byte 32) undefined-flags))
+       (case reg
+         (0
+          ;; ROL
+          (rol-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
+         (1
+          ;; ROR
+          (ror-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
+         (2
+          ;; RCL
+          (rcl-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
+         (3
+          ;; RCR
+          (rcr-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
+         (4
+          ;; SAL/SHL
+          (sal/shl-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
+         (5
+          ;; SHR
+          (shr-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
+         (7
+          ;; SAR
+          (sar-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
+         ;; The guard for this function will ensure that we don't
+         ;; reach here.
+         (otherwise
+          (mv 0 0 0))))
 
        ;; Update the x86 state:
 
@@ -385,21 +385,21 @@
 
        (input-rflags (the (unsigned-byte 32) (rflags x86)))
 
-       ((mv result
-            result-undefined?
-            (the (unsigned-byte 32) output-rflags)
-            (the (unsigned-byte 32) undefined-flags))
-        (case opcode
-          ((#xA4 #xA5) (shld-spec
-                        operand-size dst-value src-value count input-rflags))
-          ((#xAC #xAD) (shrd-spec
-                        operand-size dst-value src-value count input-rflags))
-          (otherwise (mv 0 nil 0 0)))) ; unreachable
+      ((mv result
+           result-undefined?
+           (the (unsigned-byte 32) output-rflags)
+           (the (unsigned-byte 32) undefined-flags))
+       (case opcode
+         ((#xA4 #xA5) (shld-spec
+                       operand-size dst-value src-value count input-rflags))
+         ((#xAC #xAD) (shrd-spec
+                       operand-size dst-value src-value count input-rflags))
+         (otherwise (mv 0 nil 0 0)))) ; unreachable
 
-       ((mv result x86)
-        (if result-undefined?
-            (undef-read x86)
-          (mv result x86)))
+      ((mv result x86)
+       (if result-undefined?
+           (undef-read x86)
+         (mv result x86)))
 
        ;; update the state:
 
--- a/books/kestrel/x86/alt-defs.lisp
+++ b/books/kestrel/x86/alt-defs.lisp
@@ -86,26 +86,28 @@
 
 ;; TODO: Other sizes.
 (defthm ror-spec-64-alt-def
-  (equal (ror-spec-64 x n rflags)
-         (let* ((chopped-n (bvchop 6 n))
-                (result-value (rightrotate 64 n x)) ; note the unchopped n here
+  (equal (ror-spec-64 dst src input-rflags)
+         (let* ((chopped-src (bvchop 6 src))
+                (result-value (rightrotate 64 src dst)) ; note the unchopped n here
                 )
            (mv result-value
                ;; output flags:
-               (if (equal chopped-n 0)
-                   (bvchop 32 rflags)
-                 (if (equal chopped-n 1)
-                     ;; nicer than what the naive definition does?:
-                     (!rflagsbits->cf
-                       (getbit 0 x) ;(getbit 63 (rightrotate 64 1 x))
-                       (!rflagsbits->of
-                         (bitxor (getbit 0 x) ; (getbit 63 (rightrotate 64 1 x))
-                                 (getbit 63 x) ; (getbit 62 (rightrotate 64 1 x))
-                                 )
-                         rflags))
-                   (!rflagsbits->cf (getbit 63 result-value) rflags)))
+               (if (equal chopped-src 0)
+                   (bvchop 32 input-rflags)
+                 (if (equal chopped-src 1)
+                     (let ((cf (getbit 0 dst)) ; (getbit 63 (rightrotate 64 1 x))
+                           (of (bitxor (getbit 0 dst) ; (getbit 63 (rightrotate 64 1 x))
+                                       (getbit 63 dst) ; (getbit 62 (rightrotate 64 1 x))
+                                       )))
+                       (!rflagsbits->cf cf
+                                        (!rflagsbits->of of input-rflags)))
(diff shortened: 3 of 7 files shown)


books/projects/x86isa/machine/instructions/rotate-and-shift.lisp
```diff
@@ -207,35 +207,42 @@
...
-           ;; SAR
-           (sar-spec reg/mem-size reg/mem shift/rotate-by input-rflags))
-          ;; The guard for this function will ensure that we don't
-          ;; reach here.
-          (otherwise
-           (mv 0 0 0))))
+      ((mv result
+           (the (unsigned-byte 32) output-rflags)
+           (the (unsigned-byte 32) undefined-flags))
+       (if (eql shift/rotate-by 0)
```
yaso9: Looking at this code a little closer, I think flag computation should probably be absorbed into the spec functions.

books/projects/x86isa/machine/instructions/rotate-and-shift.lisp
```diff
@@ -385,21 +392,26 @@
...
-
-       ((mv result x86)
-        (if result-undefined?
-            (undef-read x86)
-          (mv result x86)))
+      ((mv result
+           result-undefined?
+           (the (unsigned-byte 32) output-rflags)
+           (the (unsigned-byte 32) undefined-flags))
+       (if (eql count 0)
```
yaso9: Same here regarding absorbing flag preservation into the spec functions.

books/projects/x86isa/machine/inst-listing.lisp
```diff
@@ -12318,7 +12318,7 @@
                :OP2 '(H X)
                :OP3 '(W X))
           '(x86-vandp?/vandnp?/vorp?/vxorp?/vpand/vpandn/vpor/vpxor-vex
-            (operation . #x5))
+            (operation . #x1))
```
acoglio:
```suggestion
            (operation . #.*OP-OR*))
```

books/projects/x86isa/machine/inst-listing.lisp
```diff
@@ -12328,7 +12328,7 @@
...
```
acoglio:
```suggestion
            (operation . #.*OP-OR*))
```

acoglio (approved):
The fixes against the Intel manual look right to me.

I've made two (related) suggestions to use a symbolic constant for greater clarity, as done in other entries of the instruction tables.

yaso9:
Applied the suggested fixes

PR #1959, opened 2026-06-22
REMORA - Adding 8 float operations into `primitives-evaluation.lisp`

ltmquan:
Adding 8 new operations and updated the documentation:
 - The 4 arithmetic functions `prim-float-add/sub/mul/div`: handle special cases consistently with Haskell.
 - The 4 conversion functions `prim-float-truncate/round/ceiling/floor`: also consistent with Haskell
 
@acoglio


ltmquan:
Added all remaining float operations, with some comments:
 - The `prim-float-sqrt` function currently just returns an error, as there's not a good way to represent irrational results right now in ACL2 and Remora has yet to specify a concrete model for handling floats and rounding.
 - The `prim-float-expt` function currently only consider cases where the exponent is a `NaN`, a `posinf`, a `neginf`, or an integer and returns an error if the exponent is a rational but not an integer. This is consistent with how `expt` works in ACL2 and avoids the same issue with irrational results.

books/kestrel/remora/primitives-evaluation.lisp
```diff
@@ -51,7 +52,63 @@
...
+               @(tsee prim-int-lt), @(tsee prim-int-gt),
+               @(tsee prim-int-leq), @(tsee prim-int-geq).")
+    (xdoc::li "@(tsee prim-int-to-float), @(tsee prim-int-to-bool)."))
+   (xdoc::p
+    "The boolean primitives currently implemented are
+     @(tsee prim-bool-not), @(tsee prim-bool-and), @(tsee prim-bool-or),
+     @(tsee prim-bool-eq), @(tsee prim-bool-neq),
+     @(tsee prim-bool-to-int), and @(tsee prim-bool-to-float).")
+   (xdoc::p
+    "The float primitives currently implemented are
```
acoglio: This needs to be updated.

acoglio (approved):
Looks good, just an update needed.

acoglio (approved):

PR #1964, opened 2026-06-27
Remora mono

westfold:
First version of monomorphization. Mainly lacks handling of built-in functions

### DIFF
--- a/books/kestrel/remora/monomorphize.lisp
+++ b/books/kestrel/remora/monomorphize.lisp
@@ -13,9 +13,11 @@
 (include-book "abstract-syntax-trees")
 (include-book "integer-lists")
 (include-book "variable-substitution-operations")
+(include-book "utility-transforms")
 
 (include-book "kestrel/fty/deffold-map" :dir :system)
 (include-book "kestrel/fty/deffold-reduce" :dir :system)
+(include-book "kestrel/fty/string-nat-map" :dir :system)
 (include-book "std/basic/two-nats-measure" :dir :system)
 (include-book "std/alists/put-assoc-equal" :dir :system)
 
@@ -29,7 +31,8 @@
 
 (defxdoc+ monomorphize
   :parents (remora)
-  :short "Monomorphize a Remora program by instantiating @(':cfun') definitions."
+  :short "Monomorphize a Remora program by instantiating
+          @(':cfun') and @(':ifun') definitions."
   :long
   (xdoc::topstring
    (xdoc::p
@@ -40,22 +43,30 @@
      the evaluated nat values and partially evaluating dimensions throughout the
      type and body.")
    (xdoc::p
+    "Index-function definitions are handled the same way: every @(':iapp')
+     call to a known @(':ifun') with non-empty ispace arguments is replaced by
+     a reference to a freshly generated, fully-instantiated @(':val') definition
+     (an @(':ifun') abstracts only ispace parameters, so the instance has no
+     value parameters and is a @(':val') rather than a @(':fun')).")
+   (xdoc::p
     "The two main maps are:")
    (xdoc::ul
     (xdoc::li
      "@('fn-info-map'): a @(tsee fn-info-map), i.e. an alist from @(':cfun')
-      name strings to @(tsee bind+bind-map) pairs, where the @('bind-map')
-      component (a @(tsee bind-map)) maps instance-name strings to the
-      corresponding @(':fun') @(tsee bind) nodes.")
+      and @(':ifun') name strings to @(tsee bind+bind-map) pairs, where the
+      @('bind') component is the definition (:cfun or :ifun) and the
+      @('bind-map') component (a @(tsee bind-map)) maps instance-name strings
+      to the corresponding @(':fun') (for a @(':cfun')) or @(':val') (for an
+      @(':ifun')) @(tsee bind) nodes.")
     (xdoc::li
-     "@('dim-var-map'): a @(tsee string-nat-map), i.e. an alist from ispace
+     "@('dim-var-map'): a @(tsee acl2::string-nat-map), i.e. an omap from ispace
       dimension-variable name strings to their nat values, used to evaluate
       @(':dim') ispace arguments."))
    (xdoc::p
     "A termination fuel parameter bounds how many levels of nested cfun
      instantiation are performed.  When the fuel reaches zero the body of a
      new instance is obtained by dimension-substitution only (via
-     @(tsee partial-eval-dims)) rather than full recursive monomorphization.
+     @(tsee ast-partial-eval-dims)) rather than full recursive monomorphization.
      The top-level entry point @(tsee monomorphize-prog) supplies
      @('(* 10 (expr-count (prog->expr prog)))') as the initial fuel: a margin
      above the expression size, since each cfun instantiation grows the tree
@@ -66,7 +77,7 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-; Fixtypes for the maps used to accumulate cfun/instance information.
+; Fixtypes for the maps used to accumulate cfun/instance and ifun/instance information.
 
 (fty::defalist bind-map
   :short "Fixtype of alists from instance-name strings to @(tsee bind) nodes."
@@ -91,15 +102,6 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-; Extract the string name from a type variable (works for both :atom and :array).
-
-(define type-var-name ((v type-varp))
-  :returns (name stringp)
-  :short "Extract the name string from a type variable."
-  (type-var-case v :atom v.name :array v.name))
-
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-
 ; Map from type-variable name strings to types, used to resolve type arguments.
 ; We use the library @(tsee string-type-map) omap so that the type-variable
 ; substitution operations from @(see variable-substitution-operations) can be
@@ -112,7 +114,7 @@
   :short "Extend @('type-map') with @('tvars[i] -> tys[i]') for each index."
   (if (or (endp tvars) (endp tys))
       (string-type-map-fix type-map)
-    (b* ((name (type-var-name (car tvars)))
+    (b* ((name (type-var->name (car tvars)))
          (ty   (type-fix (car tys))))
       (extend-type-var-map (cdr tvars) (cdr tys)
                            (omap::update name ty
@@ -135,7 +137,7 @@
              :int   "i"
              :float "f")
      :var (b* ((type-map (string-type-map-fix type-map))
-               (name (type-var-name ty.var))
+               (name (type-var->name ty.var))
                (v-ty-pr (omap::assoc name type-map))
                ((unless v-ty-pr) "unbound"))
             (short-name-for-type (cdr v-ty-pr)
@@ -182,36 +184,22 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-; Extract the string name from an ispace variable (works for both :dim and :shape).
-
-(define ispace-var-name ((v ispace-varp))
-  :returns (name stringp)
-  :short "Extract the name string from an ispace variable."
-  (ispace-var-case v :dim v.name :shape v.name))
-
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-
-; Extend ispace-value-map by pairing up ispace vars with nat values.
-
-(fty::defalist string-nat-map
-  :short "Fixtype of alists from strings to nats."
-  :key-type string
-  :val-type natp
-  :pred string-nat-mapp
-  :true-listp t)
+; Extend dim-var-map by pairing up ispace vars with nat values.  We use the
+; library acl2::string-nat-map omap so that this map is handled the same way
+; as the string-type-map used for type arguments.
 
 (define extend-ispace-val-map ((ivars ispace-var-listp)
                                (nats nat-listp)
-                               (dim-var-map string-nat-mapp))
-  :returns (new-dim-var-map string-nat-mapp)
+                               (dim-var-map acl2::string-nat-mapp))
+  :returns (new-dim-var-map acl2::string-nat-mapp)
   :short "Extend @('dim-var-map') with @('ivars[i] -> nats[i]') for each index."
   (if (or (endp ivars) (endp nats))
-      (string-nat-map-fix dim-var-map)
-    (b* ((name (ispace-var-name (car ivars)))
-         (val  (nfix (car nats)))
-         (dim-var-map  (string-nat-map-fix dim-var-map)))
+      (acl2::string-nat-map-fix dim-var-map)
+    (b* ((name (ispace-var->name (car ivars)))
+         (val  (nfix (car nats))))
       (extend-ispace-val-map (cdr ivars) (cdr nats)
-                             (cons (cons name val) dim-var-map)))))
+                             (omap::update name val
+                                           (acl2::string-nat-map-fix dim-var-map))))))
 
 
 (define nat-sum-list ((nats nat-listp))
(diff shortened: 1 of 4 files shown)


books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+
+(local (include-book "std/typed-lists/string-listp" :dir :system))
+(local (include-book "std/omaps/top" :dir :system))
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defxdoc+ monomorphize
+  :parents (remora)
+  :short "Monomorphize a Remora program by instantiating @(':cfun') definitions."
```
acoglio: Should the transformation also monomorphize `:ifun` definitions to `:val` definitions? If that's a future plan, it would be good to mention it as such in this doc topic.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+     @(':cfun') that carries non-empty ispace arguments with a plain @(':app')
+     call to a freshly generated, fully-instantiated @(':fun') definition.
+     The new function is built by binding the @(':cfun')'s ispace parameters to
+     the evaluated nat values and partially evaluating dimensions throughout the
+     type and body.")
+   (xdoc::p
+    "The two main maps are:")
+   (xdoc::ul
+    (xdoc::li
+     "@('fn-info-map'): a @(tsee fn-info-map), i.e. an alist from @(':cfun')
```
acoglio, line 56: In the `bind+bind-map` pair, is the `bind` component the definition of the `:cfun`?

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+  :key-type string
+  :val-type bind+bind-map
+  :pred fn-info-mapp
+  :true-listp t)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Extract the string name from a type variable (works for both :atom and :array).
+
+(define type-var-name ((v type-varp))
```
acoglio: This already exists in `abstract-syntax-structurals.lisp`, and is called `type-var->name`.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+        (str::cat base "/" ints-suffix)
+      (if (null nats)
+          (str::cat base "/" strs-suffix)
+        (str::cat base "/" strs-suffix "/" ints-suffix)))))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Extract the string name from an ispace variable (works for both :dim and :shape).
+
+(define ispace-var-name ((v ispace-varp))
```
acoglio: This already exists in `abstract-syntax-structurals.lisp`, and is called `ispace-var->name`.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+  :returns (name stringp)
+  :short "Extract the name string from an ispace variable."
+  (ispace-var-case v :dim v.name :shape v.name))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Extend ispace-value-map by pairing up ispace vars with nat values.
+
+(fty::defalist string-nat-map
```
acoglio: There is a `string-nat-map` fixtype for omaps (not alists) in `[books]/kestrel/fty/string-nat-map.lisp`. In many libraries including Remora, we normally use the `<keytype>-<valtype>-map` naming scheme for omaps from `<keytype>` to `<valtype>`, and `<keytype>-<valtype>-alist` for alists from `<keytype>` to `<valtype>`. This kind of naming scheme (also used for `-set`, `-option`, `-list`, `-result`) approximates a postfix notation for polymorphic types as in Isabelle, and the names "compose".

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+  (if (or (endp ivars) (endp nats))
+      (string-nat-map-fix dim-var-map)
+    (b* ((name (ispace-var-name (car ivars)))
+         (val  (nfix (car nats)))
+         (dim-var-map  (string-nat-map-fix dim-var-map)))
+      (extend-ispace-val-map (cdr ivars) (cdr nats)
+                             (cons (cons name val) dim-var-map)))))
+
+
+(define nat-sum-list ((nats nat-listp))
```
acoglio: This is already a `nat-list-sum` in `nat-lists.lisp` in this Remora directory.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+  :returns (sum natp)
+  :short "Sum a list of natural numbers."
+  (if (endp nats)
+      0
+    (+ (nfix (car nats)) (nat-sum-list (cdr nats)))))
+
+(define nat-product-list ((nats nat-listp))
```
acoglio: This is already a `nat-list-product` in `nat-lists.lisp` in this directory.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+  :returns (prod natp)
+  :short "Multiply a list of natural numbers."
+  (if (endp nats)
+      1
+    (* (nfix (car nats)) (nat-product-list (cdr nats)))))
+
+(define nat-sub-list ((nats nat-listp))
```
acoglio: There is already a `nat-list-subtraction` in `nat-lists.lisp` in this directory.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+  :short "Collect the variable and constant leaf dims occurring in an ispace tree."
+  :types (dims
+          shapes/ispaces)
+  :result dim-listp
+  :default nil
+  :combine append
+  :override
+  ((dim :var   (list (dim-fix dim)))
+   (dim :const (list (dim-fix dim)))
+   (shape :var (list (dim-var (shape-var->name shape)))))
```
acoglio, line 239: Why turning a shape variable into a dimension variable?

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+  :name ast-leaf-dims)
+
+(local
+ (defthm consp-of-assoc-equal-when-string-nat-mapp
+   (implies (and (string-nat-mapp x) (assoc-equal k x))
+            (consp (assoc-equal k x)))))
+
+(define eval-var-nat ((name stringp) (dim-var-map string-nat-mapp))
```
acoglio: This and the following eval functions are similar to the ones in the evaluation semantics. Maybe at some point we could factor them into their own file/topic, so they can be used for different purposes. I'm not suggesting we do anything yet in this PR though.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+     evaluates each one: a variable through @('dim-var-map'), a constant to its
+     value.  Returns @('(mv t nil)') if any variable was missing, otherwise
+     @('(mv nil nats)')."))
+  (eval-leaf-dims (ispace-list-leaf-dims isps) dim-var-map))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Logic-mode helpers for constant-folding dimension arithmetic.
+
+(define all-dim-const-p ((dims dim-listp))
```
acoglio, line 292: This kind of operation is very similar to others in `abstract-syntax-structurals.lisp`. I've just added one equivalent to this there (not pushed yet at the time of writing this review, but maybe soon).

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+    (b* ((d (car dims)))
+      (and (dim-case d :const t :otherwise nil)
+           (all-dim-const-p (cdr dims))))))
+
+(define dim-const-val* ((d dimp))
+  :returns (v natp)
+  :short "Extract the nat value from a @(':const') dim, returning 0 otherwise."
+  (dim-case d :const d.val :otherwise 0))
+
+(define dim-list-const-vals ((dims dim-listp))
```
acoglio, line 306: If this were guarded by `all-dim-const-p`, it could use `dim-const->val` instead of `dim-const-val*` above, which is only used in `dim-list-const-vals`. With that guard, it would be very similar to other operations in `abstract-syntax-structurals.lisp`. I've just added one equivalent to (a guarded version of) this one there (not pushed at the time of writing this review, but maybe soon).

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+     non-negative (natural).  Everything else is left structurally unchanged."))
+  :types (dims)
+  :extra-args ((dim-var-map string-nat-mapp))
+  :override
+  ((dim :var (b* ((dim-var-map (string-nat-map-fix dim-var-map))
+                  (pair (assoc-equal dim.name dim-var-map))
+                  ((unless pair) (dim-var dim.name)))
+               (dim-const (nfix (cdr pair)))))
+   (dim :add (b* ((new-dims (dim-list-partial-eval-dim-arith dim.dims dim-var-map))
+                  ((unless (all-dim-const-p new-dims)) (dim-add new-dims)))
```
acoglio, line 356: If `dim-list-const-vals` were guarded as suggested above, the guard here would easily prove, since we check `all-dim-const-p` just above.

acoglio: Unrelated to the comment above: if `all-dim-const-p` does not hold, is there any purpose to at least combine all the ones that are constant into one? Maybe not. (If yes, some code to do that is in `ispace-equivalence.lisp`, which normalizes dimension additions.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+               (dim-const (nfix result)))))
+  :name ast-partial-eval-dim-arith)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Fold 2: propagate partial dim evaluation through shapes, ispaces, types,
+; and the full expression/atom/bind hierarchy.  Every embedded dim is handled
+; by the already-defined partial-eval-dim-arith fold above.
+
+(fty::deffold-map partial-eval-ast-dims
```
acoglio: Is there a technical reason to separate this fold from the previous one? They both have the same extra-args.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+   (shape :append (shape-append (shape-list-partial-eval-ast-dims shape.shapes dim-var-map)))
+   (shape :splice (shape-splice (ispace-list-partial-eval-ast-dims shape.ispaces dim-var-map)))
+   (type :bracket (type-bracket (type-partial-eval-ast-dims type.elem dim-var-map)
+                                (ispace-list-partial-eval-ast-dims type.ispaces dim-var-map)))
+   (ispace :dim (ispace-dim (dim-partial-eval-dim-arith  ispace.dim dim-var-map))))
+  :name ast-partial-eval-ast-dims)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define partial-eval-dims ((x exprp) (dim-var-map string-nat-mapp))
```
acoglio: An alternative to introducing this synonym (presumably to have a shorter name) could be to tweak the name of the fold above, e.g. `(fty::deffold-map partial-eval-dims ...)` or `(fty::deffold-map peval-dims ...)`, and then the name from the fold would just have `expr-` in front.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+      (expr-fix body)
+    (expr-let (list (bind-fix (car binds)))
+              (nest-let-binds (cdr binds) body))))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Fold: rewrite every multi-bind :let into nested single-bind :lets, recursing
+; through the whole expression/atom/bind hierarchy.
+
+(fty::deffold-map singletonize-let
```
acoglio: This could be an independent transformation. I'm not suggesting to change this PR, but maybe later it could be put into a separate file/topic.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+    :let (expr-let (bind-list-fix (append (bind-list-fix binds) body.binds))
+                   body.body)
+    :otherwise (expr-let (bind-list-fix binds) (expr-fix body))))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Fold: collapse chains of :let expressions into single multi-bind :lets.
+; This is the left inverse of singletonize-let on the latter's image.
+
+(fty::deffold-map flatten-let
```
acoglio: Same observation as for the previous let transformation.

books/kestrel/remora/monomorphize.lisp
```diff
@@ -0,0 +1,987 @@
...
+                            (expr-flatten-let expr.body))))
+  :name ast-flatten-let)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Lemmas relating fn-info-mapp lookups to the bind+bind-map value type;
+; needed for guard verification of the assoc-equal-based lookups below.
+
+(local
+ (defthm consp-of-assoc-equal-when-fn-info-mapp
```
acoglio, line 391: This and similar theorems might be avoidable by (locally) including std/alists/top.lisp.

acoglio (approved):
I haven't followed all the details, but the transformation looks reasonable. I left some suggestions for your consideration.

books/kestrel/remora/monomorphize.lisp
acoglio: It's now in: `dim-list-case-const`.

books/kestrel/remora/monomorphize.lisp
acoglio: It's now in: `dim-const-list->val`.

books/kestrel/remora/monomorphize.lisp
westfold: It does. The documentation wasn't updated when I added it. I'll do so.

books/kestrel/remora/monomorphize.lisp
westfold: Yes (or :fun). I'll document that.

books/kestrel/remora/monomorphize.lisp
westfold: I'll replace.

books/kestrel/remora/monomorphize.lisp
westfold: ditto

books/kestrel/remora/monomorphize.lisp
westfold: I'll reuse.

books/kestrel/remora/monomorphize.lisp
westfold: No, historical. I'll combine.

books/kestrel/remora/monomorphize.lisp
westfold: Good idea. In fact that function was not even used.

books/kestrel/remora/monomorphize.lisp
westfold: Yes. I've put this into a file utility-transforms.lisp. Is that a good name?

books/kestrel/remora/monomorphize.lisp
westfold: Put in the same file.

books/kestrel/remora/monomorphize.lisp
westfold: No longer relevant  because of other changes.

westfold:
Thanks for the great review, Alessandro! It will help me review future AI generated code.

books/kestrel/remora/monomorphize.lisp
acoglio: Sounds good to me. Easy to change in the future if we want to.

books/kestrel/remora/monomorphize.lisp
acoglio: Yes, I saw that when I got to the actual code. I forgot to update this initial comment I made.

PR #1973, opened 2026-07-05
[Remora] Added `head` and `tail` to `values.lisp`

ltmquan:
@acoglio


books/kestrel/remora/values.lisp
```diff
@@ -580,6 +580,16 @@
   (:bool-neq ())
   (:bool-to-int ())
   (:bool-to-float ())
+  (:head ())
+  (:head-t ((tval type-value)))
+  (:head-t-d-s ((tval type-value)
+		(d nat)
```
acoglio: The indentation looks off. Is it using tabs?

books/kestrel/remora/values.lisp
```diff
@@ -844,6 +854,26 @@
      :bool-neq bool-binop-tv
      :bool-to-int bool-to-int-tv
      :bool-to-float bool-to-float-tv
+     :head (prog2$ (impossible) (type-value-base (base-type-bool)))
+     :head-t (prog2$ (impossible) (type-value-base (base-type-bool)))
+     :head-t-d-s (make-type-value-fun
```
acoglio: Does this file certify? It looks like it's applying `make-type-value-fun` to `:elem`.

acoglio (changes requested):
This is part of what's needed (modulo the comments I left), but we also need the actual evaluation of these operations. There are also other parts in `values.lisp` that need updating. And the documentation, which currently says that `length` is the only polymorphic primitive operation supported.

acoglio (approved):
Looks good. As I mentioned at today's meeting, you could consider generalizing `primitives-evaluation-length-tests.lisp` to `primitives-evaluation-list-tests.lisp`, and extend it with tests for `head` and `tail` (and later other list operations).

ltmquan:
Got it, I'll work on adding the tests, as well as other functions, next.

acoglio:
Thanks. Let's get this PR merged in after adding the tests and before doing other functions.

ltmquan:
I added the `:hyp` like we discussed in Zulip, but I ran into another issue when proving `expr-value-wfp-of-eval-primop-fun` for the case of `prim-tail`. In order to prove that the result of `prim-tail` is well-formed, I needed the rule `expr-value-wfp-of-expr-value-vector` to fire, but its hyps contain `(list-repeatp (dims-of-expr-value-list vals))`. So I needed to add two lemmas (lines 1368-1386 in `primitives-evaluation.lisp`) to resolve it. With that addition, everything certifies, but I'm not sure if I should have the two lemmas be in that file or if there's a better way to do this. What do you think I should do?

acoglio:
Can you try disabling `cdr-of-dims-of-expr-value-list` in the proof that was failing, and see if it goes through without the additional theorems?

ltmquan:
I think I needed that rule for my `list-repeatp-of-dims-of-cdr-of-expr-value-vector->elems` rule. I can disable it in the `defret` proof that was failing, but if I disable my rule above, then it fails. In general, I think we do want `cdr-of-dims-of-expr-value-list` based on the structure of the subgoal.

ltmquan:
I just merged the recent changes to `testing-kestrel`, should be good now.

acoglio (approved):

PR #1975, opened 2026-07-10
Start on proofs for unique-names and renaming

westfold:
Proof that uniquify-names makes names unique. Add call to it in monomorphize. Begin proof that it preserves dynamic semantics (not properly reviewed yet). Also same for renaming.

### DIFF
--- a/books/kestrel/remora/renaming-evaluation.lisp
+++ b/books/kestrel/remora/renaming-evaluation.lisp
@@ -1,0 +1,5198 @@
+; Remora Library
+;
+; Copyright (C) 2026 Kestrel Institute (http://www.kestrel.edu)
+;
+; License: A 3-clause BSD license. See the LICENSE file distributed with ACL2.
+;
+; Author: Stephen Westfold
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(in-package "REMORA")
+
+(include-book "evaluation")
+(include-book "unique-names")
+
+(include-book "kestrel/utilities/defopeners" :dir :system)
+
+(include-book "portcullis")
+
+(local (include-book "std/omaps/top" :dir :system))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defxdoc+ renaming-evaluation
+  :parents (unique-names-validation)
+  :short "Evaluation is invariant under variable renaming:
+          the structurally recursive levels."
+  :long
+  (xdoc::topstring
+   (xdoc::p
+    "This book begins the proof of
+     @('eval-top-expr-of-expr-uniquify-names')
+     (see @(see unique-names-validation)) with its structurally recursive
+     levels: the evaluation of dimensions, shapes, ispaces, and types,
+     under a renaming of the ispace variables.
+     The values produced by the first three levels (integers, lists of
+     naturals, ispace values) contain no abstract syntax, so evaluating a
+     renamed AST yields results literally equal to the original ones.
+     Type values may embed abstract syntax (the bodies of universal,
+     product, and sum types) along with dynamic environments captured
+     for its free variables, so evaluating a renamed type yields the
+     @('renaming') of the original type value, where renaming a type value
+     (see @(tsee type-value-rename-ispace-vars)) renames the abstract
+     syntax embedded in it and, recursively, the captured environments.")
+   (xdoc::p
+    "The correspondence between the two dynamic environments is expressed
+     by the relations @(tsee denv-ispace-vars-renamed-p) and
+     @(tsee denv-type-vars-ispace-renamed-p): looking up (the renaming of)
+     a variable in the new environment yields the (renaming of the)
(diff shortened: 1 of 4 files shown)


books/kestrel/remora/renaming-evaluation.lisp
```diff
@@ -0,0 +1,5198 @@
...
+  :rewrite :direct)
+
+(in-theory (disable denv-ispace-vars-renamed-p))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Evaluation of renamed dimensions.
+
+(defret-mutual eval-of-rename-dim-vars
+  (defret eval-dim-of-dim-rename-dim-vars
```
acoglio, line 159: The defret formulation of this theorem makes it hard (for me, at least) to easily see that `int` in fact refers to `(eval-dim dim denv)`, because here we are far from the definition of `eval-dim`. It would clearer to use the `defthm-flag` generated by that `defines` to state the theorem in more explicit terms here.

acoglio: The same applies to other theorems.

books/kestrel/remora/renaming-evaluation.lisp
```diff
@@ -0,0 +1,5198 @@
...
+                       rename-ispace-var
+                       ispace-denv-lookup-ispace))
+   (and acl2::stable-under-simplificationp
+        '(:use ((:instance denv-ispace-vars-renamed-p-necc
+                           (var (ispace-var-shape
+                                 (shape-var->name shape)))))))))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Renaming of ispace variables in type values.
```
acoglio, line 242: These renamings of variables in (ispace, type, and expression) values are more general operations that should be put into their own file(s) and topic(s). It should be also possible to use `deffold-map`, like done for ASTs. The "connection" with the already existing renaming ops on ASTs can be done by overriding certain cases of (type and expression) values to call those renaming ops on ASTs.

books/kestrel/remora/renaming-evaluation.lisp
```diff
@@ -0,0 +1,5198 @@
...
+     :types (type-var-type-value-map-rename-ispace-vars
+             (type-denv->types denv) dim-renam shape-renam)))
+
+  :verify-guards :after-returns
+
+  ///
+
+  (fty::deffixequiv-mutual type-value-rename-ispace-vars)
+
+  (defret not-reserrp-of-type-value-rename-ispace-vars
```
acoglio, line 416: This kind of theorems seem unnecessary: the function in question returns a `type-valuep`, and we already have a theorem `not-reserrp-when-type-valuep`, which, combined with the return type of the function, should play the role of the theorem stated here in other proofs. Although the theorems of the form `not-reserrp-when-T` are disabled by default to prevent always backchaining form reserrp to all the T's that happen to be around, they can be enabled locally at the beginning of a file that may need them, or even enabled in the needed proofs if there's only a few of them.

books/kestrel/remora/renaming-evaluation.lisp
```diff
@@ -0,0 +1,5198 @@
...
+                  (and (stringp name)
+                       (set::in (ispace-var-shape name) vars))))
+  :induct t
+  :enable (dim/shape-names-of-ispace-vars
+           equal-of-ispace-var-shape
+           set::in))
+
+; The free ispace variables of dimensions are all dimension variables.
+
+(local
```
acoglio, line 546: `(local (defrule ...))` can be abbreviated to `(defrulel ...)`.

books/kestrel/remora/renaming-evaluation.lisp
```diff
@@ -0,0 +1,5198 @@
...
+   (implies (set::emptyp (set::intersect (set::insert k s) b))
+            (not (set::in k b)))
+   :use ((:instance not-in-when-emptyp-of-intersect
+                    (a k)
+                    (x (set::insert k s))
+                    (y b)))
+   :disable not-in-when-emptyp-of-intersect))
+
+(local
+ (defruled emptyp-intersect-of-insert-left
```
acoglio, line 4458: This kind of theorem belongs to the oset library, but perhaps there are simpler versions, or better ways to use existing oset theorems. It would be helpful to factor library extensions into separate files and topics; there are already some under`library-extensions.lisp`.

books/kestrel/remora/renaming-evaluation.lisp
```diff
@@ -0,0 +1,5198 @@
...
+         (:instance not-in-when-emptyp-of-intersect
+                    (a (set::head (set::intersect s b)))
+                    (x (set::insert k s))
+                    (y b)))
+   :disable (set::in-head not-in-when-emptyp-of-intersect)))
+
+; Commutation and absorption of key deletions on the renaming maps.
+
+(local
+ (defruled delete*-of-delete*-commute
```
acoglio, line 4471: Same comment I made above about oset theorems, but for omaps.

acoglio (changes requested):
This is a lot of code, so I only did a light and partial review. But I've made suggestions that I think would help making things shorter and clearer. I understand it's work in progress, so if you prefer to merge it now and then continue working on it, that's fine with me.

PR #1977, opened 2026-07-14
Remora - Added `append`

ltmquan:
Similarly to `head` and `tail`, I needed some additional lemmas (placed before `eval-primop-fun` in `primitives-evaluation.lisp`. I couldn't find a better way.

@acoglio


books/kestrel/remora/primitive-operation-values.lisp
```diff
@@ -470,7 +482,20 @@
                                       :elem op.tval
                                       :dims (cons op.dval op.sval)))
                            :out int-tv)
-                    :dims nil)))
+                    :dims nil)
+     :append (prog2$ (impossible) (type-value-base (base-type-bool)))
+     :append-t (prog2$ (impossible) (type-value-base (base-type-bool)))
+     :append-t-m-n-s (make-type-value-array
```
acoglio, line 490: In this call of `make-type-value-array`, although the default for `:dims` happens to be `nil`, it's best to be explicit and add `:dims nil`, as done for `:length-t-d-s`. I'm just seeing now that the same applies to head and tail here.

books/kestrel/remora/primitives-evaluation.lisp
```diff
@@ -1363,8 +1363,62 @@
...
+
+(define prim-append ((tval type-valuep) (m natp) (n natp) (s nat-listp)
+                     (val1 expr-valuep) (val2 expr-valuep))
+  :guard (and (expr-value-wfp val1) (expr-value-wfp val2))
+  :returns (val expr-value-resultp)
+  (b* ((m (lnfix m)) (n (lnfix n)) (s (nat-list-fix s))
+       ((unless (equal (dims-of-expr-value val1) (cons m s))) (reserr nil))
+       ((unless (equal (dims-of-expr-value val2) (cons n s))) (reserr nil))
+       (elems1 (if (expr-value-case val1 :vector)
+                   (expr-value-vector->elems val1) nil))
```
acoglio, line 1376: The normal style is to put the 'else' subterm in a new line. Same for the `if` below.

But I think we can do better than that: since we check just above that the dimensions of val1 and val2 are cons, it should follow that val1 and val2 are vectors, so it should suffice to do
```
(elems1 (expr-value-vector->elems val1))
```
and the same for elems2. This will engender a guard obligation. Replacing certain run-time checks never expected to fails with proofs helps validate the code.

books/kestrel/remora/primitives-evaluation.lisp
```diff
@@ -1363,8 +1363,62 @@
...
+       (elems2 (if (expr-value-case val2 :vector)
+                   (expr-value-vector->elems val2) nil))
+       (elems (append elems1 elems2)))
+    (if (consp elems)
+        (expr-value-vector elems)
+      (expr-value-vector-empty s tval))))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(defrule list-repeatp-of-append
```
acoglio: This is a general theorem about list-repeatp and append, so it should be put in lists.lisp, under list-repeatp.

books/kestrel/remora/primitives-evaluation.lisp
```diff
@@ -1363,8 +1363,62 @@
...
+           check-dims-of-expr-value
+           check-dims-of-expr-value-list-when-expr-value-list-wfp
+           acl2::not-reserrp-when-nat-list-listp))
+
+(defrule expr-value-wfp-of-expr-value-vector-of-append-of-vector->elems
+  (implies (and (expr-value-wfp val1) (expr-value-case val1 :vector)
+                (expr-value-wfp val2) (expr-value-case val2 :vector)
+                (equal (dims-of-expr-value val1) (cons m s))
+                (equal (dims-of-expr-value val2) (cons n s))
+;                (equal (cdr (dims-of-expr-value val1))
```
acoglio: Commented-out code should be removed.

books/kestrel/remora/primitives-evaluation.lisp
```diff
@@ -1363,8 +1363,62 @@
...
+;                       (cdr (dims-of-expr-value val2)))
+                (consp (append (expr-value-vector->elems val1)
+                               (expr-value-vector->elems val2))))
+           (expr-value-wfp
+            (expr-value-vector (append (expr-value-vector->elems val1)
+                                       (expr-value-vector->elems val2)))))
+  :use ((:instance list-repeatp-of-append
+                   (x (dims-of-expr-value-list (expr-value-vector->elems val1)))
+                   (y (dims-of-expr-value-list (expr-value-vector->elems val2)))))
+  )
```
acoglio: Join with previous line.

books/kestrel/remora/primitives-evaluation.lisp
```diff
@@ -1363,8 +1363,62 @@
...
 
+(defrule list-repeatp-of-append
+  (implies (and (list-repeatp x)
+                (list-repeatp y)
+                (equal (car x) (car y)))
+           (list-repeatp (append x y)))
+  :induct t
+  :enable list-repeatp)
+
+(defrule dims-of-expr-value-of-car-of-expr-value-vector->elems
```
acoglio, line 1386: This and the next theorem look very specific, so I think there should be ways to avoid and/or generalizing them — I was able to do that for one of the theorems for head and tail. But I'd have to look into that myself, so fine to leave them for now.

books/kestrel/remora/primitives-evaluation.lisp
```diff
@@ -1560,6 +1619,10 @@
                           (list (type-var-atom "t"))))
                  (reserr nil)))
              (expr-value-primop (primop-value-length-t (first tvals))))
+   :append (b* (((unless (type-values-match-type-vars-p
+                          tvals (list (type-var-atom "t"))))
```
acoglio: Line break between arguments for readability and consistency with head, tail, length.

acoglio (changes requested):

books/kestrel/remora/primitives-evaluation.lisp
ltmquan: I think there's still the case of `elems1` being a `:vector-empty` when the first element of its dimension is a `0` and having the `if` check accounts for that.

acoglio: Yes, you are right.

acoglio (approved):

PR #1978, opened 2026-07-14
[Remora] Make unary iapp "primary".

acoglio:
*UPDATE*: I pushed more changes here, which also unarize :ilambda and :pi (which go well with :iapp, and eliminate the two case splits mentioned earlier, as expected). The changes also resolve the conflicts that were previously present due to drift.

- Expr :iapp renamed to :iappn.
- Expr :iapp re-added as unary.
- CST-to-AST maps iapp of one arg to :iapp, iapp of 2+ args to :iappn.
- Type checking handles :iapp directly on one arg (hence simpler), and handles :iappn as sequence of :iapp.
- Evaluation handles :iapp directly on one arg (hence simpler), and handles :iappn as sequence of :iapp.
- Primop values gain additional stages, because they are now curried on the ispaces.

(A couple of case splits in some code will go away once the same treatment is done to atom :ilambda and to type-value :pi. See added documentation about that.)

The general idea (for this and upcoming constructs) is that all the code operating on n-ary things will operate on unary things, making things strictly simpler. n-ary things are sugar for nested unary things; since our ASTs include sugar, we handle them by repeatedly calling the code for unary things. Mirrors the latest Haskell implementation of Remora.

Definitions of core and desugaring not updated yet — no need yet.


bendyarm (commented):
* Should `:appn` be excluded as core in `abstract-syntax-core.lisp:130`?
* Should expr-desugar convert `:iappn` into nested unary applications?
* It looks like the monomorphizer handles `:iappn` but not unary `:iapp`.
* Desugaring a combined application with one ispace argument constructs a one-argument :iappn.
* `(i-app (i-fn (@s) 7) 3)` is rejected but Haskell accepts it by treating `3` as the shape `[3]`

I would be glad to develop more tests for this.  It might be fine to merge it and work on the issues later, since we don't want it to diverge too much.

acoglio:
I was planning to update core and desugar after everything has been migrated to unary.

I've updated the monomorphizer as best as I could, but there might be something additional to do. I also have more local commits that take the migration to unary further, and might solve some issues.

Is the auto-lifting of 3 to a shape in the example orthogonal to the unary changes?

bendyarm:
> Is the auto-lifting of 3 to a shape in the example orthogonal to the unary changes?
Yes, it just popped up as a counterexample to "matching Haskell?"

acoglio:
> > Is the auto-lifting of 3 to a shape in the example orthogonal to the unary changes?
> > Yes, it just popped up as a counterexample to "matching Haskell?"

Thanks, I've made a note of the issue. We should look into that independently from this PR.

acoglio:
Merging this as discussed at the meeting.

PR #1979, opened 2026-07-15
[remora] more type checking tests

bendyarm:


acoglio (approved):
Let's merge this as is, so it's easy to access, and then we can work on the type checker and the tests.

PR #1980, opened 2026-07-15
REMORA - Added `reverse` function

ltmquan:
Added `prim-reverse` and also updated docs and added tests for `append` and `reverse` in `primitives-evaluation-polymorphic-test.lisp`.

@acoglio


ericwhitmansmith:
It would be good to link to https://github.com/remora-lang/remora/issues/31 (in the comment that mentions the bug has been reported).

books/kestrel/remora/lists.lisp
```diff
@@ -125,6 +125,18 @@
                     (equal (car x) (car y)))))
     :induct t)
 
+  (defrule car-of-rev-when-list-repeatp
```
acoglio: This should be disabled because it's triggered by (car (rev x)), which may have nothing to do with list-repeatp. In general, we try to avoid backchaining from more general concepts to more specific concepts, for enabled-by-default rules; they can be enabled where needed. Unfortunately not all rules in all libraries follow this principle.

books/kestrel/remora/lists.lisp
```diff
@@ -125,6 +125,18 @@
...
+    (implies (list-repeatp x)
+             (equal (car (rev x)) (car x)))
+    :induct t
+    :enable (list-repeatp rev))
+
+  (defrule list-repeatp-of-rev
```
acoglio, line 134: Would this work in the form (list-repeatp (rev x)) = (list-repeatp x)?

books/kestrel/remora/primitives-evaluation.lisp
```diff
@@ -1709,6 +1740,72 @@
...
+     The type value @('tval') is currently only used
+     for when the output is an empty array
+     and not for checking well-formedness,
+     because our well-formedness checks on expression values
+     currently concern dimensions but not types;
+     it will be used to further check the argument cell
+     when those checks are extended to types."))
+  (b* ((d (lnfix d)) (s (nat-list-fix s))
+       ((unless (equal (dims-of-expr-value val1) (cons d s))) (reserr nil))
+       (elems (if (expr-value-case val1 :vector)
```
acoglio: This should use the new function `expr-value-vector-elements`, like `append` does. It has a stronger guard and is clearer.

acoglio (changes requested):

books/kestrel/remora/lists.lisp
```diff
@@ -125,6 +125,20 @@
...
     :induct t)
 
+  (defruled car-of-rev-when-list-repeatp
+    (implies (list-repeatp x)
+             (equal (car (rev x)) (car x)))
+    :induct t
+    :enable (list-repeatp rev))
+
+  (defrule list-repeatp-of-rev
+    (implies (list-repeatp (rev x))
```
ericwhitmansmith: Should be equal instead of implies.

acoglio: Yes, I mentioned that [here](https://github.com/acl2/acl2/pull/1980/changes#r3598583475), and I just tried, it works (as expected).

@ltmquan It makes for a better/stronger rule.

ericwhitmansmith: I think what happened is that your suggestion was only partially applied (the calls of list-repeatp were swapped but the connective was left as implies, not turned into equal).  That actually makes it a very risky rule because of the risk of backchaining to "rev of rev of rev ...".

ltmquan: Yes, sorry about that, don't know why I missed that. Should be in place now.

ltmquan:
I updated the doc for `reverse` since the bug is fixed on the Remora side.

books/kestrel/remora/lists.lisp
acoglio: Ah! I didn't notice the swap.

acoglio (approved):

PR #1981, opened 2026-07-17
REMORA - Added `index` and `index2d`

ltmquan:
Added two intrinsics `index` and `index2d` using `nth`.

@acoglio


ltmquan:
Updated `index` and `index2d` to fit the new changes related to currying.

@acoglio

acoglio (approved):

PR #1987, opened 2026-07-24
[definductive] Allow a non-recursive single predicate.

gjurgensen:
When a single predicate is defined, drop the requirement that it be recursive: it then yields a non-recursive proof validity function, generated without a measure or induction hints, and its minimality theorem is proved by a direct case analysis on the proof kind rather than by induction. Multiple predicates must still be mutually recursive (form a single clique).

### DIFF
--- a/books/std/util/definductive.lisp
+++ b/books/std/util/definductive.lisp
@@ -1366,24 +1366,27 @@
    (xdoc::p
     "Besides processing the individual rules,
      we check that the rule names are all distinct,
-     that the predicates being defined form
-     a single mutually recursive clique,
      and that at least one rule is non-recursive
-     (which provides a base case for the inductive definition).")
+     (which provides a base case for the inductive definition:
+     without it, the smallest predicate satisfying the rules is empty).")
    (xdoc::p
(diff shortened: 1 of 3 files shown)


books/std/util/definductive.lisp
```diff
@@ -3523,9 +3573,18 @@
...
       (first-kind symbolp))
      :guard (no-duplicatesp-equal (defind-irule-info-list->name infos))
-     :returns (hint-branches true-listp)
+     :returns (mv (hint-branches true-listp)
+                  (lemma-instances true-listp)
+                  (irule-valid-fns symbol-listp)
+                  (irule-acc-thms symbol-listp))
      :parents nil
-     (b* (((when (endp infos)) nil)
+     :short "Return, for each inference rule,
```
acoglio: Because of the `:parents nil`, I don't think that this doc will appear in the manual. I usually document the auxiliary (i.e. in `:prepwork`) functions in the doc for the main function, e.g. "The loop function returns, for each inference rule, ...".

acoglio (approved):
Just one comment.

And you should add your name to the headers.

PR #1991, opened 2026-07-29
[REMORA] added `prim-transpose`

ltmquan:
Added `prim-transpose` along with `transpose-list-list` that transposes a 2d list, and some theorems about it.


ltmquan:
I left most of the new rules I added in `lists.lisp` disabled for safety and I left 2 very specific lemmas needed for `prim-transpose2d` in the `:prepwork` section of it. Would love to hear your suggestions about any changes.

@acoglio

books/kestrel/remora/expression-values-and-environments.lisp
```diff
@@ -467,6 +467,13 @@
     (:reshape-t-s1-s2 ((tval type-value)
                        (s1val nat-list)
                        (s2val nat-list)))
+    (:transpose2d ())
+    (:transpose2d-t ((tval type-value)))
+    (:transpose2d-t-m ((tval type-value)
+                     (mval nat)))
```
acoglio: This, and the two lines below, do not look properly indented.

books/kestrel/remora/lists.lisp
```diff
@@ -213,7 +226,30 @@
...
+  ///
+
+  (defruled cons-listp-when-all-of-len-p
+    (implies (and (all-of-len-p vals n)
+                  (posp n))
+             (cons-listp vals))
+    :induct t
+    :enable cons-listp)
+
+  (defruledl len-of-append-all-when-all-of-len-p-aux
```
acoglio: This should be just called len-of-append-all-when-all-of-len-p, without -aux, and not be local, because it looks like a good, general theorem.

books/kestrel/remora/lists.lisp
```diff
@@ -213,7 +226,30 @@
...
+    (implies (all-of-len-p lists n)
+             (equal (len (append-all lists))
+                    (* (len lists) (nfix n))))
+    :induct t
+    :enable (append-all nfix))
+
+  (defruled len-of-append-all-when-all-of-len-p
```
acoglio: While this looks quite ad hoc, and essentially an instance of the previous one.

Is there a better way to do the proof where this is used? Sometimes a :use hint is preferable to an ad hoc rewrite rule, particularly if added as a general library theorem.

At a minimum, I would rename this to reflect its specificity, e.g. len-of-append-all-when-all-of-len-p-of-car.

books/kestrel/remora/lists.lisp
```diff
@@ -373,7 +437,93 @@
...
+  (defruled cdr-list-of-repeat
+    (equal (cdr-list (repeat n list))
+           (repeat n (cdr (true-list-fix list))))
+    :induct (repeat n list)
+    :enable repeat))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define transpose-list-list ((lists true-list-listp))
+  :returns (lists1 true-list-listp)
```
acoglio, line 470: I would add a guard saying that all the lists have the same length.

books/kestrel/remora/primitives-evaluation.lisp
```diff
@@ -2030,6 +2031,101 @@
...
+     via @(tsee append-all),
+     and arrange the result according to the transposed dimensions
+     via @(tsee expr-value-with-nonempty-dims).
+     This mirrors the interpreter in [impl],
+     which computes @('concat (transpose (split n elts))')."))
+  (b* ((m (lnfix m))
+       (n (lnfix n))
+       ((unless (equal (dims-of-expr-value val1) (list m n)))
+        (reserr nil))
+       (dims (list n m))
```
acoglio, line 2075: Move two lines above and use it in the unless above.

books/kestrel/remora/primitives-evaluation.lisp
```diff
@@ -2030,6 +2031,101 @@
...
+                              expr-value-atoms
+                              list-split-of-repeat
+                              transpose-list-list-of-repeat-of-repeat
+                              append-all-of-repeat-of-repeat
+                              len-of-append-all-when-all-of-len-p
+                              all-of-len-p-of-transpose-list-list
+                              len-of-car-of-transpose-list-list
+                              consp-of-car-list-split)))
+  :prepwork
+  ((defrule nat-list-product-of-pair
```
acoglio: These two rules are ad hoc, global (:prepwork does not make things local), and enabled.

books/kestrel/remora/primitives-evaluation-polymorphic-tests.lisp
```diff
@@ -759,9 +759,121 @@
...
 ; not an operation applicable to another cell.
 (acl2::assert-event
- (not (expr-value-case
-       (eval-primop-fun (make-primop-value-reshape-t-s1-s2 :tval *tv-int*
-                                                           :s1val (list 2 3)
-                                                           :s2val (list 6))
-                        *mat23*)
-       :primop)))
+ (reserrp
+  (eval-primop-fun-chain (make-primop-value-reshape-t-s1-s2 :tval *tv-int*
```
acoglio: eval-primop-fun-chain no longer exists. This PR needs to be merged with the latest from master.

acoglio (changes requested):

books/kestrel/remora/lists.lisp
ltmquan: My initial reasoning for this was that `all-of-len-p` only made sense if everything in `lists` is of length `(len (car lists))`, and not any arbitrary length. Should this theorem stay then, or should I only keep the previous one and use `:use` hints in proofs?

books/kestrel/remora/primitives-evaluation.lisp
ltmquan: The `unless` uses the original dims of `m x n` while the `dims` holds `n x m`.

acoglio: My bad, I misread.

books/kestrel/remora/expression-values-and-environments.lisp
```diff
@@ -1,10 +1,10 @@
-; Remora Library
-;
-; Copyright (C) 2026 Kestrel Institute (http://www.kestrel.edu)
-;
-; License: A 3-clause BSD license. See the LICENSE file distributed with ACL2.
-;
-; Author: Alessandro Coglio (www.alessandrocoglio.info)
+                                        ; Remora Library
```
acoglio: Something strange happened here.

acoglio (changes requested):
It looks like the whole expression-values-and-environments.lisp file got re-indented. We should restore the original indentation.

books/kestrel/remora/expression-values-and-environments.lisp
ltmquan: That's my bad, my Emacs auto-indent is not working properly

books/kestrel/remora/lists.lisp
acoglio: It's up to you: you could try the :use hint if you want to see if/how that works out, or you can just rename these two theorems as I've suggested and otherwise leave things as they are.

acoglio (changes requested):
Just one renaming left, after that I think it's good to merge.

books/kestrel/remora/expression-values-and-environments.lisp
acoglio: I know, it happens! 😄

ltmquan:
I renamed the functions as you suggested

acoglio (approved):

PR #1993, opened 2026-07-29
[Remora] add json deserializer for asts

Sarah-Scott:
Added fromJSON functions for translating Haskell AST JSON into ACL2 AST. Also added file entry point for deserialization.


books/kestrel/remora/deserializer.lisp
```diff
@@ -0,0 +1,1291 @@
...
+     fixtype;
+     otherwise @('erp') is @('nil') and @('x') is the resulting AST.
+     JSON objects are dispatched on their @('\"tag\"') member,
+     whose string value names the [impl] AST node being decoded."))
+  :order-subtopics t
+  :default-parent t)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;; These lemmas establish that the JSON value returned by
```
acoglio, line 54: Not for this PR, but we should probably move these to the ACL2 library for JSON.

books/kestrel/remora/deserializer.lisp
```diff
@@ -0,0 +1,1291 @@
...
+  :rule-classes :linear)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;; Dim
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defines dim-fromJSON
+  :verify-guards nil
```
acoglio: I would move this and the next line after the 2nd define forms, because these two lines are merely details to get the proofs done, so there's no need for them to prominently appear at the start.

books/kestrel/remora/deserializer.lisp
```diff
@@ -0,0 +1,1291 @@
...
+    (b* (((acl2::reterr) nil))
+      (if (consp js)
+          (b* (((acl2::erp hd)
+                (dim-fromJSON (car js)))
+               ((acl2::erp tl)
+                (dim-list-fromJSON (cdr js))))
+            (acl2::retok (cons hd tl)))
+        (acl2::retok nil)))))
+
+(verify-guards dim-fromJSON)
```
acoglio: This could go under the defines form, preceded by a `///`, which is used to separate the main components of the defines form from additional related events. (You can see several examples in other files, searching for `///`.)

books/kestrel/remora/deserializer.lisp
```diff
@@ -0,0 +1,1291 @@
...
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;; Shape
+;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defines shape-fromJSON
+  :verify-guards nil
```
acoglio: Similar comments as above apply here, and likely to other defines in this file.

books/kestrel/remora/deserializer.lisp
```diff
@@ -0,0 +1,1291 @@
...
+    (b* (((acl2::reterr) nil))
+      (if (consp js)
+          (b* (((acl2::erp hd)
+                (shape-fromJSON (car js)))
+               ((acl2::erp tl)
+                (shape-list-fromJSON (cdr js))))
+            (acl2::retok (cons hd tl)))
+        (acl2::retok nil)))))
+
+(verify-guards shape-fromJSON)
```
acoglio: And this as well, `///` as above.

acoglio (approved):
I left some suggestions, but it looks good. I didn't examine all the details of the JSON processing though. I'm sure we'll test all of this at some point.

acoglio:
I'm also adding @bendyarm as reviewer.

bendyarm:
This is what codex GPT 5.6 Sol reported, after thorough static checking and then running the code on both the Haskell and ACL2 side:

 1. [P1] Singleton quantified types become ill-formed ACL2 ASTs.
     TEForall, TEPi, and TESigma always construct the corresponding *n variant, regardless of parameter count (deserializer.lisp
     (https://github.com/acl2/acl2/blob/482e746ed62807cc81d4afb91859626ae70405d1/books/kestrel/remora/deserializer.lisp#L506-L553)). Normal Remora output contains singleton
     parameter arrays, but ACL2 requires at least two parameters for these variants. An actual serialized singleton TESigma decoded to:

     (:SIGMAN ((:DIM "d")) (:BASE (:INT)))

     with (type-wfp x) = nil. Construct :forall, :pi, or :sigma for singleton lists, the same way syntax abstraction already does; reject empty lists.

  2. [P1] Valid current Remora record programs cannot be deserialized.
     Remora’s serializer emits TERecord, Struct, and FieldProj (Serializer.hs (https://github.com/remora-lang/remora/blob/2bc2f23/src/Serializer.hs#L77-L88), expression instances
     (https://github.com/remora-lang/remora/blob/2bc2f23/src/Serializer.hs#L199-L224)), while the new decoder has none of these tags. Serializing the existing tests/rec0.remora
     produces a Struct, which expr-fromJSON rejects as unrecognized. Since ACL2’s current AST lacks record constructors, either record support is a prerequisite or the decoder’s
     supported subset must be explicitly documented instead of being described as an inverse mapping.

  3. [P2] Invalid numeric values are silently changed instead of rejected.
     DimN and shape dimensions use nfix, while IntVal uses ifix (dimension conversion
     (https://github.com/acl2/acl2/blob/482e746ed62807cc81d4afb91859626ae70405d1/books/kestrel/remora/deserializer.lisp#L153-L160), integer conversion
     (https://github.com/acl2/acl2/blob/482e746ed62807cc81d4afb91859626ae70405d1/books/kestrel/remora/deserializer.lisp#L602-L611), dimension lists
     (https://github.com/acl2/acl2/blob/482e746ed62807cc81d4afb91859626ae70405d1/books/kestrel/remora/deserializer.lisp#L1160-L1173)). ACL2 testing showed:
      - DimN with -3 returns success and dimension 0.
      - IntVal with 3/2 returns success and integer literal 0.

     Check natp or integerp and return an error before constructing the AST.

  4. [P2] The new functionality is absent from the Remora top-level book.
     Neither new book is included by books/kestrel/remora/top.lisp
     (https://github.com/acl2/acl2/blob/482e746ed62807cc81d4afb91859626ae70405d1/books/kestrel/remora/top.lisp#L11-L24). Consequently, users of remora/top do not receive the
     deserializer or its documentation. The core and file-entry books should be added, unless they are intentionally standalone and that policy is documented.

On validation, here's what Sol reported:

Both new books certify successfully. I also compared every handled tag against the current Haskell serializer and exercised the cases above in a persistent ACL2 session. The PR
  currently reports no automated status checks and adds no regression tests; serializer-generated fixtures for singleton quantifiers, records, and invalid numeric input would be
  valuable.

Sarah-Scott:
I've completed each of the recommended changes. Each change should be in its own commit.

Sarah-Scott:
The diff is a little strange due to having to merge twice. I did `git merge -X theirs` so I believe it shouldn't have messed up anyone's code. But please let me know if you see a problem. I am not particularly savvy with git yet.

acoglio:
There is an accidental change to [books]/kestrel/c/syntax/disambiguator.lisp, which takes it back to some previous version. If you get the latest version of that file from master or testing-kestrel (e.g. from a fresh clone), and copy it onto the version in your current branch, it should no longer show any diff on that file.

Sarah-Scott:
I rebased to fix those accidental file changes. The diff is back to what it should be now. I also added a couple small edits to fix some issues that Eric McCarthy caught in the deserializer.

acoglio:
The diff looks fine to me.

bendyarm:
LGTM, rebased and merged.

PR #1994, opened 2026-07-30
[FTY] Add `deftreeset` to FTY `deftypes` generators

gjurgensen:
This type generator largely mirrors other such generators. The one major change is that it generates *two* functions which go in the predicate clique. To accommodate this, all `flex*-predicate-def` functions were generalized to return lists instead of single forms.

Functions were added to the `visitor` machinery for `deftreeset`-generated types, but the functions simply throw a hard error.

The final item of note is that `centaur/fty/deftypes` depends on the treeset portcullis, but it does not actually include any of the treeset books. You have to include `kestrel/fty/deftreeset` to use the `deftreeset` generator; otherwise, it will gracefully detect the book isn't included and give a pretty error message telling you which book to include.


acoglio (approved):
I'm not familiar with the FTY internals, but it looks good as far as I can tell.

solswords (approved):
Nice -- looks good to me.
I had a thought about how one might avoid inlining a fold into the top-level recognizer -- include the user measure in an ordinal that accounts for the two-level predicate automatically -- but since the inlining is wrapped up in the type abstraction once the deftypes is done, it seems ok to leave it.

gjurgensen:
> I had a thought about how one might avoid inlining a fold into the top-level recognizer -- include the user measure in an ordinal that accounts for the two-level predicate automatically

Yes, the inlining of one level of the fold into the top-level treeset recognizer is a bit unfortunate; as you say it is just to get the recursion to decrease structurally. I thought about changing the default measure to be an ordinal, perhaps using `two-nat-measure` or `nat-list-measure`. Two issues swayed me against that direction for now.

1. I didn't want to change the FTY internals more than I needed to in the short-term.
2. More significantly, the user-supplied measure itself may be an ordinal, and I'm not sure there is a general way to "nest" that ordinal in a larger ordinal to allow a side-step in the clique like my top-level recognizer would like to do to call the sibling fold.

Perhaps this is something to consider for the future. I think this version is fine for now.

gjurgensen:
I'll merge once I do a local `regression-everything` build.

PR #1995, opened 2026-07-31
[Remora] add prim-flatten

Sarah-Scott:
Added the primitive `prim-flatten`, which merges the leading two axes into a single axis.

Implementation closely follows that of `prim-reshape`.

### DIFF
--- a/books/kestrel/remora/primitives-evaluation.lisp
+++ b/books/kestrel/remora/primitives-evaluation.lisp
@@ -2022,7 +2022,7 @@
           (expr-value-with-empty-dim s2 tval)))
        (atoms (expr-value-atoms val1)))
     (expr-value-with-nonempty-dims s2 atoms))
-  
+
   ///
 
   (defret expr-value-wfp-of-prim-reshape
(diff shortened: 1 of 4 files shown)


books/kestrel/remora/primitives-evaluation.lisp
```diff
@@ -2033,6 +2033,68 @@
...
+     dimension @('n') followed by the shape @('s'),
+     and the result is the array with the leading two axes
+     merged into a single axis,
+     whose dimensions are the dimension @('m*n')
+     followed by the shape @('s').
+     The guard requires the argument cell to be well-formed;
+     we defensively check that it has the expected dimensions.")
+   (xdoc::p
+    "If the two leading dimensions or the remaining shape
+     has a zero dimension, the result is empty:
```
acoglio:
```suggestion
     include a zero dimension, the result is empty:
```

acoglio (approved):
Looks good, but in the meanwhile there were some new changes in testing-kestrel, which now need to be merged in. I think there's no substantial conflict, maybe just the initial static environment. Some type constructors like `t[]` have changed slightly, but it should be easy to see what to do (please reach out if not).

Sarah-Scott:
I merged in the latest changes to testing-kestrel.

PR #1996, opened 2026-07-31
[C$] Three-valued integer constant expression recognizer

gjurgensen:
This creates a new book, `constant-expressions.lisp`. Currently, the book only provided support for checking integer constant expression status. It uses a three-valued logic to ease the burden of propagating unknown cases.

Currently, the recognizer is not used outside of new tests. Eventually, it may be used to distinguish fixed-length array types from variable-length arrays.

### DIFF
--- a/books/kestrel/c/syntax/constant-expressions.lisp
+++ b/books/kestrel/c/syntax/constant-expressions.lisp
@@ -1,0 +1,924 @@
+; C Library
+;
+; Copyright (C) 2026 Kestrel Institute (http://www.kestrel.edu)
+;
+; License: A 3-clause BSD license. See the LICENSE file distributed with ACL2.
+;
+; Author: Grant Jurgensen (grant@kestrel.edu)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(in-package "C$")
+
+(include-book "validation-annotations")
+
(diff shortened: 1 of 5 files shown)


books/kestrel/c/syntax/constant-expressions.lisp
```diff
@@ -0,0 +1,889 @@
...
+  :short "Three-valued check of whether @('sizeof') of a type name
+          produces an integer constant."
+  (if (tyname-annop tyname)
+      (type-sizeof-result-const-3p
+       (type-vinfo->type (tyname->info tyname)))
+    :unknown))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define typequal/attribspec-list-expr-free-p
```
acoglio, line 143: This and the next could go into a more central place, since they are just ops on ASTs.

I'm not necessarily proposing this for this PR.

books/kestrel/c/syntax/constant-expressions.lisp
```diff
@@ -0,0 +1,889 @@
...
+    ((expr exprp)
+     (evaluatedp 3p)
+     (cast-restrictionsp booleanp)
+     (dialect c::dialectp))
+    :guard (expr-unambp expr)
+    :returns (3vl 3p)
+    :short "Recursive core of the integer constant expression (ICE) check."
+    :long
+    (xdoc::topstring
+   (xdoc::p
```
acoglio: The indentation looks wrong here.

acoglio (approved):

PR #1998, opened 2026-08-04
Reduce `std/util` includes in centaur books

gjurgensen:
There are a couple of places in the `centaur` books where `std/util/top` is being included. With the recent addition of `std/util/definductive`, this caused a lot of unnecessary dependencies on `definductive`. This PR replaces `std/util/top` inclusions with just the necessary books, thereby reducing the number of unneeded dependencies.


solswords (approved):
Great, thanks!

PR #1999, opened 2026-08-04
[STS] Obtain parameter declaration type differently.

acoglio:
There is now a function to obtain the type of a parameter declaration from validation annotations, specifically the ones at the underlying parameter declarators.

### DIFF
--- a/books/kestrel/c/transformation/struct-type-split.lisp
+++ b/books/kestrel/c/transformation/struct-type-split.lisp
@@ -597,6 +597,20 @@
       (c$::expr-type expr)
     (c$::type-unknown)))
 
+(define sts-param-declon-type ((pdeclon param-declonp))
+  :guard (param-declon-annop pdeclon)
+  :returns (type typep)
+  :short "Best-effort type of a parameter declaration,
+          read from its annotation."
+  :long
+  (xdoc::topstring
+   (xdoc::p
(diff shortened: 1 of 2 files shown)


books/kestrel/c/transformation/struct-type-split.lisp
```diff
@@ -597,6 +597,20 @@
...
 
+(define sts-param-declon-type ((pdeclon param-declonp))
+  :guard (param-declon-annop pdeclon)
+  :returns (type typep)
+  :short "Best-effort type of a parameter declaration,
+          read from its annotation."
+  :long
+  (xdoc::topstring
+   (xdoc::p
+    "This is similar in spirit to @(tsee sts-param-declon-type);
```
gjurgensen: I assume this `tsee` is meant to refer to some other function, not itself. Perhaps it was meant to reference `sts-expr-type`?

gjurgensen (approved):

books/kestrel/c/transformation/struct-type-split.lisp
acoglio:
```suggestion
    "This is similar in spirit to @(tsee sts-expr-type);
```

acoglio: Indeed!

PR #2001, opened 2026-08-05
Remora - added `prim-iota/static`

ltmquan:


books/kestrel/remora/base-values.lisp
```diff
@@ -66,6 +66,13 @@
   :ok int-value
   :pred int-value-resultp)
 
+;;;;;;;;;;;;;;;;;;;;
+
+(std::defprojection int-value-list ((x integer-listp))
```
acoglio: This would be a good name, but unfortunately it creates an XDOC topic with the same name as the XDOC topic for the fixtype int-value-list. So I usually tweak the name of this kind of function, like int-value-list-of.

books/kestrel/remora/expression-values-and-environments.lisp
```diff
@@ -1171,6 +1172,17 @@
     :induct t
     :enable repeat)
 
+  (defruled dims-of-expr-value-list-of-expr-value-base-list
+    (equal (dims-of-expr-value-list (expr-value-base-list bvals))
+           (repeat (len bvals) nil))
+    :induct (expr-value-base-list bvals)
+    :enable (
```
acoglio: Can join with next line.

books/kestrel/remora/nat-lists.lisp
```diff
@@ -104,6 +114,32 @@
         ((endp (cdr nats)) (- (lnfix (car nats))))
         (t (- (lnfix (car nats)) (nat-list-sum (cdr nats))))))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define nat-list-from-to ((i natp) (n natp))
+  :returns (nats nat-listp)
+  :short "List of the naturals from @('i') (inclusive) to @('n') (exclusive)."
+  :measure (nfix (- (nfix n) (nfix i)))
```
acoglio: I would move :measure and :hints just after the body, because they are just proof details, which don't need to be some prominent to appear just after the signature and the documentation (:short).

books/kestrel/remora/primitives-evaluation-on-ispaces.lisp
```diff
@@ -38,6 +38,50 @@
...
+     the single ispace application supplies the shape @('s'),
+     and the result is the array of that shape
+     whose atoms are the naturals below the number of elements,
+     in row-major order.
+     Unlike all other operations, no argument cell is involved:
+     the ispace application directly yields the final array.")
+   (xdoc::p
+    "If the shape has a zero dimension, the result is empty;
+     the element type is always the integer atom type,
+     so, unlike @(tsee prim-transpose2d),
```
acoglio: There is no type value here. Maybe this and the next line should be deleted, leaving just the first two lines of this paragraph.

books/kestrel/remora/primitives-evaluation-on-ispaces.lisp
```diff
@@ -222,6 +266,10 @@
                                                                 :mval op.mval
                                                                 :nval ival.val))
                      :shape (reserr nil))
+   :iota/static (ispace-value-case
+              ival
```
acoglio, line 272: The indentation of this and the next two lines looks wrong.

acoglio (approved):
Looks fine. I left some suggestions. Also needs merging with the latest testing-kestrel.

acoglio (approved):

PR #2002, opened 2026-08-06
[smtlink] keep proofs independent of the size of the operator/type table

bendyarm:
This is a small change with no semantic effect.

Smtlink's tables of supported operators and types (`*SMT-basics*`,
`*SMT-types*`, ...) are consulted in two ways that currently leak into
proofs:

1. Three `:returns` theorems in `trusted/z3-py/translator.lisp` enable
   `translate-type`, whose body looks up the type tables.  All these proofs
   actually need is the `stringp-of-translate-type` returns theorem; opening
   the definition instead makes ACL2 case-split over every table entry.

2. `verified/expand-cp.lisp` tests `(member-equal fn-call *SMT-basics*)`
   directly inside the `expand` function, so proofs about `expand` unroll
   the membership test into one case per list entry.

Not much performance change just from this, but it enables adding more
operators without proof time blowup.

### DIFF
--- a/books/projects/smtlink/trusted/z3-py/translator.lisp
+++ b/books/projects/smtlink/trusted/z3-py/translator.lisp
@@ -942,8 +942,7 @@
     :returns (translated
               paragraphp
               :hints (("Goal"
-                       :in-theory (enable translate-symbol translate-type
-                                          paragraphp wordp))))
+                       :in-theory (enable translate-symbol paragraphp wordp))))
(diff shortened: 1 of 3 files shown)


MattKaufmann (approved):
I'm not familiar with the code for Smt link, but this looks fine to me, and I'm happy to defer to @mrgreenstreet .

bendyarm:
Received a go-ahead from @mrgreenstreet ; merging.

PR #2003, opened 2026-08-06
[smtlink] route declarations through the _SMT_ interface object

bendyarm:
This PR lets a different SMT solver be substituted, via `custom-smt-cnf`,
without editing the translator: the `_SMT_` interface object becomes the
single point of contact between a generated script and the solver.

Currently (before this PR), declarations are the one exception.
Generated scripts declare variables and uninterpreted functions with `z3.Const(...)` and
`z3.Function(...)`, and nothing binds the name `z3` on purpose: it resolves
only because `from z3 import *` happens to re-export the z3 package's
internal `z3.z3` submodule attribute, an undocumented accident of
z3-solver's packaging.  Everything else in a generated script already goes
through `_SMT_`.

This change emits `_SMT_.Const(...)` and `_SMT_.Function(...)` instead, and
adds the two methods to `ACL22SMT`.  Generated scripts are otherwise
unchanged.

Custom interface classes that extend `ACL22SMT` (like the existing
`RewriteExpt.py`) inherit the new methods.  A from-scratch class would need
to add them, and fails with a Python error rather than a wrong result if it
doesn't.

Also rewords a "Z3 failure" error message in `run.lisp`, since the code it
reports on is solver-generic.

### DIFF
--- a/books/projects/smtlink/trusted/run.lisp
+++ b/books/projects/smtlink/trusted/run.lisp
@@ -42,7 +42,7 @@
     :verify-guards nil
     (b* (((mv exit-status lines state) (SMT-run fname smt-conf state))
          ((unless (equal exit-status 0))
-          (mv (er hard? 'SMT-run=>SMT-interpret "Z3 failure: ~q0" lines) state))
+          (mv (er hard? 'SMT-run=>SMT-interpret "SMT solver failure: ~q0" lines) state))
          ((if (equal lines nil))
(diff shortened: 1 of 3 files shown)


MattKaufmann (approved):
I'm not familiar with the code for Smtlink, but this looks fine to me, and I'm happy to defer to @mrgreenstreet -- especially since, unlike PR #2002, I don't have any sense of what is going on here at the implementation level.

bendyarm:
Received a go-ahead from @mrgreenstreet ; merging.

PR #2004, opened 2026-08-07
Add double-ended queue library

akitaki79:


books/kestrel/data/deque/deque.lisp
```diff
@@ -0,0 +1,695 @@
...
+    (xdoc::p
+      "A <b>deque</b> (double-ended queue) is a finite sequence that supports
+       inserting and removing elements at <i>both</i> ends, in constant
+       amortized time O(1).  This is an implementation in the style of
+       Okasaki's banker's deque: a deque is a front list consed together with
+       a reversed rear list, plus the cached length of each side so the balance
+       check never recomputes @('len').  Whenever one side grows more than
+       three times the other, the deque is rebalanced by splitting the elements evenly.")
+    (xdoc::p
+      "@(see dequep) recognizes a well-formed deque and @(see deque->list)
```
gjurgensen: You should use `@(tsee ...)` instead of `@(see ...)` for code. `tsee` renders the linked topic in code typesetting instead of regular prose typesetting.

books/kestrel/data/deque/deque.lisp
```diff
@@ -0,0 +1,695 @@
...
+  :parents (deque)
+  :short "Add an element to the back of a deque.")
+
+(defxdoc front
+  :parents (deque)
+  :short "The front (leftmost) element of a deque.")
+
+(defxdoc back
+  :parents (deque)
+  :short "The back (rightmost) element of a deque.")
```
gjurgensen: It might be good to say what these functions return on an empty deque when there is no true front/back (I think the answer is `nil`; it is easy to check).

books/kestrel/data/deque/deque.lisp
```diff
@@ -0,0 +1,695 @@
...
+
+(defxdoc+ deque
+  :parents (data::data-lib)
+  :short "Purely functional double-ended queues (deques)"
+  :long
+  (xdoc::topstring
+    (xdoc::p
+      "A <b>deque</b> (double-ended queue) is a finite sequence that supports
+       inserting and removing elements at <i>both</i> ends, in constant
+       amortized time O(1).  This is an implementation in the style of
```
gjurgensen: Nitpick: technically it is only guaranteed constant amortized time under "single-threaded" or "linear" usage. Amortization may break down otherwise.

So I would say just amend this to say "in constant amortized time O(1) under single-threaded usage".

gjurgensen (changes requested):

gjurgensen:
I know I originally suggested using the `data` package, but now I'm thinking this should have its own `deque` package. The reason is that function names like `front` and `back` are too generic to take the names in the full `data` package.

gjurgensen:
Some linear rules like these would be nice. They would help with termination proofs that decrease on a deque.
```lisp
(defthm deque-size-of-pop-front-linear
  (implies (not (deque-emptyp dq))
           (< (deque-size (pop-front dq))
              (deque-size dq)))
  :rule-classes :linear)

(defthm deque-size-of-pop-back-linear
  (implies (not (deque-emptyp dq))
           (< (deque-size (pop-back dq))
              (deque-size dq)))
  :rule-classes :linear)

(defthm posp-of-deque-size-when-not-deque-emptyp
  (implies (not (deque-emptyp dq))
           (< 0 (deque-size dq)))
  :rule-classes :linear)
```

gjurgensen:
I think a rule like this is missing:
```lisp
(defthm push-back-of-back-and-pop-back
  (implies (not (deque-emptyp dq))
           (deque-equiv (push-back (back dq) (pop-back dq)) dq)))
```

books/kestrel/data/deque/deque.lisp
```diff
@@ -0,0 +1,695 @@
...
+(defthm back-is-last-of-deque->list
+  (equal (back dq)
+         (car (last (deque->list (deque-fix dq)))))
+  :hints (("Goal" :in-theory (e/d (back deque->list dequep balancedp deque-fix)
+                                  (acl2::car-of-reverse-list)))))
+
+; Remove the front element.
+(defund pop-front (dq)
+  (declare (xargs :guard (and (dequep dq)
+                              (not (deque-emptyp dq)))
```
gjurgensen: I think the `(not (deque-emptyp dq))` conjunct of the guard can be removed. That would make it match the guard of `front`.

Same for `pop-back`.

books/kestrel/data/deque/deque.lisp
```diff
@@ -0,0 +1,746 @@
...
+  :long "<p>Returns @('nil') when the deque is empty (there is no back
+         element).</p>")
+
+(defxdoc pop-front
+  :parents (deque)
+  :short "Remove the front element of a non-empty deque.")
+
+(defxdoc pop-back
+  :parents (deque)
+  :short "Remove the back element of a non-empty deque.")
```
gjurgensen: Now that the guards no longer require emptiness, we should remove the "non-empty" qualifiers. Like `front` and `back`, we should describe the behavior on empty queues (I believe popping an empty queue just returns an empty queue).

gjurgensen (changes requested):

gjurgensen (approved):
LGTM. I'll merge after doing a full build.

PR #2005, opened 2026-08-10
Defirrelevant xdoc

bendyarm:
Add xdoc

### DIFF
--- a/books/std/util/defirrelevant.lisp
+++ b/books/std/util/defirrelevant.lisp
@@ -25,6 +25,66 @@
                            (long 'nil long-p))
   :parents (std/util)
   :short "Define an irrelevant value of a type."
+  :long
+  (xdoc::topstring
+   (xdoc::p
+    "This macro defines a nullary function, named @('name'),
+     that returns a fixed value of a type:
+     the value of the term @('body'),
+     which must satisfy the predicate @('type')
+     (this is proved as a @(tsee define) @(':returns') theorem).
+     The @(':parents'), @(':short'), and @(':long') inputs
+     are for the XDOC topic of the function;
+     if @(':long') is absent, a default is generated.")
+   (xdoc::codeblock
+    "(defirrelevant name"
+    "  :type ...     ; recognizer of the type"
+    "  :body ...     ; a term whose value has the type"
+    "  :parents ...  ; optional"
+    "  :short ...    ; a string"
+    "  :long ...     ; optional"
+    "  )")
+   (xdoc::p
+    "The purpose of the function is to provide
+     a value of the type whose specific identity does not matter;
+     hence `irrelevant':
+     it is irrelevant which value of the type this is,
+     not that the function is unused.
+     Typical uses are
+     to fill the non-error result slots of a function
+     that is returning an error
+     (e.g. with @(see error-value-tuples),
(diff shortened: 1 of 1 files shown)


books/std/util/defirrelevant.lisp
```diff
@@ -25,6 +25,66 @@
...
+    "A caveat: ACL2 computes a type-prescription rule
+     for the function at admission time,
+     and if @('body') is a constant, like @('0') or @('nil'),
+     that (enabled) rule determines the value exactly,
+     defeating the opacity just described.
+     If opacity matters,
+     use a non-constant @('body'),
+     e.g. a call of a constructor,
+     whose type-prescription rule only constrains
+     the shape of the value."))
```
gjurgensen: I'd suggest we use the new `:type-prescription :none` option. If we are worried that might break some existing proofs, we could at least document here that `defirrelevant` may eventually generate a `define` with this option.

books/std/util/defirrelevant.lisp
```diff
@@ -25,6 +25,66 @@
...
+    "  :parents ...  ; optional"
+    "  :short ...    ; a string"
+    "  :long ...     ; optional"
+    "  )")
+   (xdoc::p
+    "The purpose of the function is to provide
+     a value of the type whose specific identity does not matter;
+     hence `irrelevant':
+     it is irrelevant which value of the type this is
+     (but the fact that the function is unused is not irrelevant).
```
gjurgensen: The meaning of this parenthetical isn't clear to me. When we say "the function is unused", do we mean "the value of the function is not used", i.e. users should only rely on the type, not the particular value?

gjurgensen (approved):

gjurgensen:
Another stray thought: we could generate a theory invariant that complains when the abstraction-breaking rules (the definition and executable-counterpart rules) are enabled. Probably overkill.

Anyway, not something for this PR.

books/std/util/defirrelevant.lisp
```diff
@@ -25,6 +25,66 @@
...
+     (this is proved as a @(tsee define) @(':returns') theorem).
+     The @(':parents'), @(':short'), and @(':long') inputs
+     are for the XDOC topic of the function;
+     if @(':long') is absent, a default is generated.")
+   (xdoc::codeblock
+    "(defirrelevant name"
+    "  :type ...     ; recognizer of the type"
+    "  :body ...     ; a term whose value has the type"
+    "  :parents ...  ; optional"
+    "  :short ...    ; a string"
```
acoglio, line 44: Maybe we should make :short optional too, but perhaps in a separate change (not this PR).

books/std/util/defirrelevant.lisp
acoglio: It's unclear to me as well.

books/std/util/defirrelevant.lisp
acoglio: I think it's a good idea. Not necessarily for this PR.

acoglio (approved):
It looks fine, just maybe one clarification needed for this PR.

bendyarm:
Since a lot of books recertify when `defirrelevant` is changed, and since regression
succeeds when I add `:type-prescription :none`, I decided to put this in as well.

PR #2006, opened 2026-08-10
Fix broken XDOC reference

Sarah-Scott:
Fix broken XDOC reference to `parse-file-as-json` in `deserialize-from-file.lisp`. XDOC does not exist for that function so removed `tsee` for it.


bendyarm:
Regression successful; merged.

PR #2009, opened 2026-08-11
add fixers to filepath

akitaki79:


books/kestrel/filepath/paths.lisp
gjurgensen: I think this should use the new `path-kind-fix` and `segment-list-fix`.

gjurgensen (changes requested):

books/kestrel/filepath/paths.lisp
```diff
@@ -346,10 +403,13 @@
   :hints (("Goal" :in-theory (enable resolve))))
 
 (defthm resolve-associative
-  (equal (resolve (resolve x y) z)
-         (resolve x (resolve y z)))
+  (implies (and (path-p x)
+                (path-p y)
+                (path-p z))
+           (equal (resolve (resolve x y) z)
+                  (resolve x (resolve y z))))
```
gjurgensen: This rule should be able to stay unconditional. Make sure everything is fixing its path and kind arguments. Then prove rule of the form.
```lisp
(equal (resolve (path-fix x) y)
       (resolve x y))
```

books/kestrel/filepath/paths.lisp
```diff
@@ -216,11 +259,23 @@
...
+  :hints (("Goal" :in-theory (enable path-fix))))
+
+(defthm path-fix-when-path-p
+  (implies (path-p x)
+           (equal (path-fix x) x))
+  :hints (("Goal" :in-theory (enable path-fix))))
+
 (defund absolute-path-p (x)
   (declare (xargs :guard (path-p x)))
   (eq (path-kind x) :absolute))
```
gjurgensen: Functions like these should fix their path arguments.

gjurgensen (commented):
I left some comments for possible improvements. I understand that we are short on time though, so I'm happy to merge the PR as it currently is if we don't have time to address them.

gjurgensen (approved):
Great, I'll merge after checking the build.

PR #2011, opened 2026-08-12
[deffold-map] Make guard proofs more robust.

acoglio:


books/kestrel/fty/deffold-map.lisp
```diff
@@ -367,6 +367,202 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define deffold-map-gen-sum-require-rules ((sum flexsum-p)
+                                           (suffix symbolp)
+                                           (targets symbol-listp)
+                                           (overrides alistp)
+                                           (fty-table alistp))
+  :returns (rules symbol-listp)
+  :short "Generate the names of the rules needed to verify the guards of
+          the map function for a product, sum, or option type
```
gjurgensen: The `:short` mentions option types in a way that sounds like they are supported., but the long says these are skipped.

books/kestrel/fty/deffold-map.lisp
```diff
@@ -1098,35 +1331,70 @@
...
+    "If any product or summand of the types in the clique
+     has a dependent requirement (i.e. a @(':require'))
+     and at least one field with a map function,
+     the guards of the generated functions
+     cannot be verified as part of the @(tsee define) or @(tsee defines),
+     for two reasons:
+     the guard obligations from the boilerplate constructor calls
+     need rules that are disabled by default
+     (the @('<constructor>-requirements') theorems generated by FTY,
+     and the theorems that accompany the list map functions);
```
gjurgensen: I believe the `<constructor>-requirements' rules are actually enabled by default. Of course, the user might disable them, so the implementation behavior is still reasonable. Just might want to tweak the documentation accordingly.

gjurgensen (approved):

PR #2013, opened 2026-08-14
Add theorems showing when fresh names are valid identifiers.

westfold:
Use this to show uniquify-names preserves wellformedness and monomorphize created names are valid.


acoglio:
I've added @bendyarm as reviewer since this also affects the `identifier-syntax.lisp` file.

books/kestrel/remora/fresh-variable-operations.lisp
```diff
@@ -819,6 +820,41 @@
...
+; identifier-syntax), whose conditions are that the suffix consists of
+; ASCII identifier continuation characters and that the extension is
+; not a reserved keyword; the facts that discharge those conditions for
+; a digit suffix live with IDENTIFIER-SYNTAX as well.  All that is
+; specific to this book is the connection to STR::NAT-TO-DEC-STRING:
+; the decimal representation of an index is a non-empty sequence of
+; digit code points.
+
+(local
+ (defsection expr-var-with-index-support
```
acoglio: It's odd for a defsection (1) to have no :short or :long and (2) to be local.

books/kestrel/remora/fresh-variable-operations.lisp
```diff
@@ -819,6 +820,41 @@
...
+
+   (defruled digit-bytes-of-nat-to-dec-string
```
acoglio: This is not specific to expr-var-with-index, and with fresh variables. It should go with dec-nat-listp; it's a property of dec-nat-listp, involving some string library functions.

books/kestrel/remora/fresh-variable-operations.lisp
```diff
@@ -819,6 +820,41 @@
...
+     (b* ((bytes (acl2::string=>nats (str::nat-to-dec-string n))))
+       (and (dec-nat-listp bytes)
```
acoglio: It may be more useful to split this conjunction into two theorems: the 1st conjunct a rewrite rule, and the 2nd conjunct a type prescription.

books/kestrel/remora/fresh-variable-operations.lisp
```diff
@@ -991,4 +1052,19 @@
   ///
 
   (defret fresh-expr-var-is-fresh
-    (not (set::in var (string-sfix used)))))
+    (not (set::in var (string-sfix used))))
+
+  (defrule valid-identifier-string-p-of-fresh-expr-var-loop
```
acoglio: This should go under the loop function itself.

books/kestrel/remora/fresh-variable-operations.lisp
```diff
@@ -991,4 +1052,19 @@
...
+    (implies (and (stringp prefix)
+                  (valid-identifier-string-p prefix))
+             (valid-identifier-string-p (fresh-expr-var-loop prefix index used)))
+    :induct (fresh-expr-var-loop prefix index used)
```
acoglio: Most likely `:induct t` would suffice.

books/kestrel/remora/fresh-variable-operations.lisp
```diff
@@ -991,4 +1052,19 @@
...
+    :enable fresh-expr-var-loop)
+
+  ; The generated variable's name is a legal Remora identifier if the
```
acoglio, line 1138: I would turn this comment into text in the :long, of even drop it completely — it's quite clear what the theorem says.

books/kestrel/remora/identifier-syntax.lisp
```diff
@@ -333,3 +333,374 @@
...
+    (null x)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defsection identifier-extension-theorems
+  :short "Theorems about extending identifiers with suffixes."
+
+  ;; The code-point level.
+
+  (defruled id-continue-list-p-of-append
```
acoglio: This should go with `id-continue-list-p`: it's a property of that function, no specific to identifier-extension-theorems (the title of this defsection).

If that function were defined with `std::deflist`, we wouldn't need to write down this kind of theorem. @bendyarm

books/kestrel/remora/identifier-syntax.lisp
```diff
@@ -333,3 +333,374 @@
...
+    (equal (id-continue-list-p (append a b))
+           (and (id-continue-list-p a)
+                (id-continue-list-p b)))
+    :induct (len a)
+    :enable id-continue-list-p)
+
+  (defruled id-continue-list-p-when-dec-nat-listp
```
acoglio: This is also not specific to identifier extensions. It could go with dec-nat-list or id-continue-list-p (whichever comes last), or in a new defsection with theorems that relate the different predicates over code points (maybe there are other theorems like this).

books/kestrel/remora/identifier-syntax.lisp
```diff
@@ -333,3 +333,374 @@
...
+  ;; hypothesis of the self-synchronization rule.
+
+  (defruled utf8-partition-when-decoding-succeeds
+    (implies (nat-listp (acl2::utf8=>ustring bytes))
+             (mv-nth 0 (acl2::utf8-partition bytes)))
+    :enable acl2::utf8=>ustring)
+
+  ;; The bytes of a concatenation of strings.
+
+  (defruled string=>nats-of-string-append
```
acoglio: This should be in the strings library, with `acl2::string=>nats`.

books/kestrel/remora/identifier-syntax.lisp
```diff
@@ -333,3 +333,374 @@
...
+    :use ((:instance not-remora-keyword-p-when-last-is-digit
+                     (cps (append cps d)))
+          (:instance car-of-last-when-dec-nat-listp (x d)))
+    :enable last-of-append-when-consp
+    :prep-lemmas
+    ((defruled last-of-append-when-consp
+       (implies (consp b)
+                (equal (last (append a b)) (last b)))
+       :induct (len a))
+     (defruled car-of-last-when-dec-nat-listp
```
acoglio: This theorem and the next suggest that it would be better to define dec-nat-listp as (1) a dec-natp predicate on single elements and (2) a list of those (preferably with std::deflist, so we get theorems for free).

books/kestrel/remora/identifier-syntax.lisp
```diff
@@ -333,3 +333,374 @@
...
+;; byte level as well: a digit suffix is ASCII, hence its own UTF-8
+;; encoding, and UTF-8 is self-synchronizing, so decoding the bytes of
+;; an extended identifier does not cross the boundary between the
+;; original bytes and the suffix (see
+;; UTF8=>USTRING-OF-APPEND-WHEN-UTF8-PARTITION in unicode/utf8-decode).
+;;
+;; All rules are left disabled: they are aimed at specific proofs about
+;; name generation, not at general use.
+
+(define dec-nat-listp (x)
```
acoglio: Since this is similar in spirit to some of the predicates defined earlier in the file (i.e. characterize certain categories of code points), I would move this up, split it into one for elements and one for lists, add natp and nat-listp guards, and rename to something like dec-digitp and dec-digit-listp.

books/kestrel/remora/identifier-syntax.lisp
```diff
@@ -333,3 +333,374 @@
...
+(defsection identifier-extension-theorems
+  :short "Theorems about extending identifiers with suffixes."
+
+  ;; The code-point level.
+
+  (defruled id-continue-list-p-of-append
+    (equal (id-continue-list-p (append a b))
+           (and (id-continue-list-p a)
+                (id-continue-list-p b)))
+    :induct (len a)
```
acoglio: Most likely `:induct t` would work.

books/kestrel/remora/identifier-syntax.lisp
```diff
@@ -333,3 +333,374 @@
...
+       (implies (acl2::ustring? x)
+                (mv-nth 0 (acl2::utf8-partition (acl2::ustring=>utf8 x))))
+       :induct (acl2::ustring=>utf8 x)
+       :enable (acl2::utf8-partition acl2::ustring=>utf8 acl2::ustring?))))
+
+  ;; A successful decoding is exactly a successful partitioning; this is
+  ;; how VALID-IDENTIFIER-STRING-P of an identifier supplies the other
+  ;; hypothesis of the self-synchronization rule.
+
+  (defruled utf8-partition-when-decoding-succeeds
```
acoglio: This probably goes into the Unicode library, or at least into a Unicode library extension file in this remora tree.

acoglio (changes requested):

PR #2014, opened 2026-08-17
[Remora] Haskell unchecked AST predicate

Sarah-Scott:
Added a new file `abstract-syntax-haskell.lisp`, similar to `abstract-syntax-core.lisp`, but instead characterized the subset of our ACL2 ASTs that corresponds exactly to the Haskell implementation ASTs. Predicates end in `-hip` for "Haskell implementation predicate", but I suspect we may want to add to this suffix due to the below explanation (partially copied from the XDOC).

This PR only targets Haskell's unchecked ASTs, `ProgBase TypeExp TypeParamExp NoInfo Text` (aliased as `UncheckedProg`) along with its constituent AST types. That is the instantiation our AST fixtypes mirror: types are source-level (they include array type variables, their quantifiers are n-ary, and they do not separate atom types from array types), type parameters are source-level as well (they include array type parameters), type annotations are empty, and variables are names without tags.

Characterizing the subset of our ACL2 ASTs that corresponds to the Haskell implementation unique ASTs and especially checked ASTs will be trickier and a bit more involved. We could end predicates with `-hip-unchecked`, `-hip-unique`, and `-hip-checked` since we will almost certainly need distinct predicates for each of the three Haskell AST instantiations.

### DIFF
--- a/books/kestrel/remora/abstract-syntax-haskell.lisp
+++ b/books/kestrel/remora/abstract-syntax-haskell.lisp
@@ -1,0 +1,420 @@
+; Remora Library
+;
+; Copyright (C) 2026 Kestrel Institute (http://www.kestrel.edu)
+;
+; License: A 3-clause BSD license. See the LICENSE file distributed with ACL2.
+;
+; Author: Sarah Johnson
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(in-package "REMORA")
+
+(include-book "abstract-syntax-trees")
+
+(include-book "kestrel/fty/deffold-reduce" :dir :system)
+
+(local (include-book "kestrel/utilities/ordinals" :dir :system))
+
+(acl2::controlled-configuration)
+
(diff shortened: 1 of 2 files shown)


books/kestrel/remora/abstract-syntax-haskell.lisp
```diff
@@ -0,0 +1,420 @@
...
+     the declarations of [impl]'s AST types alone,
+     not by which of their values any [impl] pass happens to build.
+     An AST of ours is in the subset when its constructors and fields
+     can be matched with those of [impl]'s constructors,
+     where a field of ours holding a list
+     matches a field of [impl] holding a single value
+     just when the list is a singleton,
+     and matches a field of [impl] holding a @('NonEmpty')
+     just when the list is not empty.
+     Note that this does not require the correspondence to be injective:
```
acoglio, line 55: In a future PR, we could consider making it injective, by choosing a particular form for our ASTs among the equivalent ones. For instance, we could choose `(-> T R)` over `(-> (T) R)`, even though they are equivalent. The advantage is that we can have an isomorphism, with inversion theorems, between this subset of our ASTs and the JSON representation of the Haskell ASTs.

acoglio (approved):

PR #2015, opened 2026-08-19
[definductive] Add minimality and induction scheme

gjurgensen:
A minimality condition was added to each relation. It does not change the meaning of the relation, but it allows the choice function underlying the existential to be used in the induction scheme. Note that non-recursive functions get the minimality condition as well, even though they don't get an induction scheme. It is added nonetheless for uniformity.

On singly-recursive relations, and induction rule-class rule is added to suggest the induction scheme. For mutually recursive relations, we generate a flag from the induction scheme clique.


books/std/util/definductive-doc.lisp
```diff
@@ -452,6 +452,28 @@
...
+      "@('p[1]-proof-minimalp')"
+      "@('...')"
+      "@('p[n]-proof-minimalp')")
+     (xdoc::p
+      "Predicates saying that a proof tree is minimal,
+       i.e. that no valid proof tree of the same conclusion
+       has a smaller count.
+       Each takes a proof and the arguments of the conclusion,
+       and universally quantifies over the other proof trees.")
+     (xdoc::p
```
acoglio: I would remove this paragraph since it describes more of an implementation detail than something really user-facing. I think that the rest of the documentation does not provide similar details.

books/std/util/definductive-doc.lisp
```diff
@@ -463,13 +485,152 @@
...
        there is a proof that is valid for those arguments,
        which are passed to @('p[i]-proof-validp').")
+     (xdoc::p
+      "The proof whose existence is asserted is also required to be minimal,
+       according to @('p[i]-proof-minimalp').
+       This does not change the meaning of the predicates,
+       because a minimal valid proof tree exists
+       exactly when any valid proof tree exists.
+       It makes the witness @('p[i]-proof') a minimal proof,
+       which is what supports reasoning by induction on proofs.
```
acoglio:
```suggestion
       which support reasoning by induction using the @('p[i]-induct') functions (see below).
```

Rationale: one can reason by induction on the proof-validp predicates, so this is more specific.

books/std/util/definductive-doc.lisp
```diff
@@ -463,13 +485,152 @@
...
      (xdoc::p
       "These predicates are currently not guard-verified,
        because they call the non-guard-verified
        @('p[i]-proof-validp') predicates.
        The @('p[i]') predicates do not have fixing theorems,
        because the formals are currently untyped."))
 
+    (xdoc::desc
+     (list
+      "@('p[1]-when-valid-proof')"
```
acoglio: Should we use the `p[i]-when-proof-validp` naming convention?

books/std/util/definductive-doc.lisp
```diff
@@ -463,13 +485,152 @@
...
+      "Theorems saying that @('(p[i] x[i,1] ... x[i,m[i]])') holds
+       if there is any valid proof tree for those arguments,
+       whether or not that proof tree is minimal.")
+     (xdoc::p
+      "Because of the minimality requirement described above,
+       the @('p[i]-suff') theorems that @(tsee defun-sk) generates
+       require the proof tree to be minimal,
+       which is inconvenient.
+       These theorems drop that requirement,
+       and are therefore exactly the @('p[i]-suff') theorems
```
acoglio: I would omit the "historical" context. I would just say something like "These theorems drop that requirement, to facilitate their application."

books/std/util/definductive-doc.lisp
```diff
@@ -463,13 +485,152 @@
...
+       that this macro generated
+       before the minimality requirement was introduced;
+       the @('p[i]-suff') theorems are disabled,
```
acoglio: Are these new theorems enabled? I thought that the suff theorems were disabled regardless, because we generate define-sk events.

books/std/util/definductive-doc.lisp
```diff
@@ -463,13 +485,152 @@
...
+       since these stronger ones replace them.")
+     (xdoc::p
```
acoglio: I would probably omit this paragraph, since it's not that important for using definductive. We don't explain how the other generated theorems are proved.

books/std/util/definductive-doc.lisp
```diff
@@ -463,13 +485,152 @@
...
+       that is what lets these theorems apply on their own
+       in those measure proofs.")
+     (xdoc::p
+      "These are generated only for the recursive predicates.
+       If XDOC is generated, they go in the @(tsee defsection)
+       mentioned just above."))
+
+    (xdoc::desc
+     (list
+      "@('p[1]-ind')"
```
acoglio: I would go for the slightly longer and clearer `p[i]-induct`.

books/std/util/definductive-doc.lisp
```diff
@@ -463,13 +485,152 @@
...
+       are put in a @(tsee defsection) whose name is obtained by
+       extending the @('name') input with the suffix @('-valid-proofs')."))
+
+    (xdoc::desc
+     (list
+      "@('p[1]-proof-count-bound')"
+      "@('...')"
+      "@('p[n]-proof-count-bound')")
+     (xdoc::p
+      "Theorems saying that the witness proof tree is minimal in fact,
```
acoglio: It may be worth naming the witness function. In fact, it looks like the doc of the `p[i]` definitions fails to say that the name of the witness function is `p[i]-proof`, specified by `:slolem-name`. (A previous shortcoming, existing before this PR.) We should also say that although `p[i]-proof` is also the name of the fixtype, fixtypes and functions live in different namespaces.

books/std/util/definductive-doc.lisp
```diff
@@ -463,13 +485,152 @@
...
+      "@('...')"
+      "@('p[n]-ind')")
+     (xdoc::p
+      "Functions expressing rule induction for the predicates.
+       Each recurses on the arguments of the conclusions of the premises
+       of the rule that the witness proof tree used,
+       and so mentions no proof tree;
+       that is what lets it serve as an induction scheme
+       for the predicate itself.
+       Their values are irrelevant:
```
acoglio:
```suggestion
       The results of these functions are irrelevant:
```

books/std/util/definductive-doc.lisp
```diff
@@ -463,13 +485,152 @@
...
+       mentioned just above."))
+
+    (xdoc::desc
+     (list
+      "@('p[1]-ind')"
+      "@('...')"
+      "@('p[n]-ind')")
+     (xdoc::p
+      "Functions expressing rule induction for the predicates.
+       Each recurses on the arguments of the conclusions of the premises
```
acoglio: Don't they recurse on the proof witness?

books/std/util/definductive-doc.lisp
```diff
@@ -463,13 +485,152 @@
...
+       of the rule that the witness proof tree used,
+       and so mentions no proof tree;
+       that is what lets it serve as an induction scheme
+       for the predicate itself.
+       Their values are irrelevant:
+       only their recursive structure matters.")
+     (xdoc::p
+      "One event is generated for each clique of predicates.
+       For a clique of a single predicate, the event is a @(tsee defun).
```
acoglio: Should we use a `define`, for uniformity? We do use `defines` and `define` for non-singleton cliques.

books/std/util/definductive-tests.lisp
```diff
@@ -962,3 +962,112 @@
    :preds ((r a))
    :irules ((ax ((natp proof$))
                 (r proof$)))))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; The generated induction scheme supports rule induction.
+
+(must-succeed*
```
acoglio: Can we integrate this with the existing test form for RTC?

books/std/util/definductive-tests.lisp
```diff
@@ -962,3 +962,112 @@
...
+                                      r*-base
+                                      r*-refl
+                                      r*-trans))))
+)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; For a clique of two or more predicates, rule induction is via the flag macro.
+
+(must-succeed*
```
acoglio: Can we integrate this with the existing test for even/odd?

books/std/util/definductive.lisp
```diff
@@ -3861,14 +4043,88 @@
...
+     exactly when any valid proof exists;
+     it makes the witness a minimal proof,
+     which is what supports reasoning by induction on proofs.")
+   (xdoc::p
+    "That requirement weakens the @('p[i]-suff') theorem
+     that @(tsee defun-sk) generates,
+     which now applies only to minimal proofs.
+     So we generate @('p[i]-when-valid-proof'),
+     which applies to any valid proof,
+     and disable @('p[i]-suff') in favor of it;
```
acoglio: This is already the case, with `define-sk`. We are not doing anything special for it. Most of the events we generate are disabled.

books/std/util/definductive.lisp
```diff
@@ -3886,23 +4144,434 @@
...
+                            ,@concl-vars)
+                nil)))
+           (defrule ,when-valid-proof
+             (implies (and (,proof-validp ,proof ,@concl-vars)
+                           (,proofp ,proof))
+                      (,pred-info.name ,@concl-vars))
+             :hints (("Goal"
+                      :induct (,descend ,proof ,@concl-vars)
+                      :in-theory (enable ,suff ,minimalp))))))
+       (disable-suff-event `(in-theory (disable ,suff)))
```
acoglio: This should be unnecessary (see earlier comments).

books/std/util/definductive.lisp
```diff
@@ -3886,23 +4144,434 @@
...
+                                                    natp
+                                                    (:t ,count-fn)
+                                                    ,@count-natp-thms)))))
+              (if (and (,proofp ,proof)
+                       (,proof-validp ,proof ,@concl-vars)
+                       (not (,minimalp ,proof ,@concl-vars)))
+                  (,descend (,minimalp-witness ,proof ,@concl-vars)
+                            ,@concl-vars)
+                nil)))
+           (defrule ,when-valid-proof
```
acoglio: Do we want this enabled by default? I think all other theorems we generate are disabled.

books/std/util/definductive.lisp
```diff
@@ -3886,23 +4144,434 @@
...
+       (body1 (defind-gen-ind-fn-body
+                (car pred-infos) irule-infos clique-preds name))
+       (print-events
+        (and (evmac-input-print->= print :result)
+             (defind-gen-ind-fns-print-events pred-infos name)))
+       ((when (endp (cdr pred-infos)))
+        (b* ((induction-thm
+              (defind-induction-thm-name pred-info1.name name)))
+          (mv
+           (list `(define ,fn-name1 (,@pred-info1.formals)
```
acoglio, line 4457: It looks like it is generating a `define`, so we just need to update the user doc.

acoglio (changes requested):
Looks good. I left some comments.

books/std/util/definductive-doc.lisp
gjurgensen: What the doc was trying to say was that a proof object is not an argument to the induction scheme function. So by "recurses on", we mean "is an argument to the function". What we are recursing on are the arguments to the premises (specifically premises of relations in the same clique) of the rule witnessed by the proof tree chosen at that particular step. Hopefully the documentation is clearer now.

books/std/util/definductive-doc.lisp
gjurgensen: Thank you, good catch. For some reason I was under the impression that the generated `-suff` theorems were enabled. That was the reason I was disabling it, and also why the "replacement" `-when-valid-proof` rules were left enabled.

I'll remove the redundant disable and disable the `-when-valid-proof` rules by default.

books/std/util/definductive-doc.lisp
acoglio: Yes, I think it's fine. It's also easy to inspect the function, if the user has doubts.

acoglio (approved):

PR #2016, opened 2026-08-19
[C$] Extend C array types.

gjurgensen:
This adds two kinds of complete array types: constant length and non-constant length. The standard does not use these terms, but it is the necessary category split to determine compatibility (under our current interpretation that initializer-inferred array lengths are treated the same as syntactically provided integer constant expression sizes).

An "unknown complete" array kind represents either of the two above cases. We use this when we can't determine whether an array is constant or non-constant length, i.e. when we can determine whether the size expression is an integer constant expression.

Finally, there is a kind for incomplete array types.

Some incidental changes:
- Compound literals need to store their type, because they are *not* necessarily the provided type. E.g., the type of `(int []){1, 2}` is *not* `int []`, it is `int [2]`.
- When a declaration declares something that was already in scope, we store the composite type, not the new type. This is what prevents code like:
```c
int [3] x;
int [] x;
int [4] x;
```


books/kestrel/c/syntax/types.lisp
```diff
@@ -168,21 +231,8 @@
...
-        using @('nil') to mean `unknown size',
-        making the type partially unknown,
-        as opposed to totally unknown
-        as in the @(':unknown') case described below.
-        We use a positive integer instead of a natural
-        because arrays are not empty [C17:6.2.5/20],
-        and so a known constant size must not be 0
-        (but we can revise this if we discover that
-        array types may need to track a 0 size).")
+        and an array kind.
```
acoglio: Stray period in ``element type.''.

books/kestrel/c/syntax/types.lisp
```diff
@@ -1393,6 +1443,139 @@
...
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(define type-array-kind-compatible-p ((x type-array-kindp)
+                                      (y type-array-kindp))
+  :returns (yes/no booleanp)
+  :short "Check whether two array kinds are potentially compatible."
+  :long
+  (xdoc::topstring
+   (xdoc::p
+    "Read literally, C17 and C23 require that two array types be compatible
```
```suggestion
    "Read literally, [C17] and [C23] require that two array types be compatible
```

books/kestrel/c/syntax/validator.lisp
```diff
@@ -5690,7 +5841,20 @@
                 types
                 vstate))
        :array
-       (b* ((type (make-type-array :of type :size nil)) ; TODO: size
+       (b* ((type (make-type-array
+                    :of type
+                    :kind (if dirabsdeclor.size?
+                              (type-array-kind-unknown-complete)
+                            (type-array-kind-incomplete))))
+            ((when (and (not dirabsdeclor.declor?)
```
acoglio: Is this early-exit branch necessary? Would the code after this do something different, in the case of absent declor and size?

books/kestrel/c/syntax/types.lisp
```diff
@@ -116,6 +116,69 @@
...
+     an array completed by its initializer as @(':const-len'),
+     but does not yet derive its numerical length from that initializer,
+     so its @('len') field is @('nil').
+     We use a positive integer instead of a natural
+     because arrays are not empty [C17:6.2.5/20] [C23:6.2.5/25],
+     and so a determined constant length must not be 0
+     (but we can revise this if we discover that
+     array types may need to track a 0 length).")
+   (xdoc::p
+    "The @(':nonconst-len') case represents complete array types
```
acoglio, line 141: It looks like this includes the case of `*` as size. If so, it could be helpful to mention that here.

books/kestrel/c/syntax/tests/validator.lisp
```diff
@@ -1909,3 +1909,90 @@ void g(void) {
...
+                      (trans-item-declon->declon item)))
+             (initdeclor (first (declon-declon->declors declon)))
+             (type (init-declor-vinfo->type
+                    (init-declor->info initdeclor))))
+          (equal type
+                 (make-type-array
+                  :of (type-sint)
+                  :kind (type-array-kind-unknown-complete)))))
+
+;; An ordinary identifier expression gives a nonconstant length in C17.
```
acoglio: Not also in C23?

books/kestrel/c/syntax/validator.lisp
```diff
@@ -2806,6 +2807,194 @@
...
+     whether the expression is an integer constant expression.")
+   (xdoc::p
+    "For an integer constant expression,
+     we also attempt to evaluate the length.
+     If constant evaluation produces a positive integer, we retain it;
+     otherwise, we leave the constant length unknown.
+     In particular, this function does not enforce the constraint that
+     a constant size expression have a value greater than zero
+     [C17:6.7.6.2/1] [C23:6.7.7.3/1]."))
+  (if (expr-unambp expr)
```
acoglio: Should this be a guard?

books/kestrel/c/syntax/validator.lisp
```diff
@@ -2806,6 +2807,194 @@
...
+      (case (expr-ice-p expr t (ienv->dialect ienv))
+        ((t) (b* ((integer? (value-to-integer
+                             (const-eval-expr expr ienv))))
+               (make-type-array-kind-const-len
+                :len (and (posp integer?) integer?))))
```
acoglio: Eventually, should it be an error if the size evaluates to a nonpositive integer?

books/kestrel/c/syntax/validator.lisp
```diff
@@ -2806,6 +2807,194 @@
...
+        (otherwise (type-array-kind-unknown-complete)))
+    (type-array-kind-unknown-complete)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define pointer-type-path-aux
+  ((pointers typequal/attribspec-list-listp)
+   (path true-listp))
+  :returns (new-path true-listp)
+  :short "Prepend a pointer-derived type path to a type path."
```
acoglio: This notion of path comes a bit out of the blue when reading the file in order. It could be good to have a forward reference to `type-array-change-kind-at-path`, which is where the notion of path is explained. Same for the other functions below, and before `type-array-change-kind-at-path`.

acoglio: Perhaps we could use a stricter type for paths, like keyword-listp, or even a list of a deftagsum with the (few) alternatives. For more abstraction/clarity.

books/kestrel/c/syntax/abstract-syntax-operations.lisp
```diff
@@ -508,6 +508,110 @@
...
+      :array (1+ (dirdeclor-derived-type-depth dirdeclor.declor))
+      :array-static1 (1+ (dirdeclor-derived-type-depth dirdeclor.declor))
+      :array-static2 (1+ (dirdeclor-derived-type-depth dirdeclor.declor))
+      :array-star (1+ (dirdeclor-derived-type-depth dirdeclor.declor))
+      :function-params (1+ (dirdeclor-derived-type-depth dirdeclor.declor))
+      :function-names (1+ (dirdeclor-derived-type-depth dirdeclor.declor)))
+    :measure (dirdeclor-count dirdeclor))
+
+  :ruler-extenders :all
+  :verify-guards nil
```
acoglio: Does `:verify-guards :after-returns` not work with these functions?

books/kestrel/c/syntax/types.lisp
```diff
@@ -168,21 +231,8 @@
       (xdoc::li
        "An array type [C17:6.2.5/20],
         derived from the ``element type.'',
```
acoglio: Stray period.

acoglio (approved):
Looks good. Left some suggestions.

PR #2017, opened 2026-08-19
[REMORA] add prim-fold

ltmquan:


books/kestrel/remora/expression-values-and-environments.lisp
```diff
@@ -502,6 +502,19 @@
                       (dval nat)
                       (sval nat-list)
                       (fval expr-value)))
+    (:fold ())
+    (:fold-t ((tval type-value)))
+    (:fold-t-t2 ((tval type-value) (t2val type-value)))
```
acoglio: For consistency with the ones above, the fields of the following summands should be one per line. (I also find it more readable.)

books/kestrel/remora/expression-values-and-environments.lisp
acoglio:
```suggestion
  (defrule expr-value-wfp-of-expr-value-stages
```

acoglio:
```suggestion
  (defrule primop-value-wfp-of-expr-value-stage-constructors
```

books/kestrel/remora/evaluation.lisp
```diff
@@ -2385,11 +2385,46 @@
...
                                     arg (1- limit))
+       :fold-t-t2-d-s-s2-f-z (prim-fold op.tval op.t2val op.dval op.sval
+                                        op.s2val op.fval op.zval
+                                        arg (1- limit))
        :otherwise (reserr (impossible))))
     :measure (nfix limit))
 
   ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+  (define prim-reduce-loop ((acc expr-valuep)
```
acoglio: Since it's a loop for both reduce and fold, this should be called something like prim-reduce/fold-loop.

acoglio (approved):
Looks, just some minor suggestions.

acoglio:
It doesn't look like the suggestions were addressed: maybe they weren't committed or pushed?

ltmquan:
Yes, I forgot to add the changes. They should be in now.

acoglio (approved):

PR #2019, opened 2026-08-24
[REMORA] add `prim-reify-dim` and `prim-reify-shape`

ltmquan:
For `prim-reify-shape`, it produces a box, so the function body is calling `make-type-value-sigma` and `make-expr-value-box` directly. Do you think this is the correct approach?

### DIFF
--- a/books/kestrel/remora/primitives-evaluation-on-ispaces.lisp
+++ b/books/kestrel/remora/primitives-evaluation-on-ispaces.lisp
@@ -13,6 +13,9 @@
 (in-package "REMORA")
 
 (include-book "expression-values-and-environments")
+(include-book "abstract-syntax-constructors")
+
+(local (include-book "kestrel/lists-light/len" :dir :system))
 
 (acl2::controlled-configuration)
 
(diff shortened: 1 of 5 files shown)


books/kestrel/remora/primitives-evaluation-on-ispaces.lisp
```diff
@@ -78,6 +81,90 @@
...
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define prim-reify-dim ((d natp))
+  :returns (val expr-valuep)
+  :short "Evaluation of dimension reification."
+  :long
+  (xdoc::topstring
+   (xdoc::p
+    "This is the semantics of the instantiated @('reify-dim') operation:
```
acoglio:
```suggestion
    "This is the semantics of the @('reify-dim') operation:
```

Since this operation does not have intermediate stages like others, we can omit 'instantiated'.

books/kestrel/remora/primitives-evaluation-on-ispaces.lisp
```diff
@@ -78,6 +81,90 @@
...
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define prim-reify-shape ((s nat-listp))
+  :returns (val expr-valuep)
+  :short "Evaluation of shape reification."
+  :long
+  (xdoc::topstring
+   (xdoc::p
+    "This is the semantics of the instantiated @('reify-shape') operation:
```
acoglio:
```suggestion
    "This is the semantics of the @('reify-shape') operation:
```

Same rationale as previous comment.

books/kestrel/remora/primitives-evaluation-on-ispaces.lisp
```diff
@@ -103,8 +190,11 @@
      a dimension for the stages that store just a type value
      (except @(':reshape-t'), which expects the first of two shapes),
      a shape for the stages that also store a dimension or a shape;
-     the uninstantiated stage of @('sum'), which has no type parameter,
-     stores nothing and expects a shape directly.
+     the uninstantiated stages of
```
acoglio: `reify-dim` and `reify-shape` have just one stage, so they don't really have uninstantiated stages. Maybe this can be rephrased slightly, e.g. "@('reify-dim') and @('reify-shape') have no type parameters".

acoglio: In fact, that's the case for `iota/static` too: it has just one stage, already instantiated, i.e. ready to be applied to the final argument.

acoglio (approved):
Looks good, just a few comments about the documentation.

acoglio:
> For `prim-reify-shape`, it produces a box, so the function body is calling `make-type-value-sigma` and `make-expr-value-box` directly. Do you think this is the correct approach?

Yes, that's the right approach.

acoglio (approved):

PR #2021, opened 2026-08-26
Prove monomorphize preserves corep and well-formedness

westfold:
Adds some new support files and a few theorems in existing support files.

### DIFF
--- a/books/kestrel/remora/variable-substitution-operations.lisp
+++ b/books/kestrel/remora/variable-substitution-operations.lisp
@@ -68,6 +68,25 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+; A map from type-variable name strings to types is used to resolve type
+; arguments.  We use the STRING-TYPE-MAP omap so that the type-variable
+; substitution operations below can be applied directly to it.
+
+(define extend-type-var-map ((tvars type-var-listp)
+                             (tys type-listp)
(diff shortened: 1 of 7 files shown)


books/kestrel/remora/variable-substitution-operations.lisp
```diff
@@ -68,6 +68,25 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+; A map from type-variable name strings to types is used to resolve type
+; arguments.  We use the STRING-TYPE-MAP omap so that the type-variable
+; substitution operations below can be applied directly to it.
+
+(define extend-type-var-map ((tvars type-var-listp)
```
acoglio: It seems strange that this operation is ignoring the kinds of the type variables, just taking their names. Could it be refactored to take names instead of type variables? I don't have the context for where this is used.

Also, I'm not sure that this operations belongs in this file, since it's not specifically about variable substitutions, and the rest of this file doesn't use/need this operation. It's an operation about the string-type-map data type, really. Maybe even a generic omap operation that updates an omap with a list of keys and values.

It's fine with me to merge as is, but we should probably revisit this.

acoglio (approved):

acoglio:
I think it would be good to merge the latest testing-kestrel into this PR before we merge it, just to ensure that things certify, given that a few more wf conditions may have been moved into the fixtypes in the meanwhile.

PR #2027, opened 2026-08-31
move from private repo and add per-round Appendix B tests

bendyarm:


bendyarm:
Closing, since the tests have been added separately by @ericwhitmansmith

PR #2028, opened 2026-08-31
[Remora] Extend frame-flattening  to bracket expressions

Sarah-Scott:
Extend frame-flattening to bracket expressions so that it yields the same results as [impl].

### DIFF
--- a/books/kestrel/remora/frame-flattening.lisp
+++ b/books/kestrel/remora/frame-flattening.lisp
@@ -51,7 +51,30 @@
      @(tsee flatten-frames-in-expr) does the merging at each
      one it finds following only frame and bracket spines.
      So a frame written explicitly with @('frame') is merged
-     into its parent only when an enclosing bracket reaches it."))
+     into its parent only when an enclosing bracket reaches it.")
+   (xdoc::p
+    "The general idea is that @(tsee flatten-frames-in-expr)
+     turns an expression of the form")
+   (xdoc::codeblock
+    "(frame [a1 ... aN] (frame [b1 ... bM] x11 ... x1P)"
+    "                   (frame [b1 ... bM] x21 ... x2P)"
+    "                   ..."
+    "                   (frame [b1 ... bM] xQ1 ... xQP))")
+   (xdoc::p
+    "into")
+   (xdoc::codeblock
+    "(frame [a1 ... aN b1 ... bM] x11 ... xQP)")
+   (xdoc::p
+    "and an expression of the form")
+   (xdoc::codeblock
+    "(frame [a1 ... aN] (array [b1 ... bM] x11 ... x1P)"
+    "                   (array [b1 ... bM] x21 ... x2P)"
+    "                   ..."
+    "                   (array [b1 ... bM] xQ1 ... xQP))")
+   (xdoc::p
+    "into")
+   (xdoc::codeblock
+    "(array [a1 ... aN b1 ... bM] x11 ... xQP)"))
   :order-subtopics t
   :default-parent t)
 
@@ -80,7 +103,7 @@
    (xdoc::p
     "An expression is array-like if it "
      (xdoc::seetopic "desugaring" "desugars")
-     " to a non-empty array.")
+     " to a non-empty @(':array') expression.")
    (xdoc::p
     "This matches [impl], whose parser turns atoms and strings
      into arrays as it parses, so its flattening sees arrays
@@ -96,7 +119,14 @@
 (std::deflist expr-array-like-listp (x)
   :guard (expr-listp x)
   :short "Recognizer for a true list of array-like expressions."
-  (expr-array-likep x))
+  (expr-array-likep x)
+
+  ///
+
+  (defrule expr-array-like-listp-of-expr-list-fix
+    (equal (expr-array-like-listp (expr-list-fix x))
+           (expr-array-like-listp x))
+    :induct t))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
@@ -108,7 +138,7 @@
              :array expr.dims
              :atom nil
              :string (list (len expr.chars))
-             :otherwise nil)
+             :otherwise (impossible))
   :guard-hints (("Goal" :in-theory (enable expr-array-likep))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -137,7 +167,7 @@
              :string (atom-base-list
                       (base-lit-int-list
                        (char-lit-list-desugar expr.chars)))
-             :otherwise nil)
+             :otherwise (impossible))
   :guard-hints (("Goal" :in-theory (enable expr-array-likep))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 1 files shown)


books/kestrel/remora/frame-flattening.lisp
```diff
@@ -60,6 +72,207 @@
...
 
+(define expr-array-likep ((expr exprp))
+  :returns (yes/no booleanp)
+  :short "Recognizer for array-like expressions."
+  :long
+  (xdoc::topstring
+   (xdoc::p
+    "An expression is array-like if it "
+     (xdoc::seetopic "desugaring" "desugars")
+     " to a non-empty array.")
```
acoglio:
```suggestion
     " to a non-empty @(':array') expression.")
```

Rationale: 'array expression' could be interpreted ambiguously, since in fact every Remora expression returns an array value. The suggested edit makes it clear that we refer to a very specific AST.

books/kestrel/remora/frame-flattening.lisp
```diff
@@ -60,6 +72,207 @@
...
+
+(define expr-array-like->dims ((expr exprp))
+  :guard (expr-array-likep expr)
+  :returns (dims nat-listp)
+  :short "Get the dimensions of an array-like expression."
+  (expr-case expr
+             :array expr.dims
+             :atom nil
+             :string (list (len expr.chars))
+             :otherwise nil)
```
acoglio:
```suggestion
             :otherwise (impossible))
```

This generates a guard obligation that this `:otherwise` is unreachable. Simple given the definitions, but useful to double-check things, and for maintenance, e.g. if a new kind of expression is added.

books/kestrel/remora/frame-flattening.lisp
```diff
@@ -60,6 +72,207 @@
...
+    "For a string expression, these are the atoms obtained by "
+    (xdoc::seetopic "char-lit-list-desugar" "desugaring")
+    "."))
+  (expr-case expr
+             :array expr.atoms
+             :atom (list expr.atom)
+             :string (atom-base-list
+                      (base-lit-int-list
+                       (char-lit-list-desugar expr.chars)))
+             :otherwise nil)
```
acoglio:
```suggestion
             :otherwise (impossible))
```

Same as above.

books/kestrel/remora/frame-flattening.lisp
```diff
@@ -60,6 +72,207 @@
...
+
+(define expr-frame-likep ((expr exprp))
+  :returns (yes/no booleanp)
+  :short "Recognizer for frame-like expressions."
+  :long
+  (xdoc::topstring
+   (xdoc::p
+    "An expression is frame-like if it "
+     (xdoc::seetopic "desugaring" "desugars")
+     " to a non-empty frame.")
```
acoglio:
```suggestion
     " to a non-empty @(':frame') expression.")
```

Same rationale as above.

books/kestrel/remora/frame-flattening.lisp
```diff
@@ -60,6 +72,207 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define expr-frame-like->dims ((expr exprp))
+  :guard (expr-frame-likep expr)
+  :returns (dims nat-listp)
+  :short "Get the dimensions of a frame-like expression."
+  (expr-case expr
+             :bracket (list (len expr.exprs))
+             :frame expr.dims
+             :otherwise nil)
```
acoglio: Can use `(impossible)` — as earlier.

books/kestrel/remora/frame-flattening.lisp
```diff
@@ -60,6 +72,207 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define expr-frame-like->exprs ((expr exprp))
+  :guard (expr-frame-likep expr)
+  :returns (exprs expr-listp)
+  :short "Get the sub-expressions of a frame-like expression."
+  (expr-case expr
+             :bracket expr.exprs
+             :frame expr.exprs
+             :otherwise nil)
```
acoglio: Can use `(impossible)` — as earlier.

books/kestrel/remora/frame-flattening.lisp
```diff
@@ -60,6 +72,207 @@
...
+  (expr-frame-like->exprs x))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define flatten-merge ((dims nat-listp)
+                       (exprs expr-listp)
+                       (bracketp booleanp))
+  :guard (consp exprs)
+  :returns (expr exprp)
+  :hooks nil
```
acoglio: There's no reason why the fixing theorems shouldn't be provable for this function. I would remove the `:hooks nil` and try to prove them. In some cases it's necessary to add some fixer calls in the body, but those are generally no-ops in execution.

books/kestrel/remora/frame-flattening.lisp
```diff
@@ -182,16 +334,75 @@
...
+   (xdoc::p
+    "This walks the whole AST and applies @(tsee flatten-frames-in-expr)
+       at each bracket expression, which is where [impl] applies its
+       flattening.")
+   (xdoc::p
+    "A frame expression written explicitly with @('frame')
+       has no override here, so it is not merged into its parent;
+       it is only merged if some enclosing bracket expression
+       reaches it via @(tsee flatten-frames-in-expr)."))
+  :types (shapes/ispaces
```
acoglio: You only need to pass the exprs/atoms/binds clique here, because these folds are no-ops for the fixtypes in the other cliques.

books/kestrel/remora/frame-flattening.lisp
```diff
@@ -182,16 +334,75 @@
...
+                    (flatten-frames-in-expr
+                     (make-expr-bracket :exprs exprs)))))
+  :name ast-flatten-brackets)
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (defsection corep-of-flatten-frames
   :short "Frame flattening core ASTs yields core ASTs."
 
+  (defrulel atom-list-list-corep-of-expr-array-like-list->atoms
```
acoglio: There's no need for these three theorems to be local. You could remove the `l` in `defrulel`. I think it's fine for these theorems to be enabled by default.

books/kestrel/remora/frame-flattening.lisp
```diff
@@ -71,80 +284,19 @@
     :long
     (xdoc::topstring
      (xdoc::p
```
acoglio, line 322: I think that this (now deleted) text makes for a good "visual" explanation of the idea of frame flattening. Perhaps we could append it at the end of the `(defxdoc+ frame-flattening ...)`? Just after the 2nd paragraph there, we could start with this paragraph "The general idea is that ..." up to and excluding the "More precisely..." paragraph.

acoglio (changes requested):
The substance looks good, but I gave some suggestions for improvements.

Sarah-Scott:
@acoglio I made all of your suggested improvements.

acoglio (approved):

PR #2029, opened 2026-09-02
[Remora] Add AST modes to deserializer

Sarah-Scott:
Added three AST modes to the JSON deserializer specifying which instantiation of [impl]'s ASTs to expect. The modes are `:unchecked`, `:unique`, and `:checked`. Also renamed the file-I/O entry points to match similar Remora functions.


acoglio (approved):
I haven't double-checked all the code since there's a lot of it, but it looks reasonable.

PR #2030, opened 2026-09-03
Generalize monomorphize to handle shapes properly.

westfold:
Factor out part of evaluation.lisp into ispace-evaluation.lisp that is needed by monomorphize.
Update property proofs of monomorphize and add a test file.


books/kestrel/remora/monomorphize.lisp
```diff
@@ -296,88 +306,56 @@
...
+(define eval-iargs ((isps ispace-listp) (denv ispace-denvp))
+  :returns (mv (err booleanp) (ivals ispace-value-listp))
+  :short "Evaluate the ispace arguments at a call site to ispace values."
+  (b* ((ivals (eval-ispace-list isps denv))
+       ((when (reserrp ivals)) (mv t nil)))
+    (mv nil ivals)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(define nats-to-dims ((nats nat-listp))
```
acoglio: This is the same as `dim-const-list` in abstract-syntax-structurals.lisp, which also comes with theorems.

acoglio (approved):
Moving ispace evaluation to its own file and topic is good, and something we wanted to do. I haven't looked at the details of the monomorphization code, but it looks reasonable to me.

books/kestrel/remora/monomorphize.lisp
westfold: fixed

PR #2035, opened 2026-09-15
[cert.pl] fix certdeps build by changing "&> Makefile-features.out" t…

bendyarm:
features.sh runs ACL2 on cert_features.lsp through `$STARTJOB -c "..."`.
When invoked from the ACL2 GNUmakefile (target update_books_build_info,
run at the end of every ACL2 build) or from books/GNUmakefile, STARTJOB
defaults to make's SHELL, which is /bin/sh.  On Debian and Ubuntu,
/bin/sh is dash, which does not know the bash-only redirect "&>" and
parses

  $ACL2 < cert_features.lsp &> Makefile-features.out

as `$ACL2 < cert_features.lsp &` followed by `> Makefile-features.out`.
So ACL2 ran in the background, Makefile-features.out was always empty,
a failure to run ACL2 was never reported, and features.sh returned
before the .certdep files had been written.

Usually the background ACL2 finishes a moment later and nobody notices.
But in a Docker build, the RUN step that builds ACL2 ends as soon as
make returns, leaving the image without build/acl2-version.certdep and
friends, so a later cert.pl fails with "Missing build/acl2-exports.certdep".
(Observed on Ubuntu 24.04 images of ACL2 on both CCL and SBCL.)

We write the command string in POSIX sh instead.  Verified on Ubuntu 24.04
(dash) that `make update_books_build_info` now writes all .certdep files
before returning, that a bad $ACL2 now produces the "*** Failed to run
ACL2! ***" message, and that the standalone (STARTJOB=bash) and
books/GNUmakefile invocations behave as before.


gjurgensen (approved):
LGTM.

I think there may be other places where `&>` is used. But they might all be in places where we know the script is being run by bash.

gjurgensen:
Perhaps relevant, I'll just note a way in which the top-level Makefile and the books Makefile differ.

The top-level Makefile sets `STARTJOB` unconditionally to `$(SHELL)` before running the features script: https://github.com/acl2/acl2/blob/master/GNUmakefile#L313.

In contrast, the book-level Makefile only sets `STARTJOB` when it is unset. If unset, it will try to set it to `bash` or the user-specified shell: https://github.com/acl2/acl2/blob/master/books/GNUmakefile#L176-L185.

Perhaps the top-level Makefile should be consistent in defaulting `STARTJOB` to bash.

solswords (approved):
this is fine, but I do also endorse Grant's suggestion of setting Make's SHELL to bash so that we have less exposure to these compatibility issues.

MattKaufmann:
In the top level GNUmakefile, it looks like STARTJOB is only set in target update_books_build_info, here:

```
export ACL2=$(ACL2) ; export STARTJOB=$(SHELL) ; cd books/build ; (./features.sh 2>&1) > features.out
```

It doesn't seem to be valid bash syntax to change `export STARTJOB=$(SHELL)` to `export STARTJOB?=$(SHELL)`.  What change should be made?  Should that line stay as is but instead define SHELL in GNUmakefile as it is defined in books/GNUmakefile?  Should we wait till after the release ("if it ain't broke, don't fix it" -- at least when not yet necessary?)?

gjurgensen:
I agree setting `SHELL` the same we do in the books Makefile would be best.

I think delaying the change until after the release is fine.

PR #2036, opened 2026-09-15
[emacs] Name the link face directly in `acl2-doc-keywords`.

gjurgensen:
Emacs 30 marks `font-lock-keyword-face` and friends obsolete as variables, so binding `font-lock-keyword-face` buffer-locally to `acl2-doc-link-face` produces a byte-compile warning. This uses the `(MATCHER SUBEXP FACENAME)` form of `font-lock-keywords` to name `acl2-doc-link-face` directly instead, which also removes the need to rebind a global font-lock variable.

@MattKaufmann I think this version of the elisp file is more or less identical to the non-books version. I'm not sure if you would want to modify the non-books version as well. On one hand, the non-books version targets a fixed emacs version IIRC. On the other hand, it might be best to keep them in sync if possible.


MattKaufmann (commented):
I have no objection to the code modification (not that I understand it, but I don't think I need to).  The only question I have is whether it works back through Emacs version 25.  If not, then you might need to conditionalize your code on the Emacs version.  There's no need to change the non-books version, which is only for Emacs version 24 (as noted in the README file in its directory).  Once you address that issue (possibly saying that it's not an issue and your change is fine for Emacs version 25), I'll approve the PR.

gjurgensen:
Yes, this fix works in older Emacs. I tested on 25, 29, and 30.

MattKaufmann (approved):
Thanks for confirming.  Let me know if you'd like me to merge.

PR #2038, opened 2026-09-18
Add README note about books vs the ACL2 system.

ericwhitmansmith:
Added at Matt's request, to avoid problems if people see commits to ACL2 from folks other than Matt and J (like me).


MattKaufmann (approved):
Looks great; thanks, @ericwhitmansmith .

PR #2039, opened 2026-09-18
[bin] Add myself to purity script.

ericwhitmansmith:
As discussed with Matt via phone.  I will still have Matt approve any PRs I make that affect the system.


MattKaufmann (approved):
Looks fine to me.

PR #2047, opened 2026-09-19
Fix all-vars!1.

ericwhitmansmith:
The call (lambda-object-formals (ffn-symb term)) could not have been right, because in this case term is a quoted constant.

Tested with a clean regression-everything.


MattKaufmann (approved):

MattKaufmann:
Was this too found by Claude?  If so, were any examples provided of observably bad behavior (especially, unsound behavior)?

ericwhitmansmith:
Yes, Claude found it.  Claude thinks it is benign (note: I have not vetted this reasoning):

"
The Warning comment on `all-vars!-of-fn` (`history-management.lisp:13202`) states
the contract: the list exists *"to allow us to genvar a new variable symbol that
is not used as a variable in any way in fn."* A missed formal is therefore a
fresh-variable generator that can return a name already in use.

I did not build an exploit from it: the only consumer is the `newvar` at
`:14190`, and an *unused* lambda-object formal by construction appears nowhere in
the generated guard conjectures, so the collision is harmless there today. It is
a latent hazard rather than a live one — but incomplete `genvar` avoid-lists are
the root cause of two confirmed proofs of `nil` in this review (the
functional-instantiation rename collision, and the DO-`loop$` `mv-setq` capture),
which is why it is worth closing."

I asked Claude for an observable (event-level) test, and it created the following (which I have not reviewed in detail):

```
; event-level-test.lisp
;
; A NORMAL ACL2 EVENT whose outcome differs before and after the fix at
; history-management.lisp:13179 (ALL-VARS!1's quoted-lambda-object branch passes
; (ffn-symb term) -- i.e. the symbol QUOTE -- to LAMBDA-OBJECT-FORMALS, so the
; union is a no-op; the local OBJ is already bound, so the fix is
; (ffn-symb term) -> obj).
;
; The only consumer of ALL-VARS!-OF-FN is the NEWVAR in GUARD-CLAUSES-FOR-FN
; (history-management.lisp:14184-14190):
;
;   (genvar <base> "NEWV" nil (all-vars!-of-fn name wrld))
;
; and NEWVAR is used only in the Special loop$ Conjectures (a) and (b)
; (special-conjectures, :13392, substitutions at :13543-13574).  So to see the
; defect at the event level you need a function that
;   (1) contains a special loop$ scion call -- SUM$ qualifies -- with a
;       non-trivial lambda guard, so conjecture (a) is actually generated, and
;   (2) contains a quoted LAMBDA object with an UNUSED formal named NEWV.
;
; Note (2) can only be written with an IGNORE/IGNORABLE declaration: translate
; rejects an unused lambda-object formal outright ("must be declared IGNOREd or
; IGNORABLE ... This is violated by the body U, which fails to use the variable
; NEWV").  And the DECLARE is exactly the part of the object ALL-VARS!1 does not
; walk -- it walks LAMBDA-OBJECT-BODY and LAMBDA-OBJECT-GUARD only.
;
; RUN: /path/to/saved_acl2 < event-level-test.lisp

(include-book "projects/apply/top" :dir :system)

(defun tree-has (s x)                   ; is symbol S anywhere in tree X?
  (declare (xargs :mode :program))
  (cond ((atom x) (eq s x))
        (t (or (tree-has s (car x)) (tree-has s (cdr x))))))

;-----------------------------------------------------------------------------
; HOST5: the bug case.  NEWV is an ignored formal of the quoted lambda object,
; so ALL-VARS!-OF-FN misses it and GENVAR hands NEWV back as "fresh".

(defun host5 (lst)
  (declare (xargs :guard (true-listp lst) :verify-guards nil))
  (cons (apply$ '(lambda (u newv) (declare (ignore newv)) u) (list 3 4))
        (loop$ for e in lst sum (car e))))

; BEFORE the fix this assert SUCCEEDS  -> the fresh variable is NEWV.
; AFTER  the fix it FAILS             -> NEWV is in the avoid list, so GENVAR
;                                        picks NEWV0 instead.
(make-event
 (mv-let (erp val)
   (guard-obligation 'host5 nil nil t 'top state)
   (declare (ignore erp))
   (prog2$
    (cw "~%HOST5: fresh var is NEWV? ~x0   NEWV0? ~x1~%"
        (tree-has 'newv (cadr val)) (tree-has 'newv0 (cadr val)))
    '(value-triple :host5-reported))))

;-----------------------------------------------------------------------------
; HOST6: the control.  Same lambda object, but NEWV is ALSO a formal of the
; enclosing function, so it genuinely occurs free in the body, ALL-VARS!1 finds
; it through the ordinary variable case, and GENVAR must avoid it.  This should
; report NEWV0 both before and after the fix -- i.e. the avoid list already
; works for everything that actually occurs.

(defun host6 (lst newv)
  (declare (xargs :guard (true-listp lst) :verify-guards nil) (ignorable newv))
  (cons newv
        (cons (apply$ '(lambda (u newv) (declare (ignore newv)) u) (list 3 4))
              (loop$ for e in lst sum (car e)))))

(make-event
 (mv-let (erp val)
   (guard-obligation 'host6 nil nil t 'top state)
   (declare (ignore erp))
   (prog2$
    (cw "~%HOST6 (control): fresh var is NEWV? ~x0   NEWV0? ~x1~%"
        (tree-has 'newv (cadr val)) (tree-has 'newv0 (cadr val)))
    '(value-triple :host6-reported))))

;-----------------------------------------------------------------------------
; The discriminating event.  This ASSERT-EVENT FAILS on unfixed ACL2 and
; SUCCEEDS once (ffn-symb term) is changed to obj.

(make-event
 (mv-let (erp val)
   (guard-obligation 'host5 nil nil t 'top state)
   (declare (ignore erp))
   (list 'assert-event
         (list 'quote (tree-has 'newv0 (cadr val)))
         ':msg
         "The Special loop$ conjecture for HOST5 still uses NEWV as its fresh
          variable, which is the ignored formal of the quoted LAMBDA object in
          HOST5's body.  ALL-VARS!-OF-FN missed that formal, so GENVAR did not
          avoid it.  (history-management.lisp:13179)")))
```

